PR #6, opened 2019-11-08
Make aries-site with Nextjs

taysea:
This PR converts Aries-site to use Next js. The site renders a component coming from Aries-core.

Some notes on files added/changed: 
- `next.config.js` is necessary in order to transpile the local aries-core dependency (https://github.com/martpie/next-transpile-modules). A note from the README says this plugin aims to solve the following challenge: “code transpilation from local packages”. In this file, I provide the name of the workspace we want to transpile.


package.json
```diff
@@ -2,7 +2,6 @@
   "name": "aries",
   "private": true,
   "workspaces": [
-    "aries-core",
-    "aries-site"
+    "packages/*"
```
ShimiSun: I'm not convinced about the 'packages' directory, I think at this point it adds unnecessary abstraction we don't need. 
It also might affect unpleasantly the way `hpe-server` connects with aries-core.

taysea: I don't think it would have a negative effect, but I'll need to confirm that. Regardless, this packages/* structure mirrors what would be done if we used Lerna and seems to be a common pattern I've seen across other projects using yarn workspaces. I do agree that maybe it's an unnecessary abstraction considering we only have two workspaces at the moment, but I don't think it will have negative affects. I will look for some more resources.

taysea:
Linking to #6

PR #14, opened 2019-11-14
Publish stable branch

halocline:
Related Issues: https://github.com/hpe-design/hpe-server/issues/2

What does this PR do?
- Allows for aries-core to be consumed by other projects by including the following in their `package.json`:
```
"aries-core": "https://github.com/hpe-design/aries.git#stable"
```

Work Included:
- Added `/tools/release-stable.js`
- Added `release-stable` scripts to `package.json`
- Modify `build` script to include babel config to prep aries-core for dist


aries-core/babel.config.js
```diff
@@ -1,16 +1,15 @@
...
-    "@babel/preset-react",
-  ];
-  const plugins = [
-    "@babel/plugin-proposal-class-properties",
-  ];
+  const presets = ['@babel/preset-env', '@babel/preset-react'];
+  const plugins = ['@babel/plugin-proposal-class-properties'];
+
+  /* Ignoring test files, test directories, and stories from /dist */
+  const ignore = ['src/**/*.test.js', 'src/**/__tests__', 'src/**/stories'];
```
ShimiSun: glad you figured out how to exclude stories & tests from production

ShimiSun:
The  babel.config.js changes on this PR broke storybook with the following error:
``` Module parse failed: Unexpected token (28:45)
File was processed with these loaders:
 * ../node_modules/babel-loader/lib/index.js
 * ../node_modules/@storybook/source-loader/dist/server/index.js
You may need an additional loader to handle the result of these loaders.
| };
| 
> export const text = addSourceDecorator(() => 
, {__STORY__, __ADDS_MAP__,__MAIN_FILE_LOCATION__,__MODULE_DEPENDENCIES__,__LOCAL_DEPENDENCIES__,__SOURCE_PREFIX__,__IDS_TO_FRAMEWORKS__});
```
I've excluded the story files from the  babel.config.js `ignore` configuration, and it works as expected now.
I've compared with the grommet behavior and it looks like the story files are distributed in grommet as well, I don't think It is the optimal behavior, but I'm out of ideas for the moment on how to workaround excluding the files without letting it affect the babel loading of the stories. so unless you want to try something else I'm ok with the current solution (and with my commit of excluding the story files from babel) without further action.

halocline:
Good find and correction. Thanks for the analysis. I wonder if we can solely apply the story ignore to the `yarn build` script. That would leave it out of the babel-config.js and only apply it to the aries-core build and not the storybook build.  Thoughts?

ShimiSun:
Yep, I think it will be an interesting approach to tackle it. 
If it will work I think we should consolidate both `tests` and `stories` dir(s) to be ignored in the same place., hence moving the `ignore` of `tests` from the babel config to the build script.

PR #16, opened 2019-11-15
Revert "Test GitHub action"

halocline:
Reverts hpe-design/aries#15


ShimiSun:
why was that reverted?

halocline:
I wanted to confirm the action showed up as expected, but I want to do more testing on the configuration before committing.

PR #28, opened 2019-11-18
Simple start for site layout

taysea:
This is a very light start on the aries-site layout. It doesn't take in any of the theming that will be added but just establishes the nav component within a layout component.


taysea:
linking to #22

PR #40, opened 2019-11-21
Restructured components hierarchy

ShimiSun:
Instead of the structure of the current components:
-src
    -components
        -.....
And have a single list of components of either helpers or core components, we will adopt the following structure to distinguish between core components and helpers, it will be easier to maintain and scale.
  -src
    -components
        -core (components that are part of the published design systems)
        -helpers (helper components to achieve the core components)


ShimiSun:
#42

Issue #42, opened 2019-11-21
Maintaining solid structure of the repo


ShimiSun:
In general, I think we should adopt a more lean structure of packages and directories and every package that export components should have its own index.js file.

Regarding lean packages, let's try not to exceed more than ~7 components in a directory and think of reasonable ways to abstract to subdirectories in case we are exceeding it. So far I think we are good with that.

Before we are continuing to extend our codebase let’s try to clean it and have a solid structure of the components, packages, and index.js files.

Regarding the `index.js` files, the best practice is to put an index.js file in every directory that exports a component, so far I've completed the first two:
- [x] aries-core
- [x] aries-site/components
- [x] aries-site/layouts
- [x] aries-site/pages (not sure if needed on a next.js structure)

For Reference:
From best practices articles:
https://medium.com/hackernoon/the-100-correct-way-to-structure-a-react-app-or-why-theres-no-such-thing-3ede534ef1ed
```
The first is to put an index.js file in every directory that exports a component
```

PR #43, opened 2019-11-21
Restructure layouts

taysea:
I restructured the layouts directory so that each folder has an index.js file that exports all the components within the folder. Then, the main layouts folder has a single index.js that exports everything within each folder. This will benefit scalability in the future.


taysea:
#42

PR #45, opened 2019-11-21
Restructuring app layout

ShimiSun:
Refactoring structure with the new PageLayout component.
This PR is to enable the work on the Sidebar.

### DIFF
--- a/aries-site/src/pages/design/data-visualization.js
+++ b/aries-site/src/pages/design/data-visualization.js
@@ -1,22 +1,16 @@
 import React from 'react';
-import { Text } from 'grommet';
 
-import { ContentSection, Layout, MainContent, SideBar } from '../../layouts';
+import { ContentSection, PageLayout, MainContent } from '../../layouts';
 import { MainHeading } from '../../components';
 
 const Index = () => (
-  <>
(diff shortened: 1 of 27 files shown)


aries-site/src/pages/design/data-visualization.js
```diff
@@ -1,22 +1,25 @@
...
-      </SideBar>
-      <MainContent>
-        <ContentSection>
-          <MainHeading>Data-visualization</MainHeading>
-        </ContentSection>
-      </MainContent>
-    </Layout>
-  </>
+  <PageLayout title="Data-visualization">
+    <SideBar>
```
taysea: Some files still have the `SideBar` inside individual pages. I think we can remove it since it's now contained in `PageLayout` right?

ShimiSun: I'm on it, good catch

taysea: I didn't comment on each file that has it, but there are a couple more in addition to this one.

ShimiSun: Of course. we should be good to go now.

PR #49, opened 2019-11-22
Adding an Undecorated Anchor

ShimiSun:
For reusability of the ThemeContext undecorated AnchorGroup this component will serve both Header Nav and the Sidebar.
#46

Issue #46, opened 2019-11-22
Control site routing with Sidebar


ShimiSun:
Put the SideBar in place to resolve the routing and the corresponding content.

PR #50, opened 2019-11-22
Refactoring layout to achieve cleaner SideBar layout

ShimiSun:
Let's discuss these changes along with the changes on #49 on a call today so I can bring you up to speed on the approach.
I addition, I also made changes to couple of components on aries-core to address the site in a cleaner way.
related to #46

### DIFF
--- a/aries-site/src/layouts/main/Layout.js
+++ b/aries-site/src/layouts/main/Layout.js
@@ -3,8 +3,8 @@ import Head from 'next/head';
 import PropTypes from 'prop-types';
 import { Box, Grommet, ResponsiveContext } from 'grommet';
 import { hpe } from 'grommet-theme-hpe';
-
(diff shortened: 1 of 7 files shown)


aries-site/src/layouts/main/Layout.js
```diff
@@ -24,7 +24,7 @@ const filterChildren = (children, type) => {
 export const Layout = ({ children, title }) => {
   // Only expect a single child of the following types
   const mainContent = filterChildren(children, 'MainContent');
-  const sidebar = filterChildren(children, 'SideBar');
+  const selectedNav = 'start';
```
ShimiSun: @taysea Following our call today, this variable should be determined from the Nav component on the aries-core. 
I think it will be the final step for bringing all the functionality together 🎉

PR #56, opened 2019-11-22
Complete routing structure and sidebar navigation 

### LINKED ISSUE #46: Control site routing with Sidebar

ShimiSun:
closes #46 
We've reviewed on our call today 100% of these changes (conceptually).
I've separated this PR into small commits so it will be easier to review.
Tested with changing the selected NavHeader manually for each different sidebar and it all worked as expected, no 404 or other failures.


aries-site/src/layouts/main/Layout.js
```diff
@@ -64,7 +64,10 @@ export const Layout = ({ children, title }) => {
               }}
             >
               <Box fill="vertical">
-                <SideBar items={SideBarItemList[selectedNav]} />
+                <SideBar
+                  items={SideBarItemList[selectedNav]}
+                  prefix={selectedNav}
+                />
```
halocline: So, since this layout now always has a sidebar, I think we will need to create two new issues:
1) We will need a layout for the "landing page" which does not have a sidebar
2) Hide sidebar at the same breakpoint Nav goes from desktop to mobile

Do those items sound correct?

aries-site/src/components/content/BulletedList.js
```diff
@@ -6,7 +6,9 @@ export const BulletedList = ({ items }) => {
   return (
     <>
       {items.map(item => (
-        <Text margin={{ bottom: 'xsmall' }}>— {item}</Text>
+        <Text margin={{ bottom: 'xsmall' }} key={item}>
+          — {item}
```
halocline: I think we need an extra space between - and {.  I believe the design shows the spacing to be equivalent to the length of the dash.

aries-site/src/layouts/main/Layout.js
taysea: Agreed on both of those items, @halocline. We could make a `LandingPage` within the layouts folder. For the sidebar, the breakpoint should be at `small`

aries-site/src/pages/start/about.js
```diff
@@ -0,0 +1,74 @@
...
+      </MainDescription>
+      <Text>
+        HPE's Aries connects the ideas, technology, and creativity from Hewlett
+        Packard Enterprise and exposes it to the larger community with the hope
+        of learning and building experiences that reach a broader audience.
+      </Text>
+      <Box direction="row" gap="medium">
+        <Anchor
+          label="Our Principles"
+          href="/principles"
```
halocline: This href needs to be updated to "/start/principles"

aries-site/src/components/content/BulletedList.js
halocline: FYI, in code editor the dash and the space are equivalent, but when rendered in MetricHPE the spacing is off.
Screen Shot 2019-11-22 at 2 33 40 PM

halocline: Actually, disregard this for now. Sketch files are inconsistent in this treatment. The Invision execution has the single space, but some of the Sketch files show the double space...

taysea:
Added a small feedback on one of Matt's comments, but other than that looks good to me!

aries-site/src/pages/start/about.js
taysea:
```suggestion
          href="/start/principles"
```

aries-site/src/layouts/main/Layout.js
ShimiSun: Sounds good to me.
IMO we will need a separate layout for the landing page, but let's see how the design looks like first and go from there.

taysea: Yes, I agree that it will be a separate layout. I meant it could be a file inside the layouts directory. maybe: `layouts/main/LandingPageLayout.js`

PR #57, opened 2019-11-22
Feat aries theming

halocline:
Dropping in this theme while `grommet-theme-hpe` is still in the works. More to do, but looks pretty great.


ShimiSun:
Oh my!! BEAUTIFUL to see how all the work had come together!

aries-site/src/themes/aries.js
```diff
@@ -0,0 +1,115 @@
...
+      background: 'active-background',
+      color: 'active-text',
+    },
+    selected: {
+      background: 'selected-background',
+      color: 'selected-text',
+    },
+    graph: {
+      colors: {
+        light: ['brand\nstatus-warning'],
```
ShimiSun: is light accepting an array now rather than only string?

taysea:
Wow that was so exciting to open the project and see the new colors!

taysea:
I'm less familiar with theme, so I think it would be good to address Shimi's comment regarding the `light` array vs string, but everything else looks good to me.

aries-site/src/themes/aries.js
halocline: I need to investigate that further. Pulled from the theme designer export. Need to play with a few of these more, but thought it'd be fun to get some styling in for Monday ;)

ericsoderberghp: That should have been exported as `light: ['brand', 'status-warning']`. That's a bug in the theme designer there.

The reason this is an array so it can be used for multi-value Meter, Diagram lines, Distribution boxes, and eventual multi Chart graphics. You can set an arbitrary number of colors to cycle through in those cases. This avoid the problem we had with relying on `accent-*` where different themes had different numbers of accent colors. And, it makes it more explicit. I'm working on updating grommet to use this. Pull request coming, but might not land until December, given next week's travel.

taysea:
```suggestion
        light: ['brand', 'status-warning'],
```

halocline: I pulled the graph styling out for now. We can add it in at a later date when appropriate.

aries-site/src/themes/aries.js
```diff
@@ -0,0 +1,110 @@
...
+      'text-xweak': {
+        dark: '#999999',
+        light: '#666666',
+      },
+      border: 'background-xweak',
+      control: 'brand',
+      'active-background': 'background-weak',
+      'active-text': 'text-strong',
+      'selected-background': 'brand',
+      'selected-text': '',
```
ShimiSun: This looks suspicious, what it be preferred to remove it?

aries-site/src/themes/aries.js
```diff
@@ -0,0 +1,110 @@
...
+        dark: '#000000',
+        light: '#FFFFFF',
+      },
+      'background-weak': {
+        dark: '#44444480',
+        light: '#E8E8E880',
+      },
+      'background-xweak': {
+        dark: '#66666699',
+        light: '#CCCCCC90',
```
ShimiSun: I thought HEX color should be represented with 6 characters, is this coming from the designer? 
(the selected colors has 8 characters)

ShimiSun: Interesting! well, they might need to change the name from HEX to OCT now...

aries-site/src/themes/aries.js
halocline: Good catch. Setting as undefined for now.

aries-site/src/themes/aries.js
halocline: Yes. @taysea is correct.

@ShimiSun Are you going to lead the lobbying effort for the name change? ;)

ShimiSun: lol, someone should :)

PR #58, opened 2019-11-22
Fix alignment/spacing issues with layout

taysea:
This PR addresses the following:
- Makes the hr between content sections align with the right edge of nav items
- Maintains consistent spacing between content sections by adding a bottom margin
- Starts to introduce a mobile responsive format where the sidebar disappears when screen size is `small`. (The navigation does not yet accommodate the sidebar items, but this is a next step)


ShimiSun:
PERFECT!!! 🎉

PR #60, opened 2019-11-25
Add navigation routing - don't merge yet

taysea:
This PR sets up the tracking of which nav link is active. As an overview, aries-core handles the styling of active vs non-active nav items but remains router agnostic. The user can decide any router to use and will provide the aries-core component with the router pathname. Because there are various routing approaches (react-router, nextjs, etc.), aries-core shouldn't have a router within the component library. The user should be able to use any router they want depending on their project structure, and will just provide aries-core with some simple information to allow the routing to be reflected in the styling. Here is how this is done:

1. `Nav` uses `AnchorGroup` by default, so props are placed directly and passed onto `AnchorGroup` under the hood.
1. Creates a prop in aries-core `Nav` called `currentPath` which accepts the current pathname and another prop called `level` which refers to which portion of the path should be used to decide the active nav item. For example, if the path is `/start/about` and the level is specified as `1`, then "start" will be checked as the the item to match as active. If the level is marked as `2`, then "about" would be checked. By default (in the case that the user doesn't specify the level), it will use level `1`.
2. Within `AnchorUndecorated`, the `active` prop is used to decide styling. Right now if a link is active, it receives an underline. However, once the new theme is added this should be changed to match the active styling Chris has provided in his designs.
3. In aries-site `Layout`, we use Nextjs router to access the current pathname of the application and provide that to the aries-core nav.

With these changes, the main and secondary navigation should be appropriately tracking.


ShimiSun:
Have you seen this error:
```
10:16:54 AM: ./src/pages/design/index.js
10:16:54 AM: Module not found: Can't resolve './primer' in '/opt/build/repo/aries-site/src/pages/design'
10:16:54 AM: > Build error occurred
10:16:54 AM: Error: > Build failed because of webpack errors
10:16:54 AM:     at build (/opt/build/repo/node_modules/next/dist/build/index.js:9:900)
10:16:54 AM: error Command failed with exit code 1.
10:16:54 AM: info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
10:16:54 AM: error Command failed with exit code 1.
```

taysea:
Yes, that's one thing I'm looking at. I didn't have that issue on my local build/start so I am checking it out.

taysea:
Closing since the Nav component has been heavily reworked. An approach to how the design system handles different routers should still be considered, but this PR feels too tangential to the discussion now to remain an open PR/

PR #63, opened 2019-11-26
update nav background color

halocline:
**What does this PR do?**

Changes the color of the Nav background to match the design by:

- Modifying `Nav` to accept a `background` prop in `aries-core`
- Specifying a background color on Nav implementation on `Layout.js` on `aries-site`


taysea:
In the future, I assume the aries-core Nav will have a default value that comes from the aries theme? In which case the user wouldn't need to add this background prop onto their app to get the `background-subtle`

halocline:
> In the future, I assume the aries-core Nav will have a default value that comes from the aries theme? In which case the user wouldn't need to add this background prop onto their app to get the `background-subtle`

Yes. I had thought about setting the default now, but I think this is going to fall into one of those grey areas where we get there through experimentation.

PR #64, opened 2019-11-26
anchor styling

halocline:
**What does this PR do?**

- Creates standard execution of anchor styling for anchors appearing in Nav elements or call to actions at the end of content sections:
  - `AnchorCallToAction`
Screen Shot 2019-11-26 at 1 30 42 PM

  - `AnchorNavLink`
Screen Shot 2019-11-26 at 1 31 01 PM
    - Based on the work @taysea is doing, `AnchorNavLink` has styling for when `active` prop is `true`

- Updates usage in `AnchorGroup`, `SideBar`, and `about.js`

**What this PR does not do**
This PR does not apply proper font-size to the `AnchorNavLink`. Design calls for nav links to be 24px. At smaller viewports, the 5 links at 24px become crowded. I think there are a variety of ways to address (call for smaller font-size, adjust sizing based on breakpoints, limit # of elements appearing in NavBar, etc. ... ), but thought it would be a good point of discussion with the larger group.  I can create a branch and demo on Monday call. Created this card in backlog so we can discuss: https://github.com/orgs/hpe-design/projects/1#card-29777469

**Topic for Discussion**
- AnchorUndecorated - With these updates, AnchorUndecorated is no longer currently being applied anywhere, however, I can see us using it in other situations. How do we want to handle?
  - Leave as is, and expect to use it in other situations
  - Remove from code base as it is currently a dead end
  - Keep AnchorUndecorated, but apply styling from either AnchorNavLink or AnchorCallToAction, then remove one of those.


aries-core/src/js/components/helpers/Anchors/AnchorCallToAction.js
```diff
@@ -0,0 +1,35 @@
...
+    <ThemeContext.Extend
+      value={{
+        anchor: {
+          extend: {
+            'font-weight': '700',
+          },
+          textDecoration: 'none',
+        },
+      }}
+    >
```
ShimiSun: Why are we duplicating the AnchorUndecorated instead of trying to reuse it?
If it is because the ` 'font-weight': '700',` then maybe it should be included on AnchorUndecorated, especially if both Sidebar and NavHeader will be using it.

aries-core/src/js/components/helpers/Anchors/AnchorNavLink.js
```diff
@@ -0,0 +1,32 @@
...
+          `,
+          hover: {
+            extend: {
+              opacity: '100%',
+            },
+            textDecoration: 'none',
+          },
+          textDecoration: 'none',
+        },
+      }}
```
ShimiSun: Instead of applying ThemContext every time we use Anchor, maybe it will be a better practice to apply those styling on the theme itself, and remove all usage of ThemeContext

halocline: I think that I am in agreement. It was part of what I was touching on yesterday when mentioning my feeling that the aries theme belongs with aries-core rather than with aries-site. Right now core is consuming grommet-theme-hpe, which is why these helper components are applying ThemeContext. If aries-core consumes the aries theme, then the specification can be done on the theme file rather than in ThemeContext.

taysea:
I am wondering if `AnchorNavLink` could be simplified to `NavLink`. I understand the perspective of trying to maintain some consistency across the different anchor types we are introducing, but to me having the word `Anchor` and `Link` in the component name feels a little redundant.

halocline:
> I am wondering if `AnchorNavLink` could be simplified to `NavLink`. I understand the perspective of trying to maintain some consistency across the different anchor types we are introducing, but to me having the word `Anchor` and `Link` in the component name feels a little redundant.

I'd be fine with either `NavLink` or `AnchorNav`.

aries-core/src/js/components/helpers/Anchors/AnchorCallToAction.js
halocline: Sidebar and NavHeader will have font-weights of 400; the green links with arrows are spec'd at 700.

halocline:
@taysea and @ShimiSun I have incorporated each of your feedback. Please take another look at your convenience.

aries-core/src/js/components/helpers/Anchors/NavLink.js
```diff
@@ -18,7 +18,7 @@ export const AnchorNavLink = ({ ...rest }) => {
             },
             textDecoration: 'none',
```
taysea: Do we still need this `textDecoration: 'none'` even though it's in the theme now?

halocline: This one is at the `hover` level. The textDecoration on the theme is for normal state.

taysea: Gotcha! Makes sense then, I missed that.

aries-core/src/js/components/helpers/Anchors/AnchorNavLink.js
ShimiSun: The theme can be a custom theme part of aries-site.
I think aries-core should be able the render its components no matter what theme it will be given, and besides the HPE theme it shouldn't have a custom theme by itself. Regardless, this styling behavior is specific for aries-site so it should be placed on its theme.

aries-core/src/js/components/helpers/Anchors/NavLink.js
ShimiSun: I think this hover behavior should also be part of the aries-site theme rather then being placed on a ThemeContext component.

aries-core/src/js/components/helpers/Anchors/AnchorCallToAction.js
ShimiSun: I see you removed AnchorUndecorated on your last commit, thanks for the clean-up.
IMHO grommet should allow weight prop so we can avoid this theme context styling, and clean this component out, but it will be a separate effort from this PR.
Regardless, `textDecoration: 'none',` should be moved to the theme instead of being applied on every Anchor/Nav ThemeContext.

aries-core/src/js/components/helpers/Anchors/NavLink.js
halocline: That will apply to all instances of anchor. Is that the desired behavior? For instance, the AnchorCallToAction links, what are their hover states?

halocline:
@ShimiSun @taysea FYI, I played with a couple approaches to eliminate the use ThemeContext from NavLink:

1. Adding a Text component to populate the `label` prop of Anchor. This still required theming, but also added complexity and reduced readability. I ended up not being a fan of this approach.
2. Adding a boolean prop called `navItem` to NavLink, which if present applies styling specified in the theme. Can you take a look and see if you feel better about this approach?

aries-core/src/js/components/helpers/Anchors/NavLink.js
```diff
@@ -1,17 +1,32 @@
-import React from 'react';
+import React, { useState } from 'react';
 import PropTypes from 'prop-types';
-import { Anchor } from 'grommet';
+import { Anchor, Text } from 'grommet';
 
-export const NavLink = ({ size, ...rest }) => {
-  return <Anchor navItem size={size} {...rest} />;
+export const NavLink = ({ active, children, label, size, ...rest }) => {
```
ShimiSun: I don't think you need to pass `children` as a prop while applying `...rest`. That means it can be removed from the PropTypes as well.

halocline: Good catch. children was leftover from previous approach. Fixing.

halocline: We are actually using `children` in the `label` prop, so I think it needs to remain, correct?

ShimiSun: Correct, let's keep it then, thanks!

PR #72, opened 2019-12-02
Fix runtime error from MainContent.js

halocline:
Related Issue: https://github.com/hpe-design/aries/issues/44

**What does this PR do?**
Corrects the runtime error which was generating the warning.

**Work included**
Swapping out Array.map method for React.Children.map() method which React is generating keys.
Reference: https://reactjs.org/docs/react-api.html#reactchildrenmap


ShimiSun:
Thank you!

PR #79, opened 2019-12-05
Set max width on layout

taysea:
Based on design discussions, this sets a max width on the layout of the app so that on large window widths.

### DIFF
--- a/aries-site/src/layouts/main/Layout.js
+++ b/aries-site/src/layouts/main/Layout.js
@@ -45,34 +45,36 @@ export const Layout = ({ children, title }) => {
                 content="width=device-width, initial-scale=1.0"
               />
             </Head>
-            <Nav title="Aries" background="background-subtle">
-              <AnchorGroup
-                items={[
-                  { label: 'Start', href: '/start' },
-                  { label: 'Foundation', href: '/foundation' },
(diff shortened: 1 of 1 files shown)


aries-site/src/layouts/main/Layout.js
```diff
@@ -45,34 +45,36 @@ export const Layout = ({ children, title }) => {
...
-              {size !== 'small' && (
-                <Box fill="vertical">
-                  <SideBar
-                    items={SideBarItemList[selectedNav]}
-                    prefix={selectedNav}
-                  />
-                </Box>
-              )}
-              <Box flex>{mainContent[0]}</Box>
+            <Box width={{ max: 'xlarge' }} margin="auto">
```
ShimiSun: why do you need the extra Box? Nav should be able to accept both `width={{ max: 'xlarge' }} margin="auto"`?

taysea: In this case, I was just wrapping the entire app in a bounding box, but I agree that Nav should also be able to take those props as well. Let's put a hold on this PR though, because I am waiting on a clarification from Chris is the nav background should extend the full width of the window even if the content only spans width of xlarge.

taysea:
Closing because this was taken care of in the Landing Page PR.

PR #80, opened 2019-12-05
Add cypress to site

taysea:
This PR adds cypress to aries-site. For now, I've just added 3 simple tests to make sure that it is working with the site. In the future, we will add more. When you add cypress, it provides some placeholder files for extra customization. I've left them for now but if we don't use them later on, we can always remove them.


ShimiSun:
I think we should remove files that are not functional.
I'm not convinced on the reasoning for using Cypress and what it covers that we can't currently cover with simpler approaches, thoughts?
The reason behind my question is that I'm trying to avoid an overkill for testing framework.

taysea:
@ShimiSun I think the want to use cypress for aries-site is to test routing and deeplinking. I don't think that the unit testing or snapshots that occur with jest/chromatic for aries-core components will be able to test that, but I am open to discussion. I don't want to throw in something that is overkill.

ShimiSun:
Following up our conversation today:
1. Make sure Cypress files are not located on the production directory `dist`/`out`
2. In case it is try to remove its content during the build and try to understand the impact on the bundle in case you can't 100% remove it.
3. Let's give it a try for a couple of weeks and evaluate the value of the framework, I heard lots of great things about it but I'm suspecting it might be an overkill for the site app that doesn't have much content except of routing. 

Note: The hpe-server app is a fantastic candidate for Cypress.

taysea:
Closing because we will be using TestCafe

PR #81, opened 2019-12-05
aries styled button

halocline:
Related Issue: https://github.com/hpe-design/aries/issues/75

**What does this PR do?**

- Creates an Aries styled button for re-use.
- Implements the Aries styled button on `pages/foundation/color.js`
- Adds simple storybook story


taysea:
I think this looks pretty good @halocline. Is the rounding on the aries-site button coming from the theme? or where does that get introduced.

Also, this might be more of a note for Chris, but I'm getting an accessibility violation for the white text on green button. I'm using axe in my developer tools, and it's saying that that color combination doesn't meet WCAG2 AA contrast ratio thresholds:
```
Fix the following:
Element has insufficient color contrast of 2.21 (foreground color: #ffffff, background color: #00c781, font size: 13.5pt (18px), font weight: normal). Expected contrast ratio of 4.5:1
```

aries-site/src/pages/foundation/color.js
```diff
@@ -22,15 +23,11 @@ const Color = () => (
...
         <Button
-          label={
-            <Text
-              color={{ dark: 'text-strong', light: 'text-strong' }}
-              weight={500}
-            >
-              Download Color Swatches Pack
-            </Text>
-          }
+          label="Download Color Swatches Pack"
```
taysea: Getting an accessibility error for this color combination of white text on green background (see my other comment for more details)

halocline:
> I think this looks pretty good @halocline. Is the rounding on the aries-site button coming from the theme? or where does that get introduced.
Correct. Rounding comes from the theme. Will clean up when we move the theme to aries-core and/or update hpe theme. 
> 
> Also, this might be more of a note for Chris, but I'm getting an accessibility violation for the white text on green button. I'm using axe in my developer tools, and it's saying that that color combination doesn't meet WCAG2 AA contrast ratio thresholds:
> 
> ```
> Fix the following:
> Element has insufficient color contrast of 2.21 (foreground color: #ffffff, background color: #00c781, font size: 13.5pt (18px), font weight: normal). Expected contrast ratio of 4.5:1
> ```
Yes, we need to bring that up again. It was an issue in the Cloud Volumes site as well.

aries-site/src/pages/foundation/color.js
halocline: Yes, we need to surface this again.

PR #87, opened 2019-12-06
Wrap main content in Main

taysea:
This uses grommet's `Main` component to wrap the `MainContent`. Now, when you inspect the DOM you will see `<main>` around all of the main content.

### DIFF
--- a/aries-site/src/layouts/main/Layout.js
+++ b/aries-site/src/layouts/main/Layout.js
@@ -1,7 +1,7 @@
 import React from 'react';
 import Head from 'next/head';
 import PropTypes from 'prop-types';
-import { Box, Grommet, ResponsiveContext } from 'grommet';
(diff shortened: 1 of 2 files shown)


aries-site/src/layouts/main/Layout.js
```diff
@@ -72,7 +72,7 @@ export const Layout = ({ children, title }) => {
                   />
                 </Box>
               )}
-              <Box flex>{mainContent[0]}</Box>
+              <Main flex>{mainContent[0]}</Main>
```
ShimiSun: Main is usually flex={false} why do we need to override it?

taysea: The Main is supposed to fill the rest of the width of the window that the sidebar isn't taking up. Flex allows it to expand to fill this width. When I remove it, the section divider ( hr ) doesn't span to the far right.

PR #96, opened 2019-12-11
Changed index of Nav to export from Nav.js

ShimiSun:
I have a separate PR for Tiles/Tile to behave in the same way.
As the project scale, it will be beneficial for us to have explicit filenames when browsing between components instead of all components to be listed on index.js files.


halocline:
> As the project scale, it will be beneficial for us to have explicit filenames when browsing between components instead of all components to be listed on index.js files.

Agree. Good modification. Thank you.

PR #101, opened 2019-12-11
Make subsection text sizes dynamic for h1, h2, h3, etc.

taysea:
This PR sets up the structure that pages are made up of ContentSections which contain SubSections as children. A prop of `level` can be applied to Subsections. This level determines two things:
1. The kind of heading should be used. If level = 3, then the heading will be rendered as an h3. etc.
2. The size of the text that appears within the Subsection. Text that accompanies an h1 is large, text with an h2 is medium, and so on. I have introduced a new component called `BodyText` which will take the place of the places we were using `MainDescription` and `Text`. Instead, any text blocks inside Subsections should now use `BodyText`. The Subsection is designed to automatically pass along the level prop to its children. That means that without any modifications, BodyText will render its text as the appropriate size depending on if it is the text that accompanies an h1, h2, h3, etc.

I have demonstrated the implementation of these changes on the Colors page. For clarity of reviewing the Subsection changes, I have left MainHeading and MainDescription on the colors page, but after this merges we can refactor that to use Subsection level=1 instead. I think this implementation will help ensure that the Heading/Text size pairings are tightly coupled and will save head space because the `level` only needs to applied once on the Subsection.


aries-site/src/components/content/BodyText.js
```diff
@@ -0,0 +1,28 @@
...
+  1: 'large',
+  2: 'medium', // default font size
+  3: 'small',
+};
+
+export const BodyText = ({ children, level, ...rest }) => {
+  return (
+    <Text size={TEXT_SIZE[level]} {...rest}>
+      {children}
+    </Text>
```
halocline: General question: Are there any tradeoffs/benefits of using <Text /> vs. <Paragraph />? Text renders as <span /> and  Paragraph as <p /> correct?

aries-site/src/components/content/BodyText.js
```diff
@@ -0,0 +1,28 @@
...
+  );
+};
+
+BodyText.propTypes = {
+  children: PropTypes.string.isRequired,
+  level: PropTypes.number,
+};
+
+BodyText.defaultProps = {
+  level: 2,
```
halocline: When refactoring to apply to MainHeading and MainDescription, will this then default to 1?

aries-site/src/layouts/content/Subsection.js
```diff
@@ -22,22 +29,50 @@ export const Subsection = ({ children, name, ...rest }) => {
...
 Subsection.propTypes = {
   children: PropTypes.oneOfType([PropTypes.element, PropTypes.array])
     .isRequired,
-  headingSize: PropTypes.string,
+  level: PropTypes.number,
   name: PropTypes.string.isRequired,
 };
+
+Subsection.defaultProps = {
+  level: 2,
```
halocline: When refactoring to apply to MainHeading and MainDescription, will this then default to 1?

aries-site/src/pages/foundation/color.js
```diff
@@ -34,14 +41,23 @@ const Color = () => (
...
-      <Subsection name="Primary" headingSize="small">
-        <Text>
+      <Subsection name="Color Palettes">
+        <BodyText>
+          Color is a key way we express our brand. Using color on interface
+          elements, how typography is presented, displaying data visualizations,
+          and adjusting context of the experience to light and dark are just
+          some of the ways we illustrate our brand.
+        </BodyText>
+        <AnchorCallToAction label="Color Principles" />
```
halocline: When tabbing through, this Anchor is unreachable? Is that the behavior you get as well?

aries-site/src/components/content/BodyText.js
taysea: Paragraph will automatically set a max-width depending on the font size. I am going to change BodyText to use Paragraph because these are paragraphs more than they are just text. However, for now I will apply the `fill` prop to the paragraph so that the paragraph can expand to the width of the box.

aries-site/src/components/content/BodyText.js
taysea: I think we should discuss what we want the default to be. In terms of frequency of appearance, I think h3 occurs most so if we wanted to save time, then maybe defaulting to level: 3 would be beneficial? But in terms of clarity, it may be easier to remember that level 1 is default and anything beyond that needs specification.

aries-site/src/layouts/content/Subsection.js
taysea: See my above response on BodyText

aries-site/src/components/content/BodyText.js
halocline: I think I agree with the "clarity" approach.

aries-site/src/pages/foundation/color.js
taysea: You're right. Let me look into why that is.

taysea: The reason it is unreachable right now is because it has no href. Once we add an href to it, it becomes part of the tab flow.

aries-site/src/components/content/BodyText.js
```diff
@@ -0,0 +1,30 @@
...
+
+// Body text size should be based on if its parent heading is an
+// h2, h3, etc.
+const TEXT_SIZE = {
+  1: 'large',
+  2: 'medium', // default font size
+  3: 'small',
+};
+
+export const BodyText = ({ children, level, size, ...rest }) => {
```
ShimiSun: I don't think you need to use `children` if you use `...rest`.

aries-site/src/components/content/BodyText.js
```diff
@@ -0,0 +1,30 @@
...
+  return (
+    <Paragraph size={size || TEXT_SIZE[level]} fill margin="none" {...rest}>
+      {children}
+    </Paragraph>
+  );
+};
+
+BodyText.propTypes = {
+  children: PropTypes.string.isRequired,
```
ShimiSun: could be removed as well.

ShimiSun: Regardless, I don't think it should be limited to a string. (it could be a react Node).

aries-site/src/components/content/BodyText.js
```diff
@@ -0,0 +1,30 @@
...
+// h2, h3, etc.
+const TEXT_SIZE = {
+  1: 'large',
+  2: 'medium', // default font size
+  3: 'small',
+};
+
+export const BodyText = ({ children, level, size, ...rest }) => {
+  return (
+    <Paragraph size={size || TEXT_SIZE[level]} fill margin="none" {...rest}>
```
ShimiSun: love to see you are using the `fill` :) (It's a new prop for Paragraph)

aries-site/src/layouts/content/Subsection.js
```diff
@@ -22,22 +29,50 @@ export const Subsection = ({ children, name, ...rest }) => {
...
-          a11yTitle={`Jump to section titled ${name}`}
-          href={`#${id}`}
-          icon={<LinkIcon color={over ? 'text-xweak' : 'transparent'} />}
-        />
+      {/* This condition for gap is needed because the link icon has padding
+       * to maintain a big enough touch area for accessibility. This creates
+       * a larger than desired gap between h3's and its first child. This
+       * removes that extra space.
+       */}
+      <Box gap={GAP_SIZE[level]}>
```
ShimiSun: The context of `level` is different than `size`, level reminds me of Heading `level`, was it the purpose?

aries-site/src/layouts/content/Subsection.js
```diff
@@ -22,22 +29,50 @@ export const Subsection = ({ children, name, ...rest }) => {
...
+        <Box direction="row" justify="between" align="center">
```
ShimiSun: I think this is a good candidate for the Header component. Header has `direction="row" justify="between" align="center"` embedded in it.

aries-site/src/layouts/content/Subsection.js
```diff
@@ -22,22 +29,50 @@ export const Subsection = ({ children, name, ...rest }) => {
...
       </Box>
-      {children}
+      {React.Children.map(children, (child, index) => {
+        if (index === 0) {
+          return undefined;
+        }
+        return React.cloneElement(child, {
+          level,
+        });
+      })}
```
ShimiSun: Aren't those two blocks duplicated? can we extract it to a function to avoid duplication?

aries-site/src/layouts/content/Subsection.js
```diff
@@ -4,7 +4,13 @@ import { Anchor, Box } from 'grommet';
 import { Link as LinkIcon } from 'grommet-icons';
 import { Subheading } from '../../components';
 
-export const Subsection = ({ children, name, ...rest }) => {
+const GAP_SIZE = {
+  1: 'medium',
+  2: 'small',
+  3: undefined,
```
ShimiSun: I might be missing something, but Gap should except more than two values.

aries-site/src/components/content/BodyText.js
taysea: The children would be what is inside the paragraph, but the `...rest` would be if we wanted to allow customization of other paragraph props. If I remove `children` I get an error.

aries-site/src/layouts/content/Subsection.js
taysea: Yes, the purpose is that Subsection receives a `level` (1, 2, 3, etc.) which will determine if the headings within that subsection are an h1, h2, h3, etc. 

The comment above line 37 is meant to explain why the gap needs to be adjusted based on the heading level. If it's not clear, I can try to reword.

aries-site/src/layouts/content/Subsection.js
taysea: It does sound like a good usage, let me try that out!

aries-site/src/layouts/content/Subsection.js
taysea: This gap formatting affects line 37. It helps to ensure that the spacing between the heading and the first element that follows it has the appropriate spacing. The comment on line 32 aims to explain why an h3 should not have any added gap.

aries-site/src/components/content/BodyText.js
ShimiSun: Or was it on purpose because the child will go to Paragraph?

aries-site/src/layouts/content/Subsection.js
taysea: They're not quite duplicated. The purpose of the first block is to isolate the first child. The first child following the header needs to have the right amount of spacing. In the case where it is an h3, we don't want any gap between the heading and following child because the intrinsic height of the anchor link is creating enough space. The first block renders only the first child. The second block renders everything else.

ShimiSun: Adding these explanations to the code could be helpful.

aries-site/src/components/content/BodyText.js
taysea: Yes it was on purpose that the child will go to Paragraph. The BodyText component would be for any text blocks within a subsection. The idea behind creating a specialized component for it is that when BodyText is placed within a Subsection, the level on the Subsection will determine what size text to use. So, if a BodyText is used within a Subsection with level 2, then the paragraph size will be the text size meant to be paired with h2.

taysea: Based on Chris's designs, what goes inside of BodyText would be a block of text.

aries-site/src/components/content/BodyText.js
ShimiSun: I think you were able to figut=re it out already, marking it as resolved.

PR #102, opened 2019-12-11
Tiles & Tile components Incorporated with Header & Footer

### LINKED ISSUE #61: Tiles Component

ShimiSun:
* Added new Tiles component implemented with Grid.
* Refine existing Tile component to use Grommet Header.
* Incorporated examples with Header & Footer.
* Introducing best practice for Tile while consuming it as a child for Tiles. (moving Tile stories to Tiles)
Closes #61


ShimiSun:
Something is up with the yarn hashing, I'll take a look...

aries-core/src/js/components/core/Tiles/stories/tileHeader.js
```diff
@@ -0,0 +1,43 @@
...
+        </Header>
+        <Box pad="small">
+          <Paragraph>
+            This custome Tile includes custom properties of: height, width &
+            Heading. This Tile is built with the Grommet components of Header &
+            Footer.
+          </Paragraph>
+        </Box>
+        <Box flex />
+        <Footer pad="small" background="light-4">
```
taysea: Seems like by default, a Tile has rounded corners. With the Footer added, the rounding doesn't occur anymore on the bottom corners but it does still occur on the top corners. Do we want the rounding to be maintained by default here?

aries-core/src/js/components/core/Tiles/stories/simple.js
```diff
@@ -0,0 +1,45 @@
...
+};
+
+const content = [
+  'Shimi dev - Shimi Shimi Coco Puff.  ',
+  'Taytay dev - All tiles in the Grid container are symetric.',
+  'Gliss dev - Responsive works better with Grid.',
+];
+
+const TileCollection = () =>
+  [0, 1, 2].map(index => (
```
taysea:
```suggestion
  content.map((item, index) => (
```

aries-core/src/js/components/core/Tiles/stories/simple.js
```diff
@@ -0,0 +1,45 @@
...
+    <Tile key={index} pad heading={`Tile ${index + 1}`}>
+      {content[index]}
```
taysea:
```suggestion
      {item}
```

aries-core/src/js/components/core/Tiles/stories/simple.js
taysea: Thinking you can simplify the mapping here so if more content were added, you wouldn't need to go back and adjust the array of [0, 1, 2]

aries-core/src/js/components/core/Tiles/Tiles.js
```diff
@@ -0,0 +1,7 @@
+import React from 'react';
+
+import { Grid } from 'grommet';
+
+export const Tiles = ({ ...rest }) => (
+  <Grid columns={{ count: 'fit', size: 'small' }} {...rest} />
```
taysea: Just out of curiosity, what is the reason that we chose this as the default size for the columns?

aries-core/src/js/components/core/Tiles/stories/tileHeader.js
ShimiSun: This is just a story example, so we can go either way. 
Regarding:
>With the Footer added, the rounding doesn't occur anymore on the bottom corners
This statement is only true if you are using a background color for the footer.

aries-core/src/js/components/core/Tiles/stories/simple.js
ShimiSun: I would have agreed if this content is used more then once. Not sure it is critical in this case.

aries-core/src/js/components/core/Tiles/Tiles.js
ShimiSun: Chris's designs.

aries-core/src/js/components/core/Tiles/stories/simple.js
ShimiSun: Done.

aries-core/src/js/components/core/Tiles/stories/simple.js
ShimiSun: Done.

aries-core/src/js/components/core/Tiles/stories/tileHeader.js
taysea: Gotcha, what if I want to have a background color but maintain the rounded bottom? Will that be supported

ShimiSun: of course, set `round` on Footer. https://v2.grommet.io/box#round

ericsoderberghp: I would think setting `round` on a Tile would set `overflow="hidden"`, so that any enclosed content wouldn't exceed the Tile.

ShimiSun: I looked at the designs, and it actually makes sense. Let me come up with a quick PR for Grommet Footer.

ShimiSun: @ericsoderberghp @taysea , something like this https://github.com/grommet/grommet/pull/3617/?

ericsoderberghp: I think it belongs on Tile, not on Footer.

ShimiSun: @ericsoderberghp, Not sure what you mean, Tile doesn't have a Footer component.
The thing we can do for this story example is to add `round` to Footer as suggested on top.

ShimiSun: I pushed an example of how this could be achieved via Footer on the Tile.

aries-core/src/js/components/core/Tile/Tile.js
```diff
@@ -0,0 +1,62 @@
...
+import { Box, Header, Heading } from 'grommet';
+
+const PAD_SIZES = ['xxsmall', 'xsmall', 'small', 'medium', 'large', 'xlarge'];
+
+export const Tile = ({ children, heading, pad, ...rest }) => {
+  const tilePad =
+    pad === true ? { horizontal: 'small', top: 'small', bottom: 'small' } : pad;
+
+  return (
+    <Box elevation="medium" round="small" {...rest}>
```
ericsoderberghp: Because the Tile is setting `round="small"`, I think it should also set `overflow="hidden"` to prevent children from exceeding the rounded corners.

aries-core/src/js/components/core/Tiles/stories/tileHeader.js
```diff
@@ -0,0 +1,48 @@
...
+            This custome Tile includes custom properties of: height, width &
+            Heading. This Tile is built with the Grommet components of Header &
+            Footer.
+          </Paragraph>
+        </Box>
+        <Box flex />
+        <Footer
+          pad="medium"
+          background="light-4"
+          round={{ size: 'small', corner: 'bottom' }}
```
ericsoderberghp: If the Tile has `overflow="hidden"`, you don't need to have the Footer match the rounding of the Tile.

ShimiSun: Correct, this line is cleaned out on my last commit.

aries-core/src/js/components/core/Tile/Tile.js
ShimiSun: Done. Thanks for elaborating.

PR #106, opened 2019-12-12
updating NavLink styling

halocline:
Related Issues: https://github.com/hpe-design/aries/issues/93

Work Included:
- Updated NavLink styling to reflect latest designs

Screen Shot 2019-12-12 at 8 53 16 AM

### DIFF
--- a/aries-core/src/js/components/helpers/Anchors/NavLink.js
+++ b/aries-core/src/js/components/helpers/Anchors/NavLink.js
@@ -6,7 +6,7 @@ export const NavLink = ({ active, children, label, size, ...rest }) => {
   const textColor = 'text-strong';
   const activeFontWeight = 700;
   const defaultFontWeight = 400;
-  const [fontWeight, setFontWeight] = useState(defaultFontWeight);
+  const [fontWeight, setFontWeight] = useState(active ? activeFontWeight : defaultFontWeight);
 
   return (
     <Anchor
--- a/aries-core/src/js/components/helpers/Anchors/NavLink.js
+++ b/aries-core/src/js/components/helpers/Anchors/NavLink.js
@@ -3,20 +3,23 @@ import PropTypes from 'prop-types';
 import { Anchor, Text } from 'grommet';
 
 export const NavLink = ({ active, children, label, size, ...rest }) => {
-  const activeTextColor = active ? 'text' : null;
-  const defaultTextColor = 'text-xweak';
-  const fontWeight = 400;
-  const [textColor, setTextColor] = useState(defaultTextColor);
+  const textColor = 'text-strong';
+  const activeFontWeight = 700;
+  const defaultFontWeight = 400;
+  const [fontWeight, setFontWeight] = useState(defaultFontWeight);
 
   return (
     <Anchor
-      color={activeTextColor || textColor}
-      label={<Text weight={fontWeight}>{label || children}</Text>}
-      onBlur={() => setTextColor(defaultTextColor)}
-      onFocus={() => setTextColor(activeTextColor)}
-      onMouseOut={() => setTextColor(defaultTextColor)}
-      onMouseOver={() => setTextColor(activeTextColor)}
-      size={size}
+      color={textColor}
+      label={
+        <Text size={size} weight={fontWeight}>
+          {label || children}
+        </Text>
+      }
+      onBlur={() => setFontWeight(defaultFontWeight)}
+      onFocus={() => setFontWeight(activeFontWeight)}
+      onMouseOut={() => setFontWeight(defaultFontWeight)}
+      onMouseOver={() => setFontWeight(activeFontWeight)}
       {...rest}
     />
   );
--- a/yarn.lock
+++ b/yarn.lock
@@ -2603,12 +2603,12 @@ array-flatten@^2.1.0:
   integrity sha512-hNfzcOV8W4NdualtqBFPyVO+54DSJuZGY9qT4pRroB6S9e3iiido2ISIC5h9R2sPJ8H3FHCIiEnsv1lPXO3KtQ==
 
 array-includes@^3.0.3:
-  version "3.0.3"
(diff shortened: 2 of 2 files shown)


aries-core/src/js/components/helpers/Anchors/NavLink.js
```diff
@@ -3,20 +3,23 @@ import PropTypes from 'prop-types';
...
 
 export const NavLink = ({ active, children, label, size, ...rest }) => {
-  const activeTextColor = active ? 'text' : null;
-  const defaultTextColor = 'text-xweak';
-  const fontWeight = 400;
-  const [textColor, setTextColor] = useState(defaultTextColor);
+  const textColor = 'text-strong';
+  const activeFontWeight = 700;
+  const defaultFontWeight = 400;
+  const [fontWeight, setFontWeight] = useState(defaultFontWeight);
```
taysea:
```suggestion
  const [fontWeight, setFontWeight] = useState(active ? activeFontWeight : defaultFontWeight);
```
If the link is active, it should start at the activeFontWeight.

aries-core/src/js/components/helpers/Anchors/NavLink.js
```diff
@@ -3,20 +3,23 @@ import PropTypes from 'prop-types';
...
+      color={textColor}
+      label={
+        <Text size={size} weight={fontWeight}>
+          {label || children}
+        </Text>
+      }
+      onBlur={() => setFontWeight(defaultFontWeight)}
+      onFocus={() => setFontWeight(activeFontWeight)}
+      onMouseOut={() => setFontWeight(defaultFontWeight)}
+      onMouseOver={() => setFontWeight(activeFontWeight)}
```
ShimiSun: You removed `size={size}` for Anchor, I guess it is because you are assuming it will pass with the spread notation ('...rest')? In this case you need to remove the deconstruction of `size` from NavLink (line #5), otherwise, the size will always be undefined.

aries-core/src/js/components/helpers/Anchors/NavLink.js
```diff
@@ -3,20 +3,23 @@ import PropTypes from 'prop-types';
 import { Anchor, Text } from 'grommet';
 
 export const NavLink = ({ active, children, label, size, ...rest }) => {
-  const activeTextColor = active ? 'text' : null;
-  const defaultTextColor = 'text-xweak';
-  const fontWeight = 400;
-  const [textColor, setTextColor] = useState(defaultTextColor);
+  const textColor = 'text-strong';
+  const activeFontWeight = 700;
+  const defaultFontWeight = 400;
```
ShimiSun: Wondering if those should be hardcoded or part of the theme.

aries-core/src/js/components/helpers/Anchors/NavLink.js
halocline: I actually moved size from <Anchor /> to instead be applied to the <Text /> in the Anchor's label prop, otherwise the size wasn't flowing through as expected.

aries-core/src/js/components/helpers/Anchors/NavLink.js
halocline: done.

halocline:
> Left one comment. I also noticed that when I mouse over an element in the main nav, when the item bolds, it expands its width so the item to its left gets pushed over a little bit. I think that since we're removing the top navigation in the new designs, this won't be an issue for aries-site but maybe would cause problems in other apps.

I noticed that as well. Was wondering where the responsibility of the spacing belongs... in the NavLink's padding? Or the container holding the NavLinks?

aries-core/src/js/components/helpers/Anchors/NavLink.js
halocline: I've gone back and forth, debating the approach, and but have been leaning towards more tightly coupling the style to the component (because we have 3 distinct executions) and use the theme for the basic, shared styling.

**NavLink**
Screen Shot 2019-12-12 at 11 38 32 AM

**AnchorCallToAction**
Screen Shot 2019-12-12 at 11 38 27 AM

**"Plain" Anchor** 
Screen Shot 2019-12-12 at 11 38 10 AM

I think the alternative (putting the styling in the theme) will require us to use anchor.extend and then style based off of a prop. When exploring this option, things start to get messy and less readable.

I landed on this approach because I felt the theme should contain our "base" or "default" anchor, then the specialized use cases can modify from there.

ShimiSun: Thank you for our great offline discussion, let's continue exploring the option of allowing `type` to Anchor, while type could be any of the above Anchor options above. It will make the user more resourceful while exploring the different Anchor options.

aries-core/src/js/components/helpers/Anchors/NavLink.js
ShimiSun: Nice

taysea:
> > Left one comment. I also noticed that when I mouse over an element in the main nav, when the item bolds, it expands its width so the item to its left gets pushed over a little bit. I think that since we're removing the top navigation in the new designs, this won't be an issue for aries-site but maybe would cause problems in other apps.
> 
> I noticed that as well. Was wondering where the responsibility of the spacing belongs... in the NavLink's padding? Or the container holding the NavLinks?

Good question. Maybe the container holding them would make more sense.

aries-core/src/js/components/helpers/Anchors/NavLink.js
halocline: For documentation purposes, what we discussed was the following:

- As the design system instructions get fleshed out regarding Anchor usage and behavior (i.e. which use cases should use which type of anchor), consider creating an Anchor component which has a `type` property whose values might be (callToAction, navLink, plain, etc.)
- This would allow us to present a single, consolidated component to end users, allowing them to easily select the appropriate execution (as opposed to having to browse through all potential anchors).

Issue #108, opened 2019-12-14
Nav - Refactor to new designs, leverage new grommet component and decouple from aries-site Header


ShimiSun:
The decoupling of Nav from Header was addressed on recent PR, but there is still more work refactoring to address the new designs.

MikeWalrath:
Is there any additional work here?

taysea:
The decoupling is done. In terms of what "Nav" is, we are not currently using Nav in aries-site at all. Based on a previous server call, it seems like "Nav" would be a grommet component that only could contain anchors or buttons potentially?

I think in terms of immediate tasks on this, they've been completed and this can be closed.

What remains as potentially follow-up would be discussions about how this component might be used in Server app, if it's just imported from grommet, etc.

ShimiSun:
I think this issue is more about the Nav component on aries-core. The Nav on the core library isn't using the Grommet Nav, and it performs as a NavBar rather than a Nav so renaming should be considered.

halocline:
Has been implemented in Carte version of site.

Issue #65, opened 2019-11-27
Home Page


halocline:
<img width="1213" alt="Screen Shot 2019-12-13 at 1 19 07 PM" src="https://user-images.githubusercontent.com/1756948/70829510-31c79d00-1dab-11ea-98cc-5fb9355ff358.png">

ShimiSun:
Does it mean the landing page? do we have designs for it?

halocline:
Yes, home = landing page.
There is an early design we could work from. Work would include:
- [ ] layout w/ no sidebar
- [ ] content section with background image
- [ ] content foreground image
- [ ] copy
- [ ] button routed to /start

![Screen Shot 2019-11-27 at 11 14 31 AM](https://user-images.githubusercontent.com/1756948/69749340-8b6f6c80-1107-11ea-884c-faf89e79edf8.png)

PR #109, opened 2019-12-14
Setting ground for Landing Page by decoupling Layout/Nav/Head/Header/Sidebar components

ShimiSun:
At the moment Layout is the 'goto' component for each page since it includes the Grommet component, the main Layout, and the theme.
While starting to work on the Landing page, I noticed multiple components that are fundamentally part of Layout yet should-not be part of the main infrastructure.
This PR is attempting to decouple those components, so Layout could be used generically for the Landing page as well as the navigational pages. 
Another issue I've encountered is that the main Header & Nav are not decoupled, which is necessary for the landing page, so this PR is addressing it as well.

This PR can be the start point of further refactoring ideas and clean-ups, especially with the new designs, but it is necessary for the infrastructure of the landing page.
For example (just an idea), I think that now that Layout is leaner, we can move the content of PageLayout inside Layout and clean out the PageLayout component. I didn't attempt it in this PR because it was growing big already and there is more to come on the landing page itself.

Related to #65 #108


aries-site/src/layouts/main/Layout.js
```diff
@@ -1,91 +1,28 @@
...
-                </Box>
-              )}
-              <Main flex>{mainContent[0]}</Main>
-            </Box>
-          </>
-        )}
-      </ResponsiveContext.Consumer>
+      <Head title={title} />
+      <Header showLinks={!isLanding} />
+      {!isLanding ? <NavLayout mainContentChilds={children} /> : children}
```
taysea:
```suggestion
      {!isLanding ? <NavLayout mainContentChildren={children} /> : children}
```

aries-site/src/layouts/main/NavLayout.js
```diff
@@ -0,0 +1,60 @@
...
+    console.warn(
+      `Expected a single ${type}, received ${filteredChildren.length}.`,
+      `Only first ${type} element will be rendered.`,
+    );
+  }
+
+  return filteredChildren;
+};
+
+export const NavLayout = ({ mainContentChilds }) => {
```
taysea:
```suggestion
export const NavLayout = ({ mainContentChildren }) => {
```

aries-site/src/layouts/main/NavLayout.js
```diff
@@ -0,0 +1,60 @@
...
+  // Only expect a single child of the following types
+  const mainContent = filterChildren(mainContentChilds, 'MainContent');
```
taysea:
```suggestion
  const mainContent = filterChildren(mainContentChildren, 'MainContent');
```

aries-site/src/pages/index.js
```diff
@@ -3,6 +3,7 @@ import { initialize, pageview } from 'react-ga';
 import { Config } from '../../config';
 
 import About from './start/about';
+// import Landing from './landing/landing';
```
taysea: Is there a reason we don't want to just place the code for the index page in index.js? By creating a file in ./landing/landing, we are creating a route of aries.hpe.design/landing/landing which I don't think is a route we want people to visit. If we do want to keep the landing page code in a separate file for clarity in the file name, let's make sure to add the canonical link in the SEO that @halocline had mentioned.

aries-site/src/layouts/main/Layout.js
```diff
@@ -1,91 +1,28 @@
...
-export const Layout = ({ children, title }) => {
-  // Only expect a single child of the following types
-  const mainContent = filterChildren(children, 'MainContent');
-
-  // TODO selectedNav should be retrived from aries-core
-  // as the selected element of the NavBar
-  const selectedNav = 'start';
+import { Header, Head, NavLayout } from '..';
 
+export const Layout = ({ children, title, isLanding }) => {
```
halocline: Suggestion: rename `isLanding` to `showLinks`, keep the same functionality, but is direct to the functionality and less concerned about how it is being used.

aries-site/src/layouts/main/NavLayout.js
```diff
@@ -0,0 +1,60 @@
...
+          >
+            {size !== 'small' && (
+              <Box fill="vertical">
+                <SideBar
+                  items={SideBarItemList[selectedNav]}
+                  prefix={selectedNav}
+                />
+              </Box>
+            )}
+            <Main flex>{mainContent[0]}</Main>
```
taysea: I think the naming of 'NavLayout' made me think it was the layout of just the nav bar, not the pages with navigation so I got confused why 'Main' would be in this file. It makes sense now, but I'm wondering if there is a clearer name or a comment we could add to give context.

halocline: Having `NavLayout` controlling the render of the `<Main />` seems unintuitive to me (i.e. I wouldn't think to inspect `NavLayout` to find `<Main />`

aries-site/src/layouts/main/Layout.js
ShimiSun: The context of `showLinks` is too specific for line #14.

aries-site/src/pages/index.js
ShimiSun: I agree. I'll make the change.

ShimiSun: Done.

aries-site/src/layouts/main/NavLayout.js
ShimiSun: In general, I think that the name Nav is confusing because the Header doesn't even have Nav on the new designs.
But I'm opened to other name suggestions... I'll take it on slack.

ShimiSun: Same as the question above 👆 , taking it on slack.

aries-site/src/pages/landing/index.js
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
```
taysea: This will still be creating a route at aries.hpe.design/landing, but the actual route should just be '/'. If we don't want an unused route, then this code should all be placed in the root index.js file. If not, we should make sure to add the canonical url to the head.

ShimiSun: @taysea , I agree, I'll fix it on the next PR that will actually address the Landing page implementation.
For the moment this change won't have any impact because we are not directing to the Landing page in any way.

PR #112, opened 2019-12-16
Adding UsageExample component

halocline:
Related Issues: https://github.com/hpe-design/aries/issues/94

**What does this PR do?**
Because Aries site will demonstrate proper usage of design elements and present their usage in both light and dark theme contexts, this PR creates a `<UsageExample />` component for re-use.

`<UsageExample />` 
- Takes two primary props: 
  1) `label` which is the label associated with the demonstrated usage
  2) `themeMode` which controls whether the example is presented in light or dark mode
- Accepts and renders any children passed to it
- Any other Box props

Other work included:

- For demonstration purposes, implements `<UsageExample />` on `branding.js` and `color.js` pages.
- Minor update to `Layout.js` background color, setting background to be `background-back` which applies the correct light and dark backgrounds.
- Minor update to `Subsection.js`, setting `fill="horizontal"` 

**Interface**
Screen Shot 2019-12-16 at 9 45 34 AM

**Examples**
Screen Shot 2019-12-16 at 9 32 26 AM
Screen Shot 2019-12-16 at 9 31 43 AM
Screen Shot 2019-12-16 at 9 32 18 AM
Screen Shot 2019-12-16 at 9 31 52 AM


aries-site/src/pages/foundation/branding.js
```diff
@@ -10,6 +17,36 @@ const Branding = () => (
...
+          themeMode="light"
+          pad={{ horizontal: 'large', vertical: 'small' }}
+        >
+          <Box direction="row" align="center" gap="medium">
+            <Hpe color="brand" size="xlarge" />
+            <Box direction="row" gap="xsmall">
+              <Text weight="bold">HPE</Text>
+              <Text>Service Name</Text>
+            </Box>
+          </Box>
```
ShimiSun: I'm thinking this behavior is duplicated and can be its own separate component. (see lines 40-46)

aries-site/src/pages/foundation/color.js
```diff
@@ -1,18 +1,44 @@
...
+  textColors,
+} from '../../data';
 
 const title = 'Color';
 
+const TextExample = ({ color, hex }) => {
+  return (
+    <Box align="center" margin={{ horizontal: 'small' }}>
+      <Box direction="row" align="center">
+        <Text color={color} weight={700} size="84px">
```
ShimiSun: The custom sizing and weight doesn't feel right, is it necessary?

halocline: I think this execution is limited and specific to this particular use case and will not be reused elsewhere. The alternative would be to add an `xxlarge` prop for `Text` on the theme, but this execution would still need to overwrite line-height and font-weight.

aries-site/src/pages/foundation/branding.js
halocline: Yes, it is duplicated. I don't think it will be used other than these two instances, but I can add it as a component local to `branding.js`.

aries-site/src/pages/foundation/color.js
ShimiSun: @ericsoderberghp would love to hear your suggestion here.

aries-site/src/pages/foundation/branding.js
halocline: Done.

aries-site/src/pages/foundation/color.js
```diff
@@ -78,8 +104,68 @@ const Color = () => (
           </Box>
         )}
       </Subsection>
+      <Subsection name="Text Colors">
+        <Text>
+          Text needs to be legible in all contexts. Ensuring high contrast is
+          met for accessibility and readability is important to delivering users
+          a consistent reading experience throughout HPE.
+        </Text>
+        <AnchorCallToAction label="Typography Elements" href="#" />
```
taysea:
```suggestion
        <AnchorCallToAction label="Typography Elements" href="#" />
        </Subsection>
```
I think the Subsection should close here and the next Subsection follows it. The spacing for between Subsections is set up for them to be siblings as opposed to children/parents of each other.

aries-site/src/pages/foundation/color.js
```diff
@@ -78,8 +104,68 @@ const Color = () => (
...
+                  <TextExample
+                    key={color.name}
+                    color={color.name}
+                    hex={color.hex.dark}
+                  />
+                );
+              })}
+          </UsageExample>
+        </Subsection>
+      </Subsection>
```
taysea:
```suggestion
```
Removing this extra closing tag regarding my comment above.

aries-site/src/pages/foundation/color.js
halocline: Mmmm... interesting. I think when I was writing this, I was imagining the parent `<Subsection />` to be a level 2 section (i.e. H2), then the inner `<Subsection />` being a level 3 section since that content is supporting and directly related to its parent. In this particular execution, the level 3 "Text Color Palette" would be accompanied by a sibling level 3 "Call To Action Text".

To me the supporting content feels like it should be children of the content it is supporting. Happy to adjust, but curious on your thoughts here.

taysea: That makes sense as well, we would just need to think about how to make sure spacing is always consistent. The reason I thought they could just be siblings is because in html you could have h2, h3, etc at the same level as each other without needing to nest it inside of another div. I am open to allowing subsections to nest inside each other, but we would need to make sure spacing remains consistent.
```
<div>
<h2>Hello World</h2>
<p>This is my parapgraphy.</p>
<h3>Somthing else</h3>
</div>
...
```

aries-site/src/pages/foundation/color.js
ericsoderberghp: I've wanted a larger size on Text as well. @L0ZZI and I discussed last week and he was going to work on that. So, I think we need some larger text size options for grommet in general.

ShimiSun: Sounds good, I filed a separate issue on Grommet, we can come back and refine when more sizes are supported.

PR #115, opened 2019-12-17
Introduce SubmitFeedback component for bottom of content pages

taysea:
This PR creates a component for the submit feedback portion of the content pages (excluding landing pages and the navigation pages) and adds it as the final child to the `MainContent` component. This allows it to be rendered beneath all of the docs pages.

Screen Shot 2019-12-16 at 4 14 14 PM


taysea:
Linking to #116

aries-core/src/js/components/helpers/Anchors/Anchor.js
```diff
@@ -0,0 +1,36 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Anchor as GrommetAnchor, Box, Text } from 'grommet';
+
+export const Anchor = ({ label, size, weight, ...rest }) => {
```
halocline: For documentation purposes, as [discussed previously](https://github.com/hpe-design/aries/pull/106#discussion_r357422499), we plan to refactor Aries' anchor component with a `type` property once design executions for anchors are fleshed out further.

aries-site/src/layouts/content/MainContent.js
```diff
@@ -1,19 +1,12 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+import { SubmitFeedback } from '../../components/content';
 
 export const MainContent = ({ children }) => {
   return (
     <>
-      {children &&
```
halocline: Good clean up!

aries-core/src/js/components/helpers/Anchors/Anchor.js
```diff
@@ -0,0 +1,36 @@
...
+          </Text>
+        }
+        {...rest}
+      />
+    </Box>
+  );
+};
+
+Anchor.propTypes = {
+  label: PropTypes.string,
```
halocline: In grommet, in addition to a string we can pass `Anchor` a `label` of a JSX element, should we make that consistent here?

taysea: The reason I had only put string is because I have embedded the label within the `<Text />` so that we could control the weight on the text which is something that can't be done directly on anchor. In that case, the label you are passing to it should just be a string because it will be nested inside of Text. However, I can pull the `<Text />` block out and apply it directly on my use of Anchor in `SubmitFeedback`. If we plan to make this Anchor more general, then that would be the right move. I think my approach here aligned with my thought that this was somewhat opinionated for the use case of this kind of anchor with underline.

halocline: Sounds good. Makes sense to me.

aries-site/src/layouts/content/MainContent.js
```diff
@@ -1,19 +1,12 @@
...
-          ? React.Children.map(children, (child, index) => {
-              return React.cloneElement(child, {
-                lastSection: index === children.length - 1,
-              });
-            })
-          : React.cloneElement(children, {
-              lastSection: true,
-            }))}
+      {children}
+      <SubmitFeedback />
```
ShimiSun: @taysea I think this is the reason we are seeing the component on the landing page.

taysea: You're right! I posted a new PR to fix the placement of the component.

PR #117, opened 2019-12-17
Adjusting text and paragraph sizes in alignment with latest design

halocline:


halocline:
> So just to clarify, it looks like everything that was previously using default text size of medium is now bumped down to use small and this is because the overall font sizes have been scaled up?

Yes. `ColorRow.js`, the label in `UsageExample.js`, branding page's `HpeElementExample`, color pages `TextExample`

PR #118, opened 2019-12-17
Adjust default sizing to be consistent with type and designs

taysea:
This PR updates the defaults for aries-core components to match the designs provided by Chris. It keeps the default anchor size aligned with default text sizes, but still allows for overrides if needed.


aries-core/src/js/components/core/Nav/Nav.js
```diff
@@ -24,10 +25,12 @@ export const Nav = ({ href, title, children, background, pad }) => {
       >
         <Button href={href}>
           <Box direction="row" align="center" gap="medium">
-            <Hpe color="#01a982" size="large" />
+            <Hpe color="brand" size="large" />
```
halocline: good catch

PR #120, opened 2019-12-17
Create Philosophy page

taysea:
This PR creates the Philosophy page and updates Subsection to adhere with the new designs that include the "topic" above h1s, where topics are the main categories of "Guidelines", "Components", "Design", etc.


aries-site/src/pages/guidelines/philosophy.js
```diff
@@ -0,0 +1,57 @@
+import React from 'react';
+import { ContentSection, PageLayout, Subsection } from '../../layouts';
+import { SubsectionText } from '../../components';
+
+const title = 'Philosophy';
+
+const Philosophy = () => (
+  <PageLayout title={title}>
+    <ContentSection>
+      <Subsection level={1} name={title} topic="Guidelines">
```
halocline: Should we create a `const topic = 'Guidelines'` on line 6 similar to how `title` is treated?

taysea: Yes I think that would be a good idea.

PR #123, opened 2019-12-18
Create placeholder index.js pages

taysea:
This creates placeholder pages for the index.js pages of each main nav category. It also adjust the Layout file to create something that is generic for both the landing page and these nav pages. The padding matches that of aries-core Nav, so the content will always stay aligned with the width of the nav items.


ShimiSun:
Merging while keeping in mind that the Sidebar behavior is a placholder.

PR #126, opened 2019-12-19
Added grommet designer integration

ericsoderberghp:
This change enables the `aries-core.min.js` package to be used as a library with the Grommet Designer. The package does need to be published somewhere, so the designer can refer to it. But, we can put it somewhere other than npmjs for now.


aries-core/src/js/designer/index.js
```diff
@@ -0,0 +1,16 @@
+import { Blank } from 'grommet-icons';
+import { components } from './components';
+
+const structure = [
+  {
+    name: 'Aries',
+    Icon: Blank,
```
ShimiSun: what is the point in this placeholder of the `Blank` icon?

aries-core/src/js/designer/index.js
```diff
@@ -0,0 +1,16 @@
...
+    components: ['Tile'],
```
ShimiSun: I might be missing something, but what about the other components? 
Or are you intentionally just using one Tile and an Icon for the structure?
I guess I'm not sure what the `structure` means.

ericsoderberghp: I haven't been keeping this up to date, since this pull request is so old ;)
I thought you could check it for general issues and then I'll add the other components.

ShimiSun: The general direction of this PR LGTM.

ShimiSun:
I'll be closing this PR, feel free to update if it's still relevant.

PR #130, opened 2019-12-19
New Landing Page

### LINKED ISSUE #65: Home Page

ShimiSun:
Closes #65.
I tried to separate the contexts to different commits, let me know if you want to do another call to go over more questions you might have.


aries-site/src/pages/index.js
```diff
@@ -13,8 +25,41 @@ const Index = () => {
     }
   }, []);
 
-  // return <Landing />;
-  return <About />;
+  return (
+    <PageLayout title="Home" isLanding>
+      {/* <Grid columns={['flex', 'xxlarge', 'flex']}> */}
+      {/* <Box /> */}
```
ShimiSun: I'm keeping these comments since I might want to explore it more after we get the Nav settled.

aries-site/src/pages/index.js
```diff
@@ -13,8 +25,41 @@ const Index = () => {
...
+      <Box gap="large">
+        <HomeTiles>
+          <Tile background="white">
+            <Image src="/image-hands.png" />
```
taysea:
```suggestion
            <Image src="/image-hands.png" fit="cover" />
```
We should apply fit, otherwise the proportions of the image get squished on small screen sizes.

aries-site/src/pages/index.js
```diff
@@ -1,9 +1,21 @@
...
+import { Box, Image } from 'grommet';
+import { Tile, Tiles } from 'aries-core';
+
 import { Config } from '../../config';
+import { PageLayout } from '../layouts';
+import { data, TileContent, IntroTile } from '../components/home';
 
-import About from './start/about';
-// import Landing from './landing/landing';
+const HomeTiles = ({ ...rest }) => (
```
taysea:
```suggestion
const HomeTiles = ({ ...rest }) => {
const size = useContext(ResponsiveContext);
return (
```

aries-site/src/pages/index.js
```diff
@@ -1,9 +1,21 @@
...
+  <Tiles
+    gap="large"
+    columns={{ count: 'fit', size: 'medium' }}
+    rows="medium"
+    {...rest}
+  />
+);
```
taysea: Recommending we do this for HomeTiles because on small screens, the Grid was getting pushed off of the right side. That is because there wasn't enough room for a column size of "medium". Therefore, on small screens (when there is already only one tile in a row), we should have the tile fill the width of the column.

Additionally, on mobile there seemed to be a large gap in between each tile. In order to make the distance between the tiles and the edge of the window the same as the space between each tile, we can also adjust the gap:

```
const HomeTiles = ({ ...rest }) => {
  const size = React.useContext(ResponsiveContext);

  return (
    <Tiles
      gap={size !== 'small' ? 'large' : 'medium'}
      columns={{ count: 'fit', size: size !== 'small' ? 'medium' : 'fill' }}
      rows="medium"
      {...rest}
    />
  );
};
```

taysea: Make sure to import ResponsiveContext in the imports since we don't have it there yet.

aries-site/src/pages/index.js
```diff
@@ -13,8 +25,41 @@ const Index = () => {
...
+        <HomeTiles>
+          <Tile background="white">
+            <Image src="/image-hands.png" />
+          </Tile>
+          <IntroTile background="white" />
+        </HomeTiles>
+        <HomeTiles>
+          {data.map(item => (
+            <Tile
+              pad="medium"
```
taysea: Thinking this should be bumped up to 'large' or something bigger than medium when the screen size = 'small'. Since the padding of "medium" is responsive and changes to a smaller value for small screens, the tile content feels a little bit too close to the edge when we get to mobile.

aries-site/src/layouts/main/Layout.js
```diff
@@ -1,27 +1,34 @@
...
             <Head title={title} />
             <Header showLinks={!isLanding} />
             {!isLanding ? (
               <SidebarLayout mainContentChildren={children} />
             ) : (
               // aligns with responsive padding for aries-core Nav
-              <Main pad={size !== 'small' ? 'xlarge' : 'large'}>
-                {children}
-              </Main>
+              <Main pad={calcPad(size)}>{children}</Main>
```
taysea:
```suggestion
              <Main pad={{horizontal: calcPad(size), bottom: calcPad(size), top: "medium"}}>{children}</Main>
```
The space between the Header and the landing page content feels really big. Based on Chris's designs, the ideal distance is somewhere in between a top padding of "medium" and "large", so if we had to choose one, I think that "medium" looks nicer. This is just a thought though.

aries-site/src/pages/index.js
ShimiSun: done

aries-site/src/layouts/main/Layout.js
ShimiSun: Done. we might want to circle back to it once the Nav will have changes. but I'm OK with having it in the meanwhile.

Issue #132, opened 2019-12-20
Routing Pages - Foundation, Guidelines, Components


ShimiSun:
Setting the infrastructure for the pages that are being routed from the landing page. 
* Guidelines
* Foundation
* Components

Issue #103, opened 2019-12-11
Add Identifier component


ShimiSun:
https://hpe.invisionapp.com/share/9CU7BRLDXQS#/screens/386671637
https://hpe.invisionapp.com/share/9CU7BRLDXQS#/screens/386671650
https://hpe.invisionapp.com/share/KGU7BMHUCXV#/screens/386832392_hpe-Server-Home

These designs above are calling for the Identifier component to include an icon, title, and subtitle.
The Tile component should be refactored to use the identifier component.

Issue #24, opened 2019-11-18
Aries Foundation/Elements pages


MikeWalrath:
https://hpe.invisionapp.com/share/7KUWSK1TH4U#/screens/394162684

Foundation/Elements
- [ ] Branding/Logo
- [ ] Color
- [ ] Typography
- [ ] Layout/Composition/Spacing
- [ ] Iconography

PR #133, opened 2019-12-20
Foundation Routing 

ShimiSun:
Clicking any tile from the landing page will result in a page showing the navigation options.
This PR created a template for the Navigation options via DescriptiveHeader and NavPage.

In order to leverage the data from the Home page, I've refactored it to a data structure that will be beneficial for both Home page and the NavPage data rendering.

Ultimately, all SideBar usages will be replaced with NavPage, and SideBar will be removed from content pages. The transition should be smooth since they both have the same props and PropType.

related: #24 #132 #103


aries-site/src/components/headings/DescriptiveHeader.js
```diff
@@ -0,0 +1,24 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import { Heading, Paragraph, Box } from 'grommet';
+
+export const DescriptiveHeader = ({ subText, icon, title, ...rest }) => {
+  return (
+    <Box fill direction="row" gap="large" pad={{ vertical: 'large' }} {...rest}>
```
taysea:
```suggestion
    <Box fill direction={size !== 'small' ? 'row' : 'column'} gap="large" pad={{ vertical: 'large' }} {...rest}>
```
For mobile, the Icon should stack on top of the text.

aries-site/src/components/headings/DescriptiveHeader.js
```diff
@@ -0,0 +1,24 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import { Heading, Paragraph, Box } from 'grommet';
+
+export const DescriptiveHeader = ({ subText, icon, title, ...rest }) => {
+  return (
```
taysea:
```suggestion
const size = useContext(ResponsiveContext);
  return (
```

taysea: add Responsive Context to be able to use size.

aries-site/src/components/headings/DescriptiveHeader.js
```diff
@@ -0,0 +1,24 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import { Heading, Paragraph, Box } from 'grommet';
```
taysea:
```suggestion
import { Heading, Paragraph, Box, ResponsiveContext } from 'grommet';
```

aries-site/src/components/headings/DescriptiveHeader.js
```diff
@@ -0,0 +1,24 @@
+import React from 'react';
```
taysea:
```suggestion
import React, { useContext } from 'react';
```

aries-site/src/layouts/navigation/NavPage.js
```diff
@@ -0,0 +1,61 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Box, Text, Paragraph } from 'grommet';
+import { iconsMap } from '../../components/icons/iconsMap';
+
+import { navItemsData } from './navItemsData';
+
+const NavItem = ({ item, prefix }) => {
```
halocline: Currently, it does not look like `NavItem` scales to mobile mocks. Can we create an issue to flag that this needs to be done? (e.g. Box row gap should be reduced, plus `<Text/>` size should be reduced.

aries-site/src/layouts/navigation/NavPage.js
```diff
@@ -0,0 +1,61 @@
...
+      </Box>
+    </Box>
+  );
+};
+
+export const NavPage = ({ items, prefix, ...rest }) => {
+  return (
+    <Box {...rest}>
+      {items.map(item => (
+        <span key={item}>
```
halocline: General question: Semantically, should these be considered `<li>` and the parent `<Box as="ul">`?

aries-site/src/layouts/main/Layout.js
```diff
@@ -10,7 +10,9 @@ const calcPad = size => {
   return val;
 };
 
-export const Layout = ({ children, title, isLanding }) => {
+export const Layout = ({ children, descriptiveHeader, title, isLanding }) => {
+  // If a DescriptiveHeader is provided, then it should be a NavPage
+  const isNavPage = descriptiveHeader;
```
halocline: I wonder if would be clearer to explicitly state `isNavPage` as a boolean prop, rather that inferring type of page layout based on an element being passed or not?

aries-site/src/layouts/navigation/NavPage.js
taysea: That would make sense to me.

aries-site/src/layouts/main/Layout.js
taysea: I am open to that as well. I had that initially actually, but had changed it but maybe for clarity it would be best to define it as a prop.

aries-site/src/layouts/main/Layout.js
```diff
@@ -21,12 +23,39 @@ export const Layout = ({ children, title, isLanding }) => {
             margin="auto"
           >
             <Head title={title} />
-            <Header showLinks={!isLanding} />
-            {!isLanding ? (
+            <Header
+              showLinks={!isLanding && !isNavPage}
```
halocline: Expanding the number and complexity of conditionals starts to make it difficult to know what the `<Layout />` component is going to return. I think we are fine to move forward as is for now to keep things moving, but would like to revisit how we are approaching page layouts in the future --> I'd like to discuss moving in a direction where the consumer of a page layout is explicitly stating which type of page layout they would like (e.g. Content Page Layout, Landing Page Layout, Navigation Page Layout, etc.).

With the current conditionals, I think this is manageable, but introducing any more becomes more difficult to read and reason about, increases the number of edge cases, and increases risk of introducing bugs.

aries-site/src/layouts/navigation/NavPage.js
taysea: I incorporated this change into my recent commit so we don't need to create a separate issue for it.

aries-site/src/layouts/main/Layout.js
taysea: I agree, this is a concept we should revisit.

aries-site/src/layouts/main/Layout.js
```diff
@@ -21,12 +23,39 @@ export const Layout = ({ children, title, isLanding }) => {
...
+              background={
+                descriptiveHeader && descriptiveHeader.props.background
+              }
+            />
+            {!isLanding && !isNavPage ? (
```
halocline: https://github.com/hpe-design/aries/pull/133/files#r362637416

aries-site/src/layouts/navigation/NavPage.js
taysea: I incorporated this change into my recent commit so we don't need to create a separate issue for it.

halocline: Great!

PR #134, opened 2019-12-21
Adjustments to NavPages for header background and responsive behavior

taysea:
This builds on top of Shimi's PR regarding the NavPages. It attempts to incorporate Chris's design where the header background color expands to the max-width and affects the background of the Nav.

There may be a more concise way to achieve this result, but I wanted to post my solution to at least have as reference for now. If we'd like to iterate on it, that is fine with me but the behavior is responsive for desktop and mobile.


aries-site/src/layouts/main/Header.js
```diff
@@ -3,8 +3,8 @@ import PropTypes from 'prop-types';
 
 import { AnchorGroup, Nav } from 'aries-core';
 
-export const Header = ({ showLinks }) => (
-  <Nav title="Design System">
+export const Header = ({ background, showLinks }) => (
+  <Nav title="Design System" background={background}>
```
ShimiSun: I'm OK with that change, I'm just thinking that if we'll have `...rest` as a prop we won't need the deconstruction of the `background` prop.

ShimiSun:
Overall I think it looks good! I think it will be easier if you can just push your changes to `feat/routing` branch so I can see the diff and make sure I covered everything you added, will that work? @taysea

ShimiSun:
Continues work of https://github.com/hpe-design/aries/pull/133

taysea:
Closing since I merged it into the other PR.

PR #135, opened 2020-01-02
Search component for navigation

taysea:
This PR sets up the Search component for the navigation of the site. When the input is clicked, the suggestions open and adapt to what the user types. The suggestions include the page titles only and does not search through all page content, similar to how grommet-site works.

The piece that needs to be figured out is closing the text input when the search component goes out of focus on small screens. Per the designs, on small screens the search icon should be the only thing visible by default but should expand to show the text input when clicked. The piece I am stuck on is how to handle the suggestions closing and the text input collapsing when clicked out of. You'll notice on small screens that once the search icon comes into focus and the text input opens, even after leaving the textinput won't close again (even though it should).


halocline:
> The piece I am stuck on is how to handle the suggestions closing and the text input collapsing when clicked out of. You'll notice on small screens that once the search icon comes into focus and the text input opens, even after leaving the textinput won't close again (even though it should).

Would adding `onBlur={() => setFocused(false)}` to the outer `<Box>` returned by `Search.js` do the trick?

taysea:
> > The piece I am stuck on is how to handle the suggestions closing and the text input collapsing when clicked out of. You'll notice on small screens that once the search icon comes into focus and the text input opens, even after leaving the textinput won't close again (even though it should).
> 
> Would adding `onBlur={() => setFocused(false)}` to the outer `<Box>` returned by `Search.js` do the trick?

I tried that, and in addition to keep it available via keyboard (bc the tabIndex was getting changed to -1 after it first got accessed), I added `tabIndex={focused ? -1 : 0}`. However, what happens is after you tab past it or click away, the search icon continues to display the focus indicator even though it isn't the active element. Any thoughts on why that is? @halocline

aries-site/src/utils/search.js
```diff
@@ -0,0 +1,38 @@
+import { structure } from '../data';
+
+export const formatName = name => {
+  return name
+    .split(' ')
+    .join('-')
+    .toLowerCase();
+};
+
+const getTopicName = itemName => {
```
halocline: Glad you added these utility functions. I had started something similar, but will refactor to use this. Thank you.

halocline:
Also, can you look at how the search box appears on the NavPages? Difficult to read. I wonder what we can do make it more legible?

Screen Shot 2020-01-08 at 2 13 41 PM

PR #143, opened 2020-01-03
Create Human Centered Page

taysea:
This fills in the content for the Human Centered Page. It also makes a small tweak to ContentSection which makes sure that content is aligned start so that Buttons don't span the entire width of the section. Additionally, a formatting piece is added to NavPage to be able to handle routing to pages with spaces in their titles.

### DIFF
--- a/aries-site/src/layouts/content/ContentSection.js
+++ b/aries-site/src/layouts/content/ContentSection.js
@@ -8,6 +8,7 @@ export const ContentSection = ({ children, lastSection }) => {
   return (
     <Box
       as="section"
+      align="start"
       border={!lastSection ? { side: 'bottom' } : undefined}
       gap="medium"
       margin={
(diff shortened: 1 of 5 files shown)


aries-site/src/layouts/navigation/NavPage.js
```diff
@@ -8,12 +8,17 @@ import { navItemsData } from './navItemsData';
 const NavItem = ({ item, prefix }) => {
   const size = useContext(ResponsiveContext);
   const itemLowerCase = item.toLowerCase();
+  const formattedItem = item
```
halocline: Good add. I noticed that reviewing yesterday as well.

aries-site/src/layouts/content/ContentSection.js
```diff
@@ -8,6 +8,7 @@ export const ContentSection = ({ children, lastSection }) => {
   return (
     <Box
       as="section"
+      align="start"
```
halocline: good add

PR #146, opened 2020-01-03
Branding page content updates

halocline:
Related Issues: https://github.com/hpe-design/aries/issues/138

Work Included:
- Content updates to `branding.js`
- Tweak to `<UsageExample>`


aries-site/src/pages/foundation/branding.js
```diff
@@ -1,21 +1,38 @@
...
+  );
+};
 
 const HpeElementExample = () => {
   const textSize = 'small';
 
   return (
     <Box direction="row" align="center" gap="medium">
-      <Hpe color="brand" size="xlarge" />
+      <Hpe color="brand" size="66px" />
```
ShimiSun: Do we have to use a customized size instead of t-shirt sizing?

aries-site/src/pages/foundation/branding.js
```diff
@@ -1,21 +1,38 @@
...
 } from '../../layouts';
-import { MainHeading } from '../../components';
+import { SubsectionText } from '../../components';
+
+const ArubaIconExample = () => {
+  const textSize = 'small';
+
+  return (
+    <Box direction="row" align="center" gap="medium">
+      <Aruba color="orange!" size="30px" />
```
ShimiSun: Do we have to use a customized size instead of t-shirt sizing?

ShimiSun:
In case we think the Aruba svg will always be used with the 'bang' color, I think this one is good to go. We just need to be mindful the color changes of dark/light mode theme changes may not apply on the icons.

aries-site/src/pages/foundation/branding.js
halocline: Because this is a branding example, the ratio of font-size to icon size (and spacing between icon and text) is important; t-shirt sizing for the does not present the ratios correctly. So, in this case, I believe we need to deviate from t-shirt size for the icon.

aries-site/src/pages/foundation/branding.js
halocline: Because this is a branding example, the ratio of font-size to icon size (and spacing between icon and text) is important; t-shirt sizing for the does not present the ratios correctly. So, in this case, I believe we need to deviate from t-shirt size for the icon.

halocline:
> In case we think the Aruba svg will always be used with the 'bang' color, I think this one is good to go. We just need to be mindful the color changes of dark/light mode theme changes may not apply on the icons.

Since this is a branding example for Aruba, and 'orange!' is Aruba's brand color, I believe we should be okay.

PR #149, opened 2020-01-07
Color page content updates

halocline:
Related Issues: https://github.com/hpe-design/aries/issues/147

Work Included:

- Incorporated all current content for color page
- Removed `<LastUpdated>` component which no longer occurs in the designs
- Tweak to `<UsageExample>` implementation allowing better mobile rendering
- Tweak to `<ColorRow>` implementation
- Minor update to `branding.js` for better mobile rendering

Note:

- There is a bug with the text color being rendered on top of the dark `background-contrast` --> need to investigate the source


halocline:
> LGTM. My only comment was going to be how text color gets handled with background-contrast -- I was running into a similar thing in the Search component, seems like it's using the low opacity background as the source and thinks dark text should be used but actually the true background is already dark. This might be something in grommet?

Yes, I believe it is in Grommet, but need to verify. Added this issue https://github.com/hpe-design/aries/issues/151

PR #168, opened 2020-01-07
Consolidate Data structure

taysea:
This PR consolidates the page data into a single structure.js file. Previously data was being held in various places, causing some repeats in data. The goal of this is to consolidate the data in one place.

Additionally, uses of the term "prefix" were replaced with "topic".


aries-site/src/data/structure.js
```diff
@@ -0,0 +1,148 @@
...
+  IconResources,
+  IconBrand,
+  IconColor,
+  IconIcons,
+  IconLayout,
+  IconType,
+  IconToken,
+} from '../components/icons';
+
+export const structure = [
```
halocline: Now that this is flattened, `pages` might be more precise that `structure`. Not a big deal either way.

aries-site/src/pages/index.js
```diff
@@ -29,7 +30,7 @@ const Index = () => {
     }
   }, []);
 
-  const data = Object.values(homeData);
+  const data = structure.filter(page => page.pages);
```
halocline: `data` is a bit generic here, wonder if `topicList` is more explicit and would communicate to developer what this data represents?

halocline: What happens if a content page gets populated with a `pages` property?  Won't it now errantly get picked up in this filter?

What if we gave the object in the data structure a property of `parent: 'Home'` and the modified the filter to be `const data = structure.filter(page => page.parent === title` (Where title is a variable that then populates the `<PageLayout>` title as well)?   This way, we will only receive pages for which we are explicitly asking.

aries-site/src/layouts/main/SidebarLayout.js
```diff
@@ -27,8 +28,8 @@ export const SidebarLayout = ({ mainContentChildren }) => {
 
   // TODO selectedNav should be retrived from aries-core
   // as the selected element of the NavBar
-  const selectedNav = 'foundation';
-
+  const topic = 'Foundation';
```
halocline: We are going to have to rework how this topic variable gets populated. It should get passed in as a prop from the component consuming `<SidebarLayout>` --- outside the scope of this PR though.

taysea: Agreed.

taysea: Agreed.

aries-site/src/data/structure.js
taysea: I like pages. I'll make a switch.

aries-site/src/pages/index.js
taysea: I like this. I'll add that in.

taysea: Agreed, we can be more specific.

aries-site/src/pages/resources/index.js
```diff
@@ -1,26 +1,24 @@
 import React from 'react';
-import { PageLayout, NavPage, SideBarItemList } from '../../layouts';
+import { PageLayout, NavPage } from '../../layouts';
 import { DescriptiveHeader } from '../../components/headings';
-import { data } from '../../components/home';
+import { structure } from '../../data';
 
 const title = 'Resources';
-const prefix = title.toLowerCase();
+const [topic] = structure.filter(page => page.name === title);
```
halocline: good

aries-site/src/data/structure.js
```diff
@@ -0,0 +1,148 @@
...
+  {
```
halocline: To be comprehensive, we should add an object for the Home page.
```
{
  name: 'Home',
  pages: ['Guidelines', ... ],
}
```

Would require modification to this filter mentioned here: https://github.com/hpe-design/aries/pull/168/files?file-filters%5B%5D=.js&hide-deleted-files=true#r363999070

aries-site/src/data/structure.js
```diff
@@ -0,0 +1,148 @@
...
+    name: 'Guidelines',
+    color: 'green',
+    description:
+      'Why does HPE have a design system? All the aesthetics, best practices, and information about the platform and how to wield it.',
+    icon: size => <IconGuidelines size={size} />,
+    pages: ['Principles', 'Human Centered', 'Philosophy'],
```
halocline: Recommend adding `parent: 'Home` per this filter comment: https://github.com/hpe-design/aries/pull/168/files?file-filters%5B%5D=.js&hide-deleted-files=true#r363999070

aries-site/src/data/structure.js
```diff
@@ -0,0 +1,148 @@
...
+  },
+  {
+    name: 'Foundation',
+    color: 'red',
```
halocline: This is going to be nice to have.

aries-site/src/pages/components/index.js
```diff
@@ -1,26 +1,24 @@
 import React from 'react';
-import { PageLayout, NavPage, SideBarItemList } from '../../layouts';
+import { PageLayout, NavPage } from '../../layouts';
 import { DescriptiveHeader } from '../../components/headings';
-import { data } from '../../components/home';
+import { structure } from '../../data';
 
 const title = 'Components';
-const prefix = title.toLowerCase();
+const [topic] = structure.filter(page => page.name === title);
```
taysea: Wondering if here we can use the .find() approach as well? @halocline

PR #169, opened 2020-01-08
tweak data structure

halocline:


aries-site/src/data/structure.js
```diff
@@ -69,51 +85,61 @@ export const structure = [
       'Tutorials, case studies, videos, how-to&aposs, articles, upcoming events and more about HPE Design and the Design System from the team.',
     icon: size => <IconResources size={size} />,
     pages: ['Examples', 'Videos', 'Designer', 'Themer', 'Grommet'],
+    parent: 'Home',
   },
   {
     name: 'Principles',
     description: 'Something',
     icon: '',
+    parent: 'Guidelines',
```
taysea: Is it redundant to have 'parent' on each of these pages and to have them listed as children in the above pages arrays? Just wondering if we plan to use 'parent' how every page or if it's only needed on the Home ones.

halocline: It is redundant, but would make for more efficient queries in certain cases. However, I thought of another approach which will still be efficient, but without the redundant data.

taysea: Cool, I'd love to chat about that.

halocline: Modified implementation. Check it out now.

aries-site/src/pages/index.js
```diff
@@ -23,7 +23,11 @@ const HomeTiles = ({ ...rest }) => {
 };
 
 const title = 'Home';
-const topicList = structure.filter(page => page.parent === title);
+
+// Get details for each child page of the home page
+const topicList = structure
+  .find(page => page.name === title)
+  .pages.map(topic => structure.find(page => page.name === topic));
```
taysea: Nice!

PR #178, opened 2020-01-10
Update content for guidelines navpage

halocline:
Related Issues: https://github.com/hpe-design/aries/issues/154

Work Included:

- Updated content for Guidelines nav page
- Added Meta data for Guidelines nav page
- Added `showHeading` prop to `<Subsection>` to allow for instances where we want a subsection, but the design doesn't call for an accompanying heading.

Screen Shot 2020-01-10 at 9 12 45 AM


aries-site/src/pages/guidelines/index.js
```diff
@@ -1,10 +1,22 @@
...
+  PageLayout,
+  ContentSection,
+  Subsection,
+} from '../../layouts';
+import { DescriptiveHeader, Meta, SubsectionText } from '../../components';
 import { structure } from '../../data';
 
 const title = 'Guidelines';
 const topic = structure.find(page => page.name === title);
+const description =
```
taysea: Noticing this is also defined in structure -- is it possible for us to pull it from there instead?

halocline: Discussed offline. Going to create seoDescription property in structure.js and use it to populated seo meta descriptions

PR #185, opened 2020-01-10
Controls Page

taysea:
This PR sets up the Controls page under the Components category. It introduces an "examples" directly which is subdivided to match the site structure. This will contain any examples that are on content pages. Additionally, changes have been made to aries theme to assume the designs represent the default for aries components.


aries-site/src/examples/components/controls/AccordionExample.js
```diff
@@ -0,0 +1,30 @@
...
+
+export const AccordionExample = () => {
+  return (
+    <Box width={{ max: 'medium' }}>
+      <Accordion>
+        <AccordionPanel label="Our Company">
+          <Box pad={{ vertical: 'medium' }}>We are HPE.</Box>
+        </AccordionPanel>
+        <AccordionPanel label="Our History">
+          <Box pad={{ vertical: 'medium' }}>
```
halocline: Little opportunity to create an `<AccordianContent>` sub component since `<Box pad={{ vertical: 'medium' }} >` is being repeated 4 times. Not necessary though.

aries-site/src/examples/components/controls/MenuExample.js
```diff
@@ -0,0 +1,18 @@
...
+export const MenuExample = () => {
+  const items = [
+    { label: 'Past 12 months' },
+    { label: 'Past 6 months' },
+    { label: 'Past 3 months' },
+    { label: 'Past 30 days' },
+    { label: 'Past 7 days' },
+  ];
+  return (
+    <Box background="light-2" round="xsmall">
```
halocline: Should we try to stick to using colors in the aries theme? If so, `background-front` is the color I believe we should be using.

halocline: Also, going through this is making me wonder if these examples should be shown in both light and dark contexts? Mocks only show one context for controls, but do we want to be consistent with branding.js and color.js?

ShimiSun:
Some questions from looking at the deployment and comparing with the designs:
Component page:
https://hpe.invisionapp.com/share/NKV7IJ4BTMU#/screens/399919153
Missing to following: 
Introduction, Link to Grommet, link to Layout (or just a placeholder?)

Controls page:
Why is there a storybook anchor rather than a button looking to 'Use the Controls component' as the design suggests?
The content of the deployment is so much different than what I see on the designs, https://hpe.invisionapp.com/share/NKV7IJ4BTMU#/screens/399919149, am I looking at the right version?
image

aries-site/src/examples/components/controls/TabsExample.js
```diff
@@ -0,0 +1,33 @@
+import React from 'react';
+import { Tab, Tabs } from 'aries-core';
+import { Box } from 'grommet';
+
+export const TabsExample = () => {
+  const [index, setIndex] = React.useState();
+  const onActive = nextIndex => setIndex(nextIndex);
+
+  return (
+    <Box pad="large" background="background-front">
```
halocline: Noticing that the design mock does not have this padding...  For consistency, should we consider using `<UsageExample>` for each of these examples which would provide consistent padding from example to example, plus support light and dark themeMode illustrations ([similar](https://github.com/hpe-design/aries/pull/185/files#r365909462)) ?

aries-site/src/examples/components/controls/MenuExample.js
taysea: You're right. We should stick with the aries theme colors. I think it also should support light and dark contexts. Let me make adjustments to ensure that's the case.

aries-site/src/examples/components/controls/TabsExample.js
taysea: We can remove this from Tabs example to match the design for now. I noticed the switch in the Accordion example to have the background with padding and I thought maybe the Tabs would use that as well. For now, I will change it to be consistent with the designs. If a background/padding is added in the future I agree that UsageExample should be used.

taysea:
> Some questions from looking at the deployment and comparing with the designs:
> Component page:
> https://hpe.invisionapp.com/share/NKV7IJ4BTMU#/screens/399919153
> Missing to following:
> Introduction, Link to Grommet, link to Layout (or just a placeholder?)
> 
> Controls page:
> Why is there a storybook anchor rather than a button looking to 'Use the Controls component' as the design suggests?
> The content of the deployment is so much different than what I see on the designs, https://hpe.invisionapp.com/share/NKV7IJ4BTMU#/screens/399919149, am I looking at the right version?
> image

Hey @ShimiSun we are now using Figma for all the designs. The link you sent is not the current material. I'll send you an invite.

aries-site/src/themes/aries.js
```diff
@@ -204,6 +212,18 @@ export const aries = deepMerge(hpe, {
       xxlarge: '166px',
     },
   },
+  tab: {
```
halocline: Do we want to explicitly set `tab.active.color` to 'text' in this theme? It is the default in Grommet's base.js, but I wonder if it might be beneficial to make that clear here?

aries-site/src/examples/components/controls/MenuExample.js
taysea: The reason I had put a light background on that box was so that the Menu thought it was in a light context and would render light theme colors for the button -- otherwise, the control button blends in with the page background because there is no border. I still think the example should support light and dark modes, but I'm wondering how to better distinguish the Menu button from the background if it's showing dark mode menu in dark mode site. @halocline

aries-site/src/themes/aries.js
taysea: I am open to that for clarity, but is common convention only to define things that alter from grommet base.js?

aries-site/src/components/content/FigmaAnchor.js
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Box } from 'grommet';
+import { AnchorCallToAction } from 'aries-core';
+
+export const FigmaAnchor = ({ href }) => {
+  return (
+    <Box align="start">
+      <AnchorCallToAction
```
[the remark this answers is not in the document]
taysea: Good call.

aries-site/src/components/content/index.js
```diff
@@ -1,4 +1,8 @@
 export * from './BulletedList';
+// was getting a linting error in controls.js that it couldn't
+// find figma anchor unless I exported as such.
+export { FigmaAnchor } from './FigmaAnchor';
 export * from './ComingSoon';
 export * from './SubsectionText';
 export * from './SubmitFeedback';
+export * from './StorybookAnchor';
```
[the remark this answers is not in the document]
taysea: I understand what you're saying. I guess my thought was to avoid having to write the labels of "Use with Figma" or "See in Storybook" so many times. I am open to consolidating them. Do you have an idea for a clear name for that component?

aries-site/src/themes/aries.js
```diff
@@ -204,6 +212,18 @@ export const aries = deepMerge(hpe, {
...
+    border: {
+      color: 'text-xweak',
```
halocline: Default border is 2px. Design is showing a thin border bottom (0.5 px, but we will have to do 1px), and the active state has a 3px border. Not sure Grommet currently supports mixing the border sizes on Tab.

aries-site/src/themes/aries.js
```diff
@@ -204,6 +212,18 @@ export const aries = deepMerge(hpe, {
...
+    },
+    color: 'text-xweak',
+    margin: {
+      horizontal: 'none',
+    },
+    pad: {
+      horizontal: 'small',
```
halocline: Assuming we are using this rather than `tabs.gap` because gap creates breaks in the border instead of the continuous line as shown in mock?

aries-site/src/themes/aries.js
taysea: I don't think it supports mixing border sizes. I couldn't find a way to do that last week. We should get clarity from the designers which px value they would like if it can only be one value for active and default. Good catch on the detail.

aries-site/src/themes/aries.js
halocline: Design looks like it is using a larger spacing between tabs, but bumping up forces tab to new row...  

Also, wondering how we translate this to mobile...

taysea: Precisely. But setting "tabs.gap" to 'none' doesn't remove the gap. The horizontal margin needs to be removed on individual tabs to get the continuous line.

aries-site/src/themes/aries.js
ShimiSun: The custom theme is using base.js by default, so it may be redundant to declare styles that already exist in base.js unless they are altered.

halocline: Yes, it would be redundant. Fine with it as is.

PR #186, opened 2020-01-10
Create Coming Soon component

taysea:
This creates a coming soon component that is being used temporarily for Develop, Design, and Resources page since the content of those topics will not be part of the mvp.

We don't have designs for a coming soon layout right now, so this is a placeholder that can hopefully prompt some reactions from people.

I added a minHeight to the Layout to ensure that the footer stays at the bottom of the page even when the content is short.

### DIFF
--- a/aries-site/src/layouts/main/Layout.js
+++ b/aries-site/src/layouts/main/Layout.js
@@ -23,6 +23,7 @@ export const Layout = ({
         {size => (
           <Box
             // pad={{ horizontal: calcPad(size) }}
+            height={{ min: '100vh' }}
             margin="auto"
             width={{ max: 'xxlarge' }}
           >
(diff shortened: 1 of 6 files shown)


aries-site/src/layouts/main/Layout.js
```diff
@@ -23,6 +23,7 @@ export const Layout = ({
         {size => (
           <Box
             // pad={{ horizontal: calcPad(size) }}
+            height={{ min: '100vh' }}
```
halocline: nice.

halocline: Should we do something similar for the SidebarLayout?

taysea: Yeah we can look into seeing if the same thing would work.

PR #197, opened 2020-01-13
Use Link from next/link

taysea:
This updates the way the site navigates internally to utilize Link from next/link. Now, when you navigate to a new page it does not refresh every time. Some adjustments needed to be made to aries-core components to wrap them in forwardRef in order to work with how nextjs Link works.

For details on why forwardRef was used: https://nextjs.org/docs/api-reference/next/link#example-with-reactforwardref

For details on why passHref was added: https://github.com/zeit/next.js/#forcing-the-link-to-expose-href-to-its-child

I have included these links in comments in the files where they were used.

### DIFF
--- a/aries-site/src/pages/index.js
+++ b/aries-site/src/pages/index.js
@@ -24,14 +24,14 @@ const HomeTiles = ({ ...rest }) => {
 };
 
 // Reasoning for using forwardRef: https://nextjs.org/docs/api-reference/next/link#example-with-reactforwardref
-const TopicTile = forwardRef(({ onClick, topic }, ref) => {
+const TopicTile = forwardRef(({ topic, ...rest }, ref) => {
   return (
     <Tile
       pad="medium"
       background={topic.color}
       key={topic.color}
-      onClick={onClick}
       ref={ref}
+      {...rest}
     >
       <TileContent
         key={topic.name}
@@ -69,7 +69,11 @@ const Index = () => {
         </HomeTiles>
         <HomeTiles>
           {topicList.map(topic => (
-            <Link key={topic.name} href={`/${topic.name.toLowerCase()}`}>
+            <Link
+              key={topic.name}
+              href={`/${topic.name.toLowerCase()}`}
+              passHref
+            >
               <TopicTile topic={topic} />
             </Link>
           ))}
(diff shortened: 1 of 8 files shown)


aries-site/src/pages/index.js
```diff
@@ -48,21 +69,13 @@ const Index = () => {
...
-              pad="medium"
-              background={topic.color}
-              key={topic.color}
-              onClick={() =>
-                (window.location.href = `/${topic.name.toLowerCase()}`)
-              }
+            <Link
+              key={topic.name}
+              href={`/${topic.name.toLowerCase()}`}
+              passHref
```
halocline: Do you want documentation here for consistency? 
`// Need to pass href because of: https://github.com/zeit/next.js/#forcing-the-link-to-expose-href-to-its-child`

taysea:
> How does this work for our Nav component? Branding link in Header does not currently have this applied?
> 
> I see that `forwardRef` has been applied to any components in `aries-core` used by `aries-site`. Does this need to be applied comprehensively to any `aries-core` components used for routing (e.g. Anchor, FooterLink, others)?

The Nav component would be the one place that we still need to add this -- that was in my backlog, so thanks for calling it out. We would need to figure out how to wrap that individual piece of the Nav in the Link.

aries-site/src/pages/index.js
taysea: Yes, I should add it here. Good catch.

halocline:
> The Nav component would be the one place that we still need to add this -- that was in my backlog, so thanks for calling it out. We would need to figure out how to wrap that individual piece of the Nav in the Link.

Since the Nav interface is more complex and will need to be re-thought, are you thinking of creating an issue specifically for it? Or, incorporating into this PR?

taysea:
> > The Nav component would be the one place that we still need to add this -- that was in my backlog, so thanks for calling it out. We would need to figure out how to wrap that individual piece of the Nav in the Link.
> 
> Since the Nav interface is more complex and will need to be re-thought, are you thinking of creating an issue specifically for it? Or, incorporating into this PR?

My thought would be to create a separate issue for it considering Nav will likely see large re-work as we break down components, etc.

halocline:
> My thought would be to create a separate issue for it considering Nav will likely see large re-work as we break down components, etc.

I am good with that.

PR #230, opened 2020-01-16
Feat input page

halocline:
Related Issue: https://github.com/hpe-design/aries/issues/196

Input examples, Oh My!!!

Includes:
- Input page structure and content
- Input examples


aries-site/src/examples/components/input/RangeSelectorExample.js
```diff
@@ -0,0 +1,33 @@
...
+  return (
+    <UsageExample>
+      <Box gap="medium">
+        <Box direction="row" gap="medium">
+          <Text>Threshold</Text>
+          <Text weight={700}>
+            {values[0]}% - {values[1]}%
+          </Text>
+        </Box>
+        <Box direction="row" gap="medium" width="large" justify="between">
```
taysea:
```suggestion
        <Box direction="row" gap="medium" width="large" justify="between" align="center">
```
Think we need an align center for the range selector to be aligned w the end labels better.

aries-site/src/examples/components/input/MaskedInputExample/MaskedInputExample.js
```diff
@@ -0,0 +1,27 @@
...
+  MaskedTimeExample,
+  MaskedPhoneExample,
+} from '.';
+
+export const MaskedInputExample = () => {
+  return (
+    <UsageExample>
+      <Box gap="medium">
+        <Box direction="row-responsive" gap="large">
+          <MaskedEmailExample />
```
taysea: For each of these examples, as if I click into it and the drop opens, I can't get it to close unless I fill in the field completely. Seems like when I click outside, the drop should close -- this is how it is for MaskedInput on Storybook for grommet

aries-site/src/examples/components/input/TextInputExample.js
```diff
@@ -0,0 +1,72 @@
+import React, { useState } from 'react';
+import { Box, FormField } from 'grommet';
+import { TextInput } from 'aries-core';
+
+import { FormRow, UsageExample } from '../../../layouts';
+
+const usernamePlaceholder = 'Email address';
+const passwordPlaceholder = 'Password';
+
+const LabelInlineExample = () => {
```
taysea: This example is having a very difficult time in that in between medium/small screen where the main content width has already hit small. The username starts to overlap the textinput heavily. I'm wondering how we can improve this experience if possible.

halocline: Agree. Thinking we separate into two UsageExample instances, but was trying to stay true to design mock.

aries-site/src/examples/components/input/FormExample.js
```diff
@@ -0,0 +1,69 @@
...
+        </FormRow>
+      </Box>
+    </Form>
+  );
+};
+
+export const FormExample = () => {
+  return (
+    <Box gap="medium">
+      <UsageExample label="Login or Account Information">
```
taysea: I notice that you are using the styling for the usage example label here but that the designs call for something bigger/bolder. I think this makes sense for consistency across pages but I'd like to get confirmation from design that this standard works.

aries-site/src/examples/components/input/MaskedInputExample/MaskedInputExample.js
halocline: What's interesting here is that the Storybook MaskedInput code placed here creates the same behavior...  The difference is that here the MaskedInput is contained in a `<FormField>`. If I remove the FormField wrapper, the expected behavior returns. So, there is something going on in the interaction between FormField and MaskedInput.

Any ideas?

taysea: Hmm that's interesting -- seems like it could be a grommet thing. No solution off the top of my head but sounds like something to look into with grommet. Don't think it should hold up this PR, but should be a backlog item.

halocline: Yes, I believe it is. I will create a Codesandbox example to replicate and then file an issue with Grommet.

aries-site/src/examples/components/input/RangeSelectorExample.js
halocline: Done.

aries-site/src/examples/components/input/FormExample.js
halocline: Spoke w/ Greg and going for consistency

aries-site/src/examples/components/input/TextInputExample.js
halocline: Removed left labels per earlier discussion

PR #232, opened 2020-01-16
Components - Layout page

taysea:
This PR creates the Layout page that falls under the Components section. One note is that I needed to name the examples folder 'layouts' because 'layout' seemed to be a special folder name bc it was greyed out/didn't appear in files I had created/could push.

A couple notes/future to-dos:
- It would be create to have a responsive context that responds to the width of a bounding box -- as you shrink the window the footer/header components become crunched before the window hits the small breakpoint and shifts the layouts. It'd be best if they were responsive to when the main content width hits a width of small.
- Layout is a duplicate title for pages in Foundation and Components -- we will need to adjust the data structure to know how to find which page we want and populate the sidebar accordingly to the topic. For now, since we don't have content for Foundation/Layout, I've removed it from the data structure to allow the sidebar on this page to populate correctly.


aries-site/src/examples/components/layouts/LayerExample.js
```diff
@@ -0,0 +1,55 @@
...
+            width={size !== 'small' ? 'medium' : undefined}
+            pad="medium"
+          >
+            <Box flex={false} direction="row" justify="between">
+              <Heading level={2} margin="none">
+                This is a Layer.
+              </Heading>
+              <Button icon={<Close />} onClick={onClose} />
+            </Box>
+            <Box flex="grow" overflow="auto" pad={{ vertical: 'medium' }}>
```
halocline: Not a code issue, but from a UX perspective the content and the close button are quite far apart. Do we need the flex="grow" here? Does the button pushed to the bottom add value?

aries-site/src/pages/components/layout.js
```diff
@@ -0,0 +1,100 @@
...
+          possibilities.
+        </SubsectionText>
+        <LayerExample />
+      </Subsection>
+    </ContentSection>
+    <ContentSection>
+      <Subsection name="Main">
+        <SubsectionText>
+          The Main component is where you define the location and layout of the
+          primary context of your content. Check out the Grid component
```
halocline: Can we add an Anchor and link to the Grid section?

aries-site/src/examples/components/layouts/LayerExample.js
```diff
@@ -0,0 +1,55 @@
...
+      </Box>
+      {open && (
+        <Layer
+          position="right"
+          full={size !== 'small' ? 'vertical' : true}
+          modal
+          onClickOutside={onClose}
+          onEsc={onClose}
+        >
+          <Box
```
halocline: Why is the layer here presenting with a white background? I would think it would be "background-front".

aries-site/src/pages/components/layout.js
taysea: Good call.

aries-site/src/examples/components/layouts/LayerExample.js
taysea: Background front would work.

aries-site/src/examples/components/layouts/LayerExample.js
taysea: Agreed -- it is a bit far from the content. Was pulling from a grommet storybook example which had a bit more of a form here. I think bringing the button up would be fine.

aries-site/src/examples/components/layouts/LayerExample.js
taysea: From Slack: Running into a situation where the Layer was coming through as white because it thinks it's on a light background -- trying to figure out how to get it to know that the whole site is in dark mode

Issue #240, opened 2020-01-17
Accessibility - Add label to search input


halocline:
![Screen Shot 2020-01-17 at 8 37 01 AM](https://user-images.githubusercontent.com/1756948/72624795-ca01f600-3904-11ea-9d31-4d609dafba02.png)

taysea:
This is a duplicate of #201 , I will leave this one open and close that one.

PR #255, opened 2020-01-17
Add aria-labelledby to TextInput

taysea:
Using Search icon as the arialabelledby for the text input. Documentation on why this is a valid replacement for using label here: https://www.w3.org/WAI/tutorials/forms/labels/#using-aria-labelledby. More info can also be found here: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/forms/Basic_form_hints

At first, I tried to use grommet formField with the text input, but I was running into issues with how it affected the styling of the component. Instead, I decided to go for this aria approach since adhering to the designs would've required visually hiding the label anyway. There is enough visual context for the user to know that the textinput is a search form, so it meets the requirements of the above resources. Open to discussion on this approach.

### DIFF
--- a/aries-site/src/layouts/navigation/Search.js
+++ b/aries-site/src/layouts/navigation/Search.js
@@ -6,6 +6,7 @@ import React, {
   useCallback,
 } from 'react';
 import { useRouter } from 'next/router';
+import styled from 'styled-components';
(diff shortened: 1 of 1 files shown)


taysea:
Linking to #240

PR #270, opened 2020-01-21
Update sidebar active state

taysea:
The reason that the active state was not persisting was because of the way NavLink was removing active styling when hovered over or focused/blurred. In cases where the link is active because of the current path of the app, we don't want to have any changes to its style.

Also added quick refactor of active variable in SideBarItem for clarity.

### DIFF
--- a/aries-core/src/js/components/helpers/Anchors/NavLink.js
+++ b/aries-core/src/js/components/helpers/Anchors/NavLink.js
@@ -19,10 +19,10 @@ export const NavLink = forwardRef(
             {label || children}
           </Text>
         }
-        onBlur={() => setFontWeight(defaultFontWeight)}
-        onFocus={() => setFontWeight(activeFontWeight)}
-        onMouseOut={() => setFontWeight(defaultFontWeight)}
-        onMouseOver={() => setFontWeight(activeFontWeight)}
+        onBlur={() => !active && setFontWeight(defaultFontWeight)}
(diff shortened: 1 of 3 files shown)


aries-site/src/layouts/navigation/SideBar.js
```diff
@@ -4,16 +4,17 @@ import Link from 'next/link';
...
+  const { pathname } = useRouter();
   const path = nameToPath(item);
-  const router = useRouter();
+  const active = pathname === path;
+
   return (
     // Need to pass href because of: https://github.com/zeit/next.js/#forcing-the-link-to-expose-href-to-its-child
     <Link href={path} passHref>
-      <NavLink active={router.pathname === path}>{item}</NavLink>
+      <NavLink active={active}>{item}</NavLink>
```
halocline: 👍

aries-core/src/js/components/helpers/Anchors/NavLink.js
```diff
@@ -19,10 +19,10 @@ export const NavLink = forwardRef(
             {label || children}
           </Text>
         }
-        onBlur={() => setFontWeight(defaultFontWeight)}
-        onFocus={() => setFontWeight(activeFontWeight)}
-        onMouseOut={() => setFontWeight(defaultFontWeight)}
-        onMouseOver={() => setFontWeight(activeFontWeight)}
+        onBlur={() => !active && setFontWeight(defaultFontWeight)}
```
halocline: 👍

PR #283, opened 2020-01-22
Unwrap components from aries-core and update storybook/aries-site

### LINKED ISSUE #280: "Unwrap" grommet components from aries and update storybook/site examples

taysea:
This PR removes the wrapping/imports of grommet components to aries-core. For each of the now "unwrapped" components (those that will remain just in grommet), a simple storybook example has been created showing the component using aries theme.

It also updates the site so that imports are from grommet as opposed to aries-core.

The changes are broken up in commits for each component for clarity. If there are any additional components to unwrap from aries-core, please let me know.

closes #280


ShimiSun:
Regarding the structure of the folders, I think we shouldn't have a whole folder of Button, Accordion, etc just to contain a story file. 
can we think of a different structure to contain story files of external components?

taysea:
> Regarding the structure of the folders, I think we shouldn't have a whole folder of Button, Accordion, etc just to contain a story file.
> can we think of a different structure to contain story files of external components?

That makes sense, good feedback. Yes, let me think of a better folder structure.

taysea:
> yarn is breaking due to conflicts with yarn.lock on master

Yarn doesn't like me right now -- will clean the cache and resolve this once I get off my call.

aries-core/src/js/components/core/index.js
```diff
@@ -1,16 +1,6 @@
-export * from './Accordion';
-export * from './Button';
-export * from './Checkbox';
 export * from './Form';
```
halocline: Currently, Form is just a pass through. Should we eliminate it from `aries-core` for the time being? I know we have discussed it becoming an `aries-core` component eventually, but I think it is a quarter+ out on the roadmap. For consistency and clarity, should we treat it as the rest of the grommet components until it justifies being moved over?

taysea: Good comment. I will clean that out and take out any others that don't already have specifications.

aries-core/src/js/components/external/stories/Accordion.js
```diff
@@ -0,0 +1,35 @@
+import React from 'react';
+import { Accordion, AccordionPanel, Box, Grommet } from 'grommet';
+import { aries } from '../../../../../../aries-site/src/themes/aries';
```
halocline: Yikes! --- this import will get cleaned up when this work is completed: https://github.com/hpe-design/design-system/issues/107

aries-core/src/js/components/core/index.js
```diff
@@ -1,16 +1,6 @@
...
-export * from './MaskedInput';
 export * from './Menu';
 export * from './Nav';
-export * from './RadioButtonGroup';
-export * from './RangeInput';
 export * from './RangeSelector';
```
halocline: I think the RangeSelector is slightly modified (not 100% pass through from grommet), but could be 100% pass-through if the color specified gets moved to the theme instead.

aries-core/src/js/components/external/stories/Accordion.js
taysea: Yes (thanks goodness!)

aries-core/src/js/components/external/stories/RangeSelector.js
```diff
@@ -0,0 +1,52 @@
+import React, { useState } from 'react';
+import { RangeSelector } from 'aries-core';
```
ShimiSun: why is this story is on the `external` package while RangeSelector is part of aries-core?

taysea: Good catch. Thanks Shimi. Let me update where that file is living.

taysea:
Closing in favor of #297 and #299

Issue #202, opened 2020-01-14
Connect nextjs Link to Header button "HPE Element HPE Server"


taysea:
In order to use Next Link, it needs to be wrapped around a component. Because we are pulling in aries-core header which contains the HPE Server button as well as the search, we can't wrap Next Link directly around the HPE Server button on the far left of the header. We will need to think about how to handle this.

PR #288, opened 2020-01-23
Refactor design system Header

taysea:
This PR separates Header from Nav and creates a helper component called AppIdentity which refers to portion of a main Header that contains either the aruba icon and naming or hpe element and application name. The AppIdentity helper component is used as a child of Header which makes it easier for users (us in this case) to wrap a router link directly on that component.

AppIdentity component refers to:
Screen Shot 2020-01-23 at 11 11 06 AM
Screen Shot 2020-01-23 at 11 11 11 AM

Resource for context on the naming convention of "Identity" comes from here: https://www.muse-themes.com/blogs/news/what-is-a-website-header


taysea:
Linking to #202

aries-core/src/js/components/helpers/AppIdentity/AppIdentity.js
```diff
@@ -0,0 +1,53 @@
...
+
+    return (
+      <Button href={href} ref={ref} {...rest}>
+        <Box direction="row" align="center" gap="medium">
+          {brand === brandNames.hpe.toLowerCase() ? (
+            <Hpe color="brand" size="large" />
+          ) : (
+            undefined
+          )}
+          {brand === brandNames.aruba.toLowerCase() ? (
```
halocline: I'm not loving the repetition here. Wondering if we can do the following?
```
const brands = {
  hpe: { name: 'HPE', logo: <Hpe color="brand" size="large" /> },
  aruba: { name: 'Aruba', logo: <Aruba color="orange!" /> },
};
```
then
```
{brand && brands[brand].logo}
```

aries-core/src/js/components/helpers/AppIdentity/AppIdentity.js
```diff
@@ -0,0 +1,53 @@
...
+            <Aruba color="orange!" />
+          ) : (
+            undefined
+          )}
+          {!brandOnly && (
```
halocline: is `logoOnly` more explicit than `brandOnly`?

aries-site/src/layouts/main/Header.js
```diff
@@ -1,31 +1,37 @@
...
+
   return (
-    <Nav
-      background={background}
-      title="Design System"
-      collapse={false}
-      brandOnly={size === 'small' && searchFocused}
-    >
+    <Header background={background} {...rest}>
+      <Link href="/" passHref>
```
halocline: Liking this interface!

aries-core/src/js/components/helpers/AppIdentity/AppIdentity.js
taysea: I like.

aries-core/src/js/components/helpers/AppIdentity/AppIdentity.js
taysea: I think so, let me adjust. Good call.

halocline:
This PR will merge into your branch if you want:
https://github.com/hpe-design/design-system/pull/311

aries-core/src/js/components/core/Header/Header.js
```diff
@@ -0,0 +1,63 @@
...
+import { Header, ResponsiveContext } from 'grommet';
+
+const PAD_SIZES = ['xxsmall', 'xsmall', 'small', 'medium', 'large', 'xlarge'];
+
+const StyledHeader = ({ background, pad, ...rest }) => {
+  const size = useContext(ResponsiveContext);
+
+  return (
+    <Header
+      background={background}
```
ShimiSun: This is redundant and will be taken care of on `...rest`

aries-core/src/js/components/core/Header/Header.js
```diff
@@ -0,0 +1,63 @@
...
+      pad={
+        pad || {
+          vertical: 'medium',
+          horizontal: size !== 'small' ? 'xlarge' : 'large',
+        }
+      }
```
ShimiSun: I'm not sure the `size` adjustment assumption is right for the core library, different users may want to control the responsiveness of their app in different ways.

taysea: Users can supply pad on their own if they want to override, these would be the default paddings. I think we want to have the design system to define this padding, but there is still option provided to override if wanted.

taysea:
> I would consider not to have Header as a core component and plainly use the Grommet Header since it doesn't differ except of the suggested `pad` which I dropped a comment on.

I would think the design system wants to create defined padding for this component since its a main feature of an app and padding around the logo/title and nav elements seems like a key way that consistency could be lost if padding could be controlled -- I'm wondering if we even want to allow pad on Header to be customized.

halocline:
☝️ I had been thinking that this use case would be opinionated as well. I could see the naming of the component to be `AppHeader` which would be opinionated, but `Header` could come from Grommet and remain flexible (for use within other components such as it is being used within Tile).

aries-core/src/js/components/core/Header/Header.js
ShimiSun: I don't think that a whole component should exist for the sake of padding during responsive behavior, but I might be wrong here, would like to hear more thoughts.

taysea: I think that's a valid point, padding for a general header could probably be taken care of in theming. I think we should discuss what you brought up previously (and what Matt mentioned below) about maybe this component in the way it's being used is something more specific and would be "Header" under the hood but named something else, maybe it requires the "AppIdentity" child or has other specifications.

PR #300, opened 2020-01-28
Update icons for Components pages

taysea:
This updates the icons for the pages under "Components".


aries-site/src/components/icons/components/index.js
```diff
@@ -0,0 +1,4 @@
+export * from './IconControls';
+export * from './IconInput';
+export * from './IconMedia';
```
ShimiSun: I don't see this Icon being used anywhere, is it on the pipeline?

taysea: It's in designs, but we don't have a page for the content yet. We can remove it for now until we have the accompanying page. Would that be best? @ShimiSun

ShimiSun: I think so, yes. let's keep the code lean.

PR #313, opened 2020-01-30
Added ButtonExample

ericsoderberghp:
Check out the Button example and we can discuss.

The code link won't work in the preview until this PR is merged ;)


ShimiSun:
I haven't looked at the code yet, here are some comments regarding the functionality:
1. I liked the Icons' navigation vs. Anchors.
2. How will the Tooltips work on mobile?
3. `React Documentation` naming on the tooltip made me look for a React link vs. grommet docs.
4. There is no association of the code with its visual, in order to visualize a code the user will need to copy-paste the code from github to the editor that is placed on a different link, so I'll need to open to links to get something visualized.
5. The designers get to their useful resources in one click, while devs have a more convoluted way (they need to guess) on how to bring visualization with code together.
6. `aries-core` components will be missing on the editor, along with hooks limitation on the editor.

ShimiSun:
Round two:
1. Code inline is fun.
2. Code inline doesn't look good on the `light` mode.
3. The piece I'm still missing is the visualization around the code snippet. would opening a live editor where you can view the rendered components be a better approach?!
4. Once clicking on the *Component Properties* icon, it will be nice if it won't override the existing website page as it opens.
 5. Code inline is very specific to a component and it won't answer the needs of more complex patterns we will create in the near future.

ericsoderberghp:
> Round two:
> 
> 1. Code inline is fun.
:)
> 2. Code inline doesn't look good on the `light` mode.
Agreed.
> 3. The piece I'm still missing is the visualization around the code snippet. would opening a live editor where you can view the rendered components be a better approach?!
> 4. Once clicking on the _Component Properties_ icon, it will be nice if it won't override the existing website page as it opens.
Changed to use `target="_blank"`.
> 5. Code inline is very specific to a component and it won't answer the needs of more complex patterns we will create in the near future.
From Slack:

aries-site/src/layouts/content/Example.js
```diff
@@ -0,0 +1,164 @@
...
+      />
+      {hover && (
+        <Drop
+          target={ref.current}
+          align={{ bottom: 'top', right: 'right' }}
+          plain
+          stretch={false}
+        >
+          <Box
+            background={colors['background-contrast']}
```
ShimiSun: I think you should be able to use:
```suggestion
            background='background-contrast'
```

aries-site/src/layouts/content/Example.js
```diff
@@ -0,0 +1,164 @@
...
+              onClick={() => setShowCode(!showCode)}
+            />
+          )}
+        </Box>
+      )}
+      {showCode && (
+        <Box border background="background-contrast" pad="medium">
+          <Text size="xsmall" color="text">
+            <Syntax>
+              <code ref={codeRef} className="language-jsx">
```
ShimiSun: First time I see you are using `className` :)

aries-site/src/examples/components/controls/ButtonExample.js
```diff
@@ -0,0 +1,22 @@
...
+import { Next } from 'grommet-icons';
+
+export const ButtonExample = () => {
+  return (
+    <Box direction="column" align="start" gap="large">
+      <Button
+        label="Check it out!"
+        icon={<Next />}
+        reverse
+        primary
```
ShimiSun: I think we said we should have one or two instances of `primary` Button, I think the ButtonExample may be used more than a couple of times.

aries-site/src/layouts/content/Example.js
ericsoderberghp: Yep, required by `prismjs`.

aries-site/src/layouts/content/Example.js
```diff
@@ -0,0 +1,164 @@
...
+  // so the styling change is rendered.
+  React.useEffect(() => setSyntax(syntax[theme.dark ? 'dark' : 'light']), [
+    theme.dark,
+  ]);
+
+  return (
+    <Box margin={{ vertical: 'small' }}>
+      <Box
+        direction="row"
+        background={colors['background-front']}
```
ShimiSun: Same here:
background='background-front'

Issue #256, opened 2020-01-17
Home page tiles should be anchors


karatechops:
Href attributes should only live in `a` tags. I believe this is also breaking the keyboard navigation tab flow. I will open a separate issue for this.

![Screen Shot 2020-01-17 at 5 30 11 PM](https://user-images.githubusercontent.com/5983843/72650659-1ffb8b80-394f-11ea-9589-8189b593c3d2.png)

![Screen Shot 2020-01-17 at 5 32 11 PM](https://user-images.githubusercontent.com/5983843/72650737-53d6b100-394f-11ea-9081-124aff847466.png)

taysea:
I think these aren't explicitly anchors or buttons because it is using the onClick prop that was recently added to grommet Box. However, this is a good point about where hrefs should appear and we should check on that. For context, the href that is appearing is being passed down from Nextjs Link that's wrapping each of these tiles.

The tabbing issues aren't because of the fact that it's a div though, because grommet handles keyboard for boxes with 'onClick' props, but is coming from a different issue with how Drop changes tabindex of other elements (link to a PR is on the other issue you opened about this).

PR #316, opened 2020-01-30
Make Tiles and NavPage items <a> in DOM

taysea:
Motivated by Issue #256: In order for links to be crawlable, elements with href need to be <a> in the DOM. This PR adjust the home tiles and navpage items to be <a href=... /> in the DOM.

Resource here: https://support.google.com/webmasters/answer/9112205?hl=en

### DIFF
--- a/aries-site/src/layouts/navigation/NavPage.js
+++ b/aries-site/src/layouts/navigation/NavPage.js
@@ -1,7 +1,7 @@
 import React, { useContext } from 'react';
 import Link from 'next/link';
 import PropTypes from 'prop-types';
-import { Box, Text, Paragraph, ResponsiveContext } from 'grommet';
+import { Box, Button, Text, Paragraph, ResponsiveContext } from 'grommet';
 import { iconsMap } from '../../components/icons/iconsMap';
 import { structure } from '../../data';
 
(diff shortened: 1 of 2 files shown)


taysea:
Will close: #256

aries-site/src/layouts/navigation/NavPage.js
```diff
@@ -16,21 +16,26 @@ const NavItem = ({ item, topic }) => {
...
+        <Box
+          direction="row"
+          margin={{ vertical: 'large' }}
+          gap={size !== 'small' ? 'large' : 'medium'}
+        >
+          {/* Adds placeholder icon for the meantime
+           * while we wait to get all the icons */}
+          {itemData.icon
+            ? itemData.icon('xlarge')
+            : iconsMap.branding('xlarge')}
```
halocline: Is this condition still needed? Opportunity to eliminate while we are in here?

aries-site/src/layouts/navigation/NavPage.js
```diff
@@ -16,21 +16,26 @@ const NavItem = ({ item, topic }) => {
...
-        {/* Adds placeholder icon for the meantime
-         * while we wait to get all the icons */}
-        {itemData.icon ? itemData.icon('xlarge') : iconsMap.branding('xlarge')}
-        <Box>
-          <Text weight="bold" size={size !== 'small' ? 'large' : 'medium'}>
-            {itemData.name}
-          </Text>
-          <Paragraph size="small">{itemData.description}</Paragraph>
+      {/* Needs to be <a> in DOM for web crawling: https://support.google.com/webmasters/answer/9112205?hl=en */}
+      <Button fill>
```
halocline: This approach makes sense to me.

aries-site/src/layouts/navigation/NavPage.js
taysea: The only reason I would leave it is if we think there'd be a case where a page didn't have an icon yet and it we want to leave it as a fallback -- at least we'd have confidence that the page layout would look correct as pages get added.

aries-site/src/pages/index.js
```diff
@@ -25,11 +25,15 @@ const HomeTiles = ({ ...rest }) => {
...
   return (
+    // Needs to be <a> in DOM for web crawling: https://support.google.com/webmasters/answer/9112205?hl=en
     <Tile
+      as="a"
       pad="medium"
       background={topic.color}
       key={topic.color}
       ref={ref}
+      fill
+      style={{ textDecoration: 'none' }}
```
ShimiSun: I'd consider creating a new styled component of <AnchoredTile /> that will have a combination of lines 30, 35 and 36, to distinguish it from a regular tile, not a must but cleaner and more readable.

aries-site/src/layouts/navigation/NavPage.js
halocline: I believe the `passHref` prop from line 18 is getting passed to the button, which then is being treated as `<Button>` with an `href` prop.

taysea: Yes, what Matt said is correct, the passHref puts the href on the button and then Grommet knows to make it an <a> in the DOM.

aries-site/src/pages/index.js
taysea: Pulling over my reply on slack:
I'm open to that, I didn't want to refactor to another component if it didn't seem necessary, but if I were to, I could call it `AnchorTile`?

ShimiSun: `AnchorTile` sounds good to me.

aries-site/src/layouts/navigation/NavPage.js
ShimiSun: Correct, I was looking for the `href` on the Button level.  thanks for clarifying.

PR #327, opened 2020-02-03
Remove old HPE theme dependency

ShimiSun:
Currently, the new HPE theme is using the old theme as a fallback, I'm removing the old theme so we can make sure we are no longer dependent on it.

### DIFF
--- a/aries-site/src/themes/aries.js
+++ b/aries-site/src/themes/aries.js
@@ -173,6 +173,7 @@ export const aries = {
     },
   },
   anchor: {
+    color: 'text',
     fontWeight: 700,
     textDecoration: 'none',
     hover: {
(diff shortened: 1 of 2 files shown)


aries-site/src/themes/aries.js
```diff
@@ -19,7 +19,8 @@ const fontSizing = factor => {
   };
 };
 
-export const aries = deepMerge(hpe, {
+// export const aries = deepMerge(hpe, {
```
halocline: Anchor loses color (need to set anchor color = 'text' so that it is white): 
Screen Shot 2020-02-03 at 3 49 00 PM

halocline: Screen Shot 2020-02-03 at 3 55 56 PM

taysea: Good catch on that

ShimiSun:
Added `color: 'text'` for Anchor, good live visual testing ;)

We don't have issues with the `status-*` colors, its all define in colors.

aries-site/src/themes/aries.js
```diff
@@ -173,6 +173,7 @@ export const aries = {
     },
   },
   anchor: {
+    color: 'text',
```
halocline: 👍 
Looks good on my end.

PR #333, opened 2020-02-05
Lay groundwork for TestCafe

taysea:
This PR creates a folder `src/tests` that will contain all the test files for the site. The package.json test script was updated. Currently, the tests are using the flag `all` which will run tests on all locally installed browsers. However, in the future, we should look to connect this with BrowserStack to be able to test on all browsers we are interested in.

Note: Refinements will be made to these tests. However, they are just put in place so Brittany can get started writing tests as well.

### DIFF
--- a/aries-site/package.json
+++ b/aries-site/package.json
@@ -22,7 +22,7 @@
     "export": "yarn build && next export",
     "postexport": "node next.postexport.js",
     "start": "next",
-    "test": "jest",
+    "test": "testcafe all src/tests/",
     "lint": "eslint src",
     "lint-fix": "eslint src --fix"
   },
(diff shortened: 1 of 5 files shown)


aries-site/package.json
```diff
@@ -31,6 +31,7 @@
     "test"
   ],
   "devDependencies": {
-    "babel-plugin-styled-components": "^1.10.6"
+    "babel-plugin-styled-components": "^1.10.6",
+    "testcafe": "^1.8.1"
```
ShimiSun: would it make sense to use `testCafe` on the core library? if so this devdependency would be a candidate to the parent package.json

PR #335, opened 2020-02-05
Aligning the HPE Theme changes

ShimiSun:
Before publishing the new HPE theme, I've updated the custom theme with the HPE theme changes, let's see it is still working as expected.


aries-site/src/themes/aries.js
```diff
@@ -188,30 +168,195 @@ export const aries = {
...
+          maxWidth: '278.4px',
+        },
+        xlarge: {
+          size: '11.6px',
+          height: '17.6px',
+          maxWidth: '278.4px',
+        },
       },
     },
-    weight: 700,
```
halocline: We will want to keep heading font-weight = 700 for the bold appearance.

ShimiSun: `weight` is specified on the base.js theme on 600, and I think we decided it to use it from there.

ericsoderberghp: I'm fine if the design side wants 700 for the HPE theme, leaving grommet base at 600.

aries-site/src/themes/aries.js
```diff
@@ -250,15 +395,41 @@ export const aries = {
...
-  paragraph: {
-    small: fontSizing(0),
-    medium: fontSizing(1),
-    large: fontSizing(3),
+    xsmall: {
+      size: '8.4px',
+      height: '14.4px',
+      maxWidth: '201.6px',
+    },
+    small: {
```
taysea: Noticing this size of "11.6" doesn't match what was previously defined as "18" -- when I run the size, the text isn't coming through at the right values. Here is what a home tile is looking like. I am fine with hard-coding but let's make sure the values match what the previous calculations mapped to (unless there has been rationale to shift these values)

Screen Shot 2020-02-05 at 10 38 57 AM

aries-site/src/themes/aries.js
ShimiSun: Thanks @halocline and @ericsoderberghp , theme got updated.

aries-site/src/themes/aries.js
ShimiSun: Finalizing Text/HEading/Paragraph sizes is currently work in progress, thanks for raising it. I'll keep this PR updated with upcoming changes.

aries-site/src/themes/aries.js
```diff
@@ -212,6 +332,33 @@ export const aries = {
...
     },
   },
+  paragraph: {
+    small: {
+      size: '14px',
+      height: '20px',
+      maxWidth: '336px',
+    },
+    medium: {
+      size: '18px',
```
halocline: This should be bumped up to 24px.

aries-site/src/themes/aries.js
```diff
@@ -250,15 +397,42 @@ export const aries = {
...
+      height: '18px',
+      maxWidth: '288px',
+    },
+    small: {
+      size: '14px',
+      height: '20px',
+      maxWidth: '336px',
+    },
+    medium: {
+      size: '18px',
```
halocline: Because 'medium' is the default, I would recommend that it be defined as the 24px size, then small = 18px.

aries-site/src/themes/aries.js
ShimiSun: We need to make sure this bump/change is derived from the `scale` and `spacing` parameters and not just changed manually (though it could have been much easier :) )

aries-site/src/themes/aries.js
ShimiSun: I agree, checking what is the right scaling to get the site values back. At the moment I'm thinking it is relying on the factors. WIP

taysea:
I am fine with the text values, but we will need to go into the site and change the sizes we have applied to the fonts, for example the NavPages need the title of "Layout" now to be Text `xxlarge` to maintain same sizing and the description needs to be `medium` instead of `small`. Notice how the text is smaller now than it is on the live site.

This branch:
Screen Shot 2020-02-06 at 11 57 37 AM

Live site:
Screen Shot 2020-02-06 at 12 01 57 PM

Do we plan to fix those text sizes on the site before merging this? @ShimiSun

ShimiSun:
> I am fine with the text values, but we will need to go into the site and change the sizes we have applied to the fonts, for example the NavPages need the title of "Layout" now to be Text `xxlarge` to maintain same sizing and the description needs to be `medium` instead of `small`. Notice how the text is smaller now than it is on the live site.
> 
> This branch:
> Screen Shot 2020-02-06 at 11 57 37 AM
> 
> Live site:
> Screen Shot 2020-02-06 at 12 01 57 PM
> 
> Do we plan to fix those text sizes on the site before merging this? @ShimiSun

Yep, regardless, this Text should be Heading for semantic accuracy. I'll push a fix for it. Good Catch!

ShimiSun:
Discussed and approved with @L0ZZI & @taysea , merging.

PR #338, opened 2020-02-06
Add Lighthouse CI action

karatechops:


.github/workflows/main.yml
```diff
@@ -0,0 +1,15 @@
+name: Lighthouse audit
+on: pull_request
+
+jobs:
+  audit:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Lighthouse audit Netlify deploy preview
+        uses: jakejarvis/lighthouse-action@master
```
ShimiSun: Any reason you chose this particular source https://github.com/jakejarvis/lighthouse-action? I noticed there are others available on the web but I haven't done a full comparison myself.

karatechops: The other options don't integrate with Netlify unfortunately. We could fork a more full-featured option and add [this check](https://github.com/jakejarvis/lighthouse-action/blob/master/entrypoint.sh#L8-L14).

PR #341, opened 2020-02-06
Connect with Browserstack

taysea:
This PR connects with browserstack to allow us to test on various browsers. However, we have limited number of testing minutes (200min/month) with our current plan. I am looking at how to optimize these minutes. I created two scripts:
1. `yarn test` will connect to browserstack and run on the browsers we provide
2. `yarn test:local` will run locally only and not use any testing minutes. However, it can only run on locally downloaded browsers. For most of us, that is likely Chrome, Safari, and Firefox. This should be sufficient to test basic changes while working on the site and will conserve testing minutes. **When running tests while working, aim to use `yarn test:local` when possible.**

### DIFF
--- a/aries-site/src/scripts/start-tests.js
+++ b/aries-site/src/scripts/start-tests.js
@@ -1,0 +1,59 @@
+const createTestCafe = require('testcafe');
+const BrowserStack = require('browserstack');
+
+const browserStackCredentials = {
+  username: process.env.BROWSERSTACK_USERNAME,
+  password: process.env.BROWSERSTACK_PASSWORD,
+};
+
+// TBD: What browsers to we want to test against
+const SUPPORTED_BROWSERS = ['browserstack:firefox@72.0:Windows 10'];
+
+async function getRunningBrowserstackSessions() {
+  const client = BrowserStack.createClient(browserStackCredentials);
+  const workerStatus = await new Promise((resolve, reject) => {
+    client.getApiStatus((error, workers) => {
+      if (error) reject(error);
+      else resolve(workers);
+    });
+  });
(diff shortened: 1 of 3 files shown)


aries-site/src/scripts/start-tests.js
```diff
@@ -0,0 +1,59 @@
+const createTestCafe = require('testcafe');
+const BrowserStack = require('browserstack');
+
+const browserStackCredentials = {
+  username: process.env.BROWSERSTACK_USERNAME,
+  password: process.env.BROWSERSTACK_PASSWORD,
+};
+
+// TBD: What browsers to we want to test against
+const SUPPORTED_BROWSERS = ['browserstack:firefox@72.0:Windows 10'];
```
ShimiSun: Can we add more browsers and use other OS on the same runs?
I'm thinking Chrome on both WIN and MAC is a must.

taysea: Yes, we should definitely add more. I just did one for now when I was checking to make sure that it was connecting to the account correctly.

ShimiSun: We'll probably need to find the right line on how many combinations of OS+browsers we should support without effecting the build time too bad. will be interesting to see how it grows.

britt6612: I think if the majority of us have Chrome on our desktop we shouldn't use it in the Supported Browsers on the browser stack I think if we are limited on min/month we should test the browsers that we don't have like ie and edge and maybe Firefox but I think most of us have chrome and safari on our local computer. 

Something to discuss in the future as we get closer to that min/month limit

PR #342, opened 2020-02-07
Helper PR to #340

taysea:
Just made a couple tweaks to remove that hard coded route and adjust naming to better match the component.


taysea:
I think we are good on the tests you wrote for now, I would like to improve the tests as we continue to learn TestCafe to make them more dynamic for keyboard testing. Taken from our offline discussion:

I think what would be ideal though is that after each tab, we could get the href of the focused element and then if it matches the expectedPath, then we hit enter.

PR #346, opened 2020-02-07
Do Not Merge - Feat build process

halocline:
Related Issues: https://github.com/hpe-design/design-system/issues/301, https://github.com/hpe-design/design-system/issues/199

**Background on Approach**
Looking at optimizing the "build" process in three areas:

1. Prior to build (i.e. catch issues earlier in the development process)
  - Check Grommet/stable version in Yarn.lock, update if newer commits
  - Run linter
  - Run tests
    - Jest (aries-core)
    - TestCafe (aries-site)
  **Actions:** --> 
  - Implement checks with Git Hooks via Husky (I initially was looking at "pre-commit" but chose "husky" because it seemed to be play nicer and was more intuitive in monorepo environment).
  - Run linter script
  - Run test script (this is disabled for now because of a failing test and non-existent tests on aries-core)
2. During build (i.e. evaluate which dependencies are bundled, caching, etc.)
  - Move any applicable dependencies to peerDependencies or devDependencies
  - Cache dependencies
  **Actions:** --> 
  - Update `package.json`s
3. When do builds get triggered
  - Do we run builds on all branches? Or just specific ones?
  **Actions**: --> 
  - Recommend leave alone for now. Deploy previews are useful for PR and stakeholder reviews. Opportunity in the future if needed.

### DIFF
--- a/aries-core/package.json
+++ b/aries-core/package.json
@@ -45,6 +45,7 @@
     "test": "jest",
     "storybook": "start-storybook -p 6006",
     "build-storybook": "build-storybook",
+    "preinstall": "cd .. && node grommet-sync.js",
     "release-stable": "babel-node ./tools/release-stable"
   },
   "keywords": [
--- a/aries-core/package.json
+++ b/aries-core/package.json
@@ -45,7 +45,6 @@
     "test": "jest",
     "storybook": "start-storybook -p 6006",
     "build-storybook": "build-storybook",
-    "preinstall": "cd .. && node grommet-sync.js",
     "release-stable": "babel-node ./tools/release-stable"
   },
   "keywords": [
--- a/aries-core/package.json
+++ b/aries-core/package.json
@@ -7,7 +7,9 @@
   "dependencies": {
     "grommet": "https://github.com/grommet/grommet/tarball/stable",
     "grommet-icons": "^4.4.0",
-    "grommet-theme-hpe": "https://github.com/grommet/grommet-theme-hpe/tarball/stable",
+    "grommet-theme-hpe": "https://github.com/grommet/grommet-theme-hpe/tarball/stable"
+  },
+  "peerDependencies": {
     "react": "^16.12.0",
     "react-dom": "^16.12.0",
     "styled-components": "^5.0.0"
@@ -28,7 +30,10 @@
     "del": "^5.1.0",
     "dotenv": "^8.2.0",
     "fs-extra": "^8.1.0",
+    "react": "^16.12.0",
+    "react-dom": "^16.12.0",
     "simple-git": "^1.126.0",
+    "styled-components": "^5.0.0",
     "webpack": "^4.29.3",
     "webpack-cli": "^3.2.3",
     "webpack-dev-server": "^3.1.14"
@@ -42,10 +47,6 @@
     "build-storybook": "build-storybook",
     "release-stable": "babel-node ./tools/release-stable"
   },
-  "pre-commit": [
-    "lint-fix",
-    "test"
-  ],
   "keywords": [
     "aries",
     "aries-core",
--- a/aries-site/src/data/structure.js
+++ b/aries-site/src/data/structure.js
@@ -11,7 +11,6 @@ import {
   IconHumanCentered,
   IconIcons,
   IconInput,
-  IconLayout,
   IconLayoutComponents,
   IconPhilosophy,
   IconPrinciples,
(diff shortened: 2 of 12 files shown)


aries-site/src/data/structure.js
```diff
@@ -11,7 +11,6 @@ import {
   IconHumanCentered,
   IconIcons,
   IconInput,
-  IconLayout,
```
halocline: FYI, linter fix

aries-site/src/examples/components/input/UsageExample.js
```diff
@@ -2,7 +2,7 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import { Box, ResponsiveContext, Text } from 'grommet';
 
-import { colors } from '../../themes/aries';
+import { colors } from '../../../themes/aries';
```
halocline: FYI, linter fix

aries-site/src/pages/components/controls.js
```diff
@@ -1,9 +1,8 @@
 import React from 'react';
 import { Box, Button } from 'grommet';
 
-import { ExternalCTA, Meta, SubsectionText } from '../../components';
```
halocline: FYI, linter fix

aries-site/src/pages/components/controls.js
```diff
@@ -1,9 +1,8 @@
...
+import { Meta, SubsectionText } from '../../components';
 import {
-  ButtonRow,
```
halocline: FYI, linter fix

aries-site/src/pages/components/input.js
```diff
@@ -8,7 +8,6 @@ import {
   MaskedInputExample,
   RadioButtonGroupExample,
   RangeInputExample,
-  RangeSelectorExample,
```
halocline: FYI, linter fix

aries-site/src/utils/hooks/useEventListener.js
```diff
@@ -13,8 +13,6 @@ export const useEventListener = (eventName, handler, element = global) => {
 
     const eventListener = event => savedHandler.current(event);
     element.addEventListener(eventName, eventListener);
-    return () => {
```
halocline: FYI, linter fix

ShimiSun:
Will this approach help us with the frequent yarn breaks? how so?

aries-core/package.json
```diff
@@ -42,10 +47,6 @@
     "build-storybook": "build-storybook",
     "release-stable": "babel-node ./tools/release-stable"
   },
-  "pre-commit": [
-    "lint-fix",
-    "test"
-  ],
```
ShimiSun: Is the `pre-commit` removed? I was actually thinking we should add "prettier" on pre-commit to help with code structure consistency.

halocline: I would like to add Prettier as well.

We actually never had installed `pre-commit` as a `devDependency` and thus these config lines for `pre-commit` weren't being executed.

I am proposing that we use `husky` instead of `pre-commit`. They provide similar functionality, however I found `husky` to be easier to configure in a monorepo.

The scripts `lint-fix`, `test`, etc. will now be triggered by the `.huskyrc.js` file.

halocline:
> Will this approach help us with the frequent yarn breaks? how so?

Yes, the frequent yarn breaks are intended to be minimized by this work.

**Background**
As reminder, the yarn breaks are caused because of the following: 
- Each time a new commit is made to Grommet Stable, a new sha-1 signature is assigned to the commit. When this project's CI build process goes to retrieve the dependency, it compares the checksums for the sha-1 it receives to the sha-1 signature in the yarn.lock / build cache. If the signatures do not match, the CI build will fail. Because Grommet is being actively developed, the sha-1 changes frequently resulting in mismatching sha-1s regularly, and thus, regularly failed CI builds.

**How so?**
- This approach assumes our project's Grommet in `yarn.lock` will always be out of date (because Grommet is being actively developed resulting in the checksum mismatch mentioned above).
- Under this assumption, this approach does the following:
  - Upon each `git commit` (triggers pre-commit hook):
    - Remove "grommet@https://github.com/grommet/grommet/tarball/stable" from `yarn.lock` (this is performed in `grommet-sync.js`)
    - Install the current version of "grommet@https://github.com/grommet/grommet/tarball/stable"
    - Proceeds with rest of pre-commit checks

This approach avoids having to remove the entire contents of yarn.lock and then having to re-install all dependencies.

The limitation of this approach occurs in situations where a PR sits unmerged for a few days. In these cases, the PR may be exposed to mis-matching grommet versions as well.  I think we can explore implementing the same technique used here, but triggering them at another point in our workflow (e.g. a hook at on approval).

halocline:
Closing. This work was executed in https://github.com/hpe-design/design-system/pull/370 && https://github.com/hpe-design/design-system/pull/369 && https://github.com/hpe-design/design-system/pull/363

Issue #251, opened 2020-01-17
Color contrast audit


karatechops:
![Screen Shot 2020-01-17 at 5 18 54 PM](https://user-images.githubusercontent.com/5983843/72650116-7a93e800-394d-11ea-824f-d7ba140c3026.png)

![Screen Shot 2020-01-17 at 5 19 40 PM](https://user-images.githubusercontent.com/5983843/72650153-9a2b1080-394d-11ea-97b8-83622828cda7.png)

![Screen Shot 2020-01-17 at 5 19 27 PM](https://user-images.githubusercontent.com/5983843/72650154-9a2b1080-394d-11ea-9aa3-08ea1a37fc8e.png)

![Screen Shot 2020-01-17 at 5 19 18 PM](https://user-images.githubusercontent.com/5983843/72650155-9ac3a700-394d-11ea-9d4e-8fa436d768d5.png)

taysea:
Also related: #200

taysea:
I did an audit on all of the pages in both dark and light mode for color contrast. These are the places where there was an explicit error for color contrast. However, in addition to these I think we should address the above noted areas the subjectively lack contrast (Such as the search placeholder text).

**Dark Mode**

- Landing page: Descriptive text under develop
<img width="363" alt="Screen Shot 2020-01-23 at 3 18 42 PM" src="https://user-images.githubusercontent.com/12522275/73035555-5b89d000-3dfd-11ea-9e45-f9d5d4678a8d.png">

- Layout page under Components: Stack example
<img width="109" alt="Screen Shot 2020-01-23 at 4 28 51 PM" src="https://user-images.githubusercontent.com/12522275/73035586-78be9e80-3dfd-11ea-97e7-1e030378b1a8.png">

- Develop: HPE Design system text
<img width="273" alt="Screen Shot 2020-01-23 at 4 29 41 PM" src="https://user-images.githubusercontent.com/12522275/73035604-95f36d00-3dfd-11ea-8dc5-987fb2bf24db.png">

**Light Mode**

- Landing: Descriptive text under Foundation and Design
<img width="373" alt="Screen Shot 2020-01-23 at 4 34 27 PM" src="https://user-images.githubusercontent.com/12522275/73035765-43668080-3dfe-11ea-89ec-df40e7f56b51.png">
[...]

PR #358, opened 2020-02-11
Add accessibility testing with axe-testcafe

taysea:
Adds accessibility tests with "axe-testcafe" which tests accessibility rules defined by wcag2a, wcag2aa, wcag2.1a, wcag2.1aa, and section508. I have made adjustments to certain files to fix accessibility issues that were flagged.

Not every route is being tested, but the various page layouts are as well as pages with variables content that might not appear on all pages (in this case Branding).

NOTE: Because color contrast errors identified in issue #251 would cause these tests to fail, those page elements have been excluded from the tests temporarily to allow the rest of accessibility tests to run.

To-do:
- Discuss scope of routes to check accessibility on
- Test Darkmode for color contrast only


britt6612:
@taysea thank you for adding this! Tests pass for me!

PR #365, opened 2020-02-13
Restructure tests

taysea:
Restructures tests based on discussion below on best practices:

e2e Testing Best Practices
https://www.lambdatest.com/blog/all-you-need-to-know-about-end-to-end-testing/

1. Identify User Functions: features such as logging in, searching, logging out, etc.
2. Identify Conditions for each user function (includes timing, data, etc.): 
    1. Did error display correctly?
    2. Did validation happen?
3. Write Test cases: each test case should test a single condition

Translated to TestCafe —> each “fixture” should be a user function, each “test” should be a test case


britt6612:
Look good now we can put ours into each folder.

PR #367, opened 2020-02-13
Organize nav

britt6612:
This PR gets rid of all of the data-test-id's 

As well as organizes the nav folder


britt6612:
Thank you! I saw that I was actually going to comment to look at the topic comment thanks for fixing!

PR #370, opened 2020-02-14
Feat pre commit

halocline:
Related Issues: https://github.com/hpe-design/design-system/issues/301

This PR is a subset of https://github.com/hpe-design/design-system/pull/346, with the hope it is more focussed and easier to understand.

Work Included: 
- Adds pre-commit checks for linting and tests via `husky` (husky is a package similar to `pre-commit`.  I found `husky` easier to configure in a monorepo than `pre-commit`)
  - Removes `pre-commit` configuration from package.json. This was not functional as `pre-commit` was never added as a dependency.
  - `.huskyrc.js` becomes the configuration for pre-commit commands to be run. Additional commands are simply added to the `tasks` array.
- Switches `testcafe` port to run on 3030 -- which allows a developer to leave their dev server running on 3000.

Let me know if a call would be helpful for your reviews.


aries-site/package.json
```diff
@@ -23,7 +23,7 @@
     "start": "next",
     "start-server": "yarn build && yarn next start -p 3030",
     "test": "yarn build && node src/scripts/start-tests.js",
-    "test:local": "start-server-and-test start-server 3030 'testcafe all src/tests/'",
+    "test:local": "start-server-and-test start-server 3030 'testcafe chrome,firefox src/tests/'",
```
ShimiSun: I might be missing something shouldn't `test:local`  be added to the package.json pre-commit section to perform as a pre-commit?

ShimiSun: Another separate thought; Since we have a monorepo and I would think that for any core library change we want the pre-commit to run on the website as well, maybe it should mean the pre-commit command should be found on the parent package.json?

halocline: @ShimiSun Take a look at `huskyrc.js` which sits at the monorepo root. It serves as the controller, which is currently configured to do what you are suggesting (i.e. run aries-site tests even when changes are made to aries-core).

Using `husky` package instead of `pre-commit` because I found it to easier to implement and understand in monorepo.

halocline: `test:local` is called by `huskyrc.js` -- it is an alternative to the `pre-commit` package.

PR #371, opened 2020-02-14
Don't merge — Create script to create structure file for test

taysea:
trying to create file that eliminates the "icon" property of the structure array so that TestCafe can utilize data structure for tests.


taysea:
Closing because this approach has been de-prioritized and this PR was to allow other team members to see an attempted approach, not because it was a successful solution

PR #373, opened 2020-02-18
add deep linking tests

britt6612:
This PR adds the tests for the deep linking.

Im using the boundingClientRectProperty to get the top measurement of the visual on the page which should be close to 0 when the sub links are correctly displayed. There is a little range in there now so that the tests do not fail. 

Commented out the test with the search since that test will always fail because it goes to the top of the page until issue #320 is fixed.

### DIFF
--- a/aries-site/src/tests/navigation/deeplinks.js
+++ b/aries-site/src/tests/navigation/deeplinks.js
@@ -4,7 +4,7 @@ import { ReactSelector, waitForReact } from 'testcafe-react-selectors';
 import { Selector } from 'testcafe';
 import { baseUrl, Search } from '../utils';
 
-// A Deeplink is a subsection of the sections. This is testing if the page is taken to the right section of page to match the hash
+// A deeplink is a subsection of a page. These tests check if the website is taken to the right scroll position of the page to match the hash of the intended subsection.
 
 const getSuggestion = page => {
   return ReactSelector(`${Search} StyledDrop Button`).withText(page);
--- a/aries-site/src/tests/navigation/deeplinks.js
+++ b/aries-site/src/tests/navigation/deeplinks.js
@@ -2,18 +2,15 @@
 /* eslint-disable no-undef */
 import { ReactSelector, waitForReact } from 'testcafe-react-selectors';
 import { Selector } from 'testcafe';
-import { baseUrl } from '../utils';
+import { baseUrl, Search } from '../utils';
 
 // A Deeplink is a subsection of the sections. This is testing if the page is taken to the right section of page to match the hash
-
-// Selector name found using Chrome React dev tools on prod mode of website
-const Search = 'D Box';
 
 const getSuggestion = page => {
   return ReactSelector(`${Search} StyledDrop Button`).withText(page);
 };
 
-fixture('Search')
+fixture('Deep Linking')
   .page(baseUrl)
   .beforeEach(async () => {
     await waitForReact();
@@ -39,8 +36,6 @@ page therefore it is failing. Going to leave commented out for now.
 /*Added range to this test since now there is some issues with different browsers as to where
 the page actually lands when routed to a deep link
 */
-
-fixture('Deep linking').page(baseUrl);
 
 test('should navigate to correct hash of page when a deep link is directly routed to', async t => {
   const url = `${baseUrl}/foundation/color#background-colors`
--- a/aries-site/src/tests/navigation/deeplinks.js
+++ b/aries-site/src/tests/navigation/deeplinks.js
@@ -1,0 +1,52 @@
+
+/* eslint-disable no-undef */
+import { ReactSelector, waitForReact } from 'testcafe-react-selectors';
+import { Selector } from 'testcafe';
+import { baseUrl } from '../utils';
+
+// A Deeplink is a subsection of the sections. This is testing if the page is taken to the right section of page to match the hash
+
+// Selector name found using Chrome React dev tools on prod mode of website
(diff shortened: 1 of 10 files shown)


aries-site/src/tests/navigation/search.js
```diff
@@ -75,3 +75,19 @@ test('should navigate to correct page when user is only using keyboard ', async
...
+//   const page = 'Background Colors';
+//   const backgroundColors = Selector('#background-colors')
+//   const suggestion = getSuggestion(page);
+
+//   await t
+//     .typeText(ReactSelector(Search), 'col')
+//     .click(suggestion)
+//     .expect(backgroundColors.getBoundingClientRectProperty("top"))
+//     .within(-150, 150);
+// });
```
ShimiSun: why is this block needed?

aries-site/src/tests/navigation/homepage.js
```diff
@@ -27,3 +27,18 @@ test('should navigate to correct path when a tile is choosen via keyboard', asyn
...
+
+fixture('Deep linking').page(baseUrl);
+
+test('should navigate to correct hash of page when a deep link is directly routed to', async t => {
+  const url = `${baseUrl}/foundation/color#background-colors`
+  const pageSection = Selector('#background-colors');
+  await t
+    .navigateTo(url)
+    .expect(pageSection.getBoundingClientRectProperty("top"))
+    .within(-150, 150);
```
ShimiSun: How those numbers had been chosen?

aries-site/src/tests/navigation/search.js
britt6612: This block tests if the user for example searches background-colors from the home page it should take them to the subsection #background-colors of the page. Currently it takes the user to the top of the page which needs to be fixed from issue #320 I commented out so our tests wont fail since we know what the problem is.

aries-site/src/tests/navigation/homepage.js
britt6612: I got these numbers from  doing `document.getElementById('background-colors').getBoundingClientRect()` again this should be equal to 0 but since we have that deep linking bug I did a wider range so that way the tests don't always fail until we get that fixed

aries-site/src/tests/navigation/search.js
britt6612: If it makes sense to keep this out for now until we fix issue #320 then I can save it locally and add it back in once that is fixed.

ShimiSun: If we are planning on keeping this comment, add an inline explanation of why it is commented and what required to have it back in.

britt6612: okay I did add 
This test will fail until issue #320 is fixed. The page scroll is at the incorrect section of the 
page therefore it is failing. Going to leave commented out for now. 
 
on line 79-80 not sure if you saw that but I can explain further

aries-site/src/tests/navigation/deeplinks.js
```diff
@@ -0,0 +1,52 @@
+
+/* eslint-disable no-undef */
+import { ReactSelector, waitForReact } from 'testcafe-react-selectors';
+import { Selector } from 'testcafe';
+import { baseUrl } from '../utils';
+
+// A Deeplink is a subsection of the sections. This is testing if the page is taken to the right section of page to match the hash
```
taysea:
```suggestion
// A deeplink is a subsection of a page. These tests check if the website is taken to the right scroll position of the page to match the hash of the intended subsection.
```

aries-site/src/tests/navigation/deeplinks.js
```diff
@@ -0,0 +1,52 @@
...
+
+// Selector name found using Chrome React dev tools on prod mode of website
+const Search = 'D Box';
+
+const getSuggestion = page => {
```
taysea: Should Search and getSuggestion be pulled into utils if it's being used across files?

aries-site/src/tests/navigation/deeplinks.js
```diff
@@ -0,0 +1,52 @@
...
+  return ReactSelector(`${Search} StyledDrop Button`).withText(page);
+};
+
+fixture('Search')
```
taysea: I think my thought is that this file would have one fixture called "Deep linking" and that the two tests would fall beneath it. What do you think?

aries-site/src/tests/navigation/deeplinks.js
britt6612: I kinda thought the same since its being used now in 2 places where as before just one. I can move it

aries-site/src/tests/navigation/deeplinks.js
britt6612: I see your point I agree I will change that

aries-site/src/tests/navigation/search.js
```diff
@@ -1,10 +1,7 @@
 /* eslint-disable max-len */
 /* eslint-disable no-undef */
 import { ReactSelector, waitForReact } from 'testcafe-react-selectors';
-import { baseUrl, getLocation, repeatKeyPress } from '../utils';
-
-// Selector name found using Chrome React dev tools on prod mode of website
-const Search = 'D Box';
+import { baseUrl, getLocation, repeatKeyPress, Search } from '../utils';
 
 const getSuggestion = page => {
```
taysea: This is repeated in search and deeplinks -- seems like it could also move to utils

taysea: getSuggestion

Issue #180, opened 2020-01-10
Update light-mode background colors


halocline:

taysea:
Will be closed by: https://github.com/grommet/grommet/pull/3735

Issue #355, opened 2020-02-11
No background fill on light theme examples.


L0ZZI:
When switching from dark to light theme, the background box isn't using the background-front prop? is this part of #180 ?

![Screen Shot 2020-02-11 at 6 57 37 AM](https://user-images.githubusercontent.com/1753301/74248170-06263d80-4c9c-11ea-83ea-b39e1b41a5bb.png)
![Screen Shot 2020-02-11 at 6 57 33 AM](https://user-images.githubusercontent.com/1753301/74248172-06bed400-4c9c-11ea-83f1-61158758ef81.png)

halocline:
@L0ZZI Yes, https://github.com/hpe-design/design-system/issues/180 will fix this. `background-back` `#EFEFEF` will be applied and `background-front` will remain `#FFFFFF`.

PR #380, opened 2020-02-18
Apply grommet background prop

taysea:
Add `background="background-back"` to Grommet component so that light theme background applies correctly.

Closes: #180 #355

### DIFF
--- a/aries-site/src/components/content/ThemeMode.js
+++ b/aries-site/src/components/content/ThemeMode.js
@@ -15,6 +15,7 @@ export const ThemeMode = ({ children }) => {
 
   const body = (
     <Grommet
+      background="background-back"
       theme={aries}
       themeMode={themeMode}
       full
--- a/yarn.lock
+++ b/yarn.lock
@@ -2,36 +2,47 @@
 # yarn lockfile v1
 
 
-"@ampproject/toolbox-core@^1.1.1":
-  version "1.1.1"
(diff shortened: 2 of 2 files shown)


ShimiSun:
I think it looks OK, my only issue is while scrolling up/down the white color pops, is there a way to handle it better?

halocline:
> I think it looks OK, my only issue is while scrolling up/down the white color pops, is there a way to handle it better?

Are you commenting on white above header / below footer? This happens in dark mode as well, correct? If I am seeing what you are, I think all sites experience this: https://www.apple.com/

halocline:
> > I think it looks OK, my only issue is while scrolling up/down the white color pops, is there a way to handle it better?
> 
> Are you commenting on white above header / below footer? This happens in dark mode as well, correct? If I am seeing what you are, I think all sites experience this: https://www.apple.com/

I guess we would have to set the background on the <body> of the doc instead, if that's what we are going for.

taysea:
> > > I think it looks OK, my only issue is while scrolling up/down the white color pops, is there a way to handle it better?
> > 
> > 
> > Are you commenting on white above header / below footer? This happens in dark mode as well, correct? If I am seeing what you are, I think all sites experience this: https://www.apple.com/
> 
> I guess we would have to set the background on the of the doc instead, if that's what we are going for.

I'm going to merge for now since the white background seems to be a standard behavior for scroll. If we want to adjust later for this, we can.

PR #391, opened 2020-02-20
Fix errors

britt6612:
There was linting checks that were not previously done in the deep-linking branch. This will fix the issues that were not caught before

### DIFF
--- a/aries-site/src/tests/navigation/deeplinks.js
+++ b/aries-site/src/tests/navigation/deeplinks.js
@@ -1,8 +1,8 @@
-
+/* eslint-disable max-len */
 /* eslint-disable no-undef */
-import { ReactSelector, waitForReact } from 'testcafe-react-selectors';
+import {  waitForReact } from 'testcafe-react-selectors';
 import { Selector } from 'testcafe';
-import { baseUrl, Search, getSuggestion } from '../utils';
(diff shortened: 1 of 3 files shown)


aries-site/src/tests/navigation/deeplinks.js
```diff
@@ -1,8 +1,8 @@
-
+/* eslint-disable max-len */
```
ShimiSun: I don't think we should disable it, just wrap it on the next line. easy fix.

britt6612: I can, thats not a problem but if thats the case then we should change that in all the testing files because alot of our descriptions were past 80characters so pretty much every file has to  
`/* eslint-disable max-len */`
to stay consistent I would change that in all of them

ShimiSun: I understand. we should make sure there are no new `eslint-disable max-len`  on top of the files and change the other files gradually, in the meanwhile we should focus on fixing this file properly.

britt6612:
I did leave the `/* eslint-disable max-len */` in the axe.js file. There are some comments in there that will be taken out as soon as the color issues are fixed so at that point we can make sure we don't have that anymore

PR #392, opened 2020-02-20
Add "Templates" Section and examples

taysea:
First pass at layout from "Templates" pages the intention here is to:
1. Show screen on both desktop and mobile layouts
2. Allow users to grab code for this screen
3. Give insight into which components make up the screen

Utilize "generate" function from grommet/themes to shrink the baseSpacing for these examples -- however, hpe theme is not currently set up to also adjust text sizing based on this (grommet theme does). However, using a scaled down theme benefits demonstrating the desktop layout).

The two examples here provide two flavors of displaying the components that make up a screen.

Some things with this approach that cause issue is that there is an in between state before the screen breaks to "small" where the desktop layout doesn't have enough real estate.

To-do: Update code links, links to included components


halocline:
[Preview Link](https://deploy-preview-392--keen-mayer-a86c8b.netlify.com/)

aries-site/src/examples/templates/SidebarExample.js
```diff
@@ -0,0 +1,49 @@
...
+  Configure,
+  Terminal,
+  ShieldSecurity,
+  StatusUnknown,
+} from 'grommet-icons';
+
+export const SidebarExample = ({ direction, mobile, ...rest }) => (
+  <Box
+    direction={direction}
+    background="blue"
```
halocline: We should get clarification from @L0ZZI on the color here. The designs show the same #0E5265 (blue.dark) for the sidebar background regardless of light or dark context.

https://www.figma.com/file/JdmovKzBcjU7e4NoFCcxcE/hpe-common-cx-navigation?node-id=60%3A170

Screen Shot 2020-02-20 at 1 21 25 PM

aries-site/src/examples/templates/dashboard/DashboardExample.js
```diff
@@ -0,0 +1,64 @@
...
+import { ScreenContainer, SidebarExample } from '..';
+
+export const DashboardExample = ({ mobile }) => {
+  return (
+    <ScreenContainer mobile={mobile}>
+      <Box
+        direction="row"
+        fill
+        width={mobile && { max: 'large' }}
+        height={mobile && { max: 'xlarge' }}
```
ShimiSun: why are we using `mobile` vs the Responsive Context sizing? Aren't we ignoring the case of tablet?

aries-site/src/examples/templates/dashboard/DashboardExample.js
```diff
@@ -0,0 +1,64 @@
...
+          <PageHeader />
+          <DashboardTiles mobile={mobile} />
+        </Box>
+      </Box>
+      {mobile && (
+        <SidebarExample
+          mobile={mobile}
+          direction="row"
+          fill="horizontal"
+          style={{ position: 'absolute', bottom: 0, left: 0 }}
```
ShimiSun: @ericsoderberghp couldn't think of a grommet-ish way to workaround this style, ideas?

aries-site/src/examples/templates/service-list-screen/ServiceListScreenExample.js
```diff
@@ -0,0 +1,59 @@
...
+      </Box>
+      {mobile && (
+        <SidebarExample
+          direction="row"
+          fill="horizontal"
+          mobile={mobile}
+          style={{ position: 'absolute', bottom: 0, left: 0 }}
+        />
+      )}
+    </ScreenContainer>
```
ShimiSun: Feels like a lot of it can be reused from the previous I commented on, can we create a helper component that will be reused on these two files?

aries-site/src/layouts/content/Example.js
```diff
@@ -99,6 +106,7 @@ export const Example = ({ code, designer, docs, figma, ...rest }) => {
         direction="row"
         background="background-front"
         pad="large"
+        // style={{ transform: 'scale(0.7)' }}
```
ShimiSun: delete?

aries-site/src/pages/templates/index.js
```diff
@@ -0,0 +1,36 @@
...
+
+const title = 'Templates';
+const page = getPageDetails(title);
+
+const Templates = () => {
+  const descriptiveHeader = (
+    <DescriptiveHeader
+      background={page.color}
+      subText={page.description}
+      icon={page.icon}
```
ShimiSun: might be easier to deconstruct those values on line 8

aries-site/src/pages/templates/list-screen.js
```diff
@@ -0,0 +1,90 @@
...
+        </Example>
+        <SubsectionText>
+          Kickstarter austin heirloom whatever meggings artisan, hoodie vinyl
+          listicle. Drinking vinegar lumbersexual hell of umami palo santo YOLO.
+          Meh neutra deep v edison bulb portland jianbing. Aesthetic taxidermy
+          lyft, YOLO glossier enamel pin fam 3 wolf moon. Chartreuse subway tile
+          craft beer swag brooklyn PBR&B 8-bit live-edge edison bulb artisan.
+        </SubsectionText>
+      </Subsection>
+      <Subsection name="What components make this screen?" level={3}>
```
ShimiSun: I see a level=1 and level=3, why are we skipping level 2?

taysea: level 2 is default

aries-site/src/examples/templates/dashboard/DashboardExample.js
ericsoderberghp: We could try having the mobile screen be:

taysea: Good suggestions, thanks!

aries-site/src/examples/templates/SidebarExample.js
```diff
@@ -0,0 +1,61 @@
...
+  Apps,
+  Chat,
+  Clock,
+  Configure,
+  Terminal,
+  ShieldSecurity,
+  StatusUnknown,
+} from 'grommet-icons';
+
+export const SidebarExample = ({ direction, ...rest }) => {
```
halocline: Thoughts on naming this `GlobalNavBar` vs. `Sidebar` to align terminology with: https://github.com/hpe-design/design-system/issues/349?

`GlobalNavBar` may be more explicit so that when we see it, we know that it refers to the blue bar for navigation between HPE apps/services and is sometimes a sidebar, but also a bar at the footer on small screens.

aries-site/src/examples/templates/dashboards/DashboardExample.js
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
```
halocline: So this will be the file which is displayed in the Prism code preview, correct?

If so, does this meet our criteria of "I should be able to copy and paste into my own code"? This certainly presents the cleanest and I can see the imports, but I just want to confirm we are agreement in going this direction.

This approach also gives us the reusability of `ScreenContainer`, `AppHeader`, and `PageHeader` which we will want to use in other Template examples.

aries-site/src/examples/templates/list-views/ListScreenExample.js
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
```
halocline: Same as: https://github.com/hpe-design/design-system/pull/392/files#r384040850

aries-site/src/layouts/main/Layout.js
```diff
@@ -46,6 +46,7 @@ export const Layout = ({
                 position: 'fixed',
                 bottom: size !== 'small' ? '48px' : '24px',
                 right: '24px',
+                zIndex: 10,
```
halocline: I believe this is the same zIndex as the default zIndex in `Layer` -- any conflicts when a Layer is used?

Will this also resolve: https://github.com/hpe-design/design-system/issues/360?

aries-site/src/pages/templates/login-screen.js
```diff
@@ -0,0 +1,24 @@
+import React from 'react';
```
halocline: This file is not being used anywhere, correct?

Will this this file be implemented when work for https://github.com/hpe-design/design-system/issues/412 is done? If so, we should reference it in that issue so that this is seen and doesn't get orphaned.

aries-site/src/layouts/main/Layout.js
taysea: I will double check this -- I added it in because when I created the positioning of sidebar for mobile, as I scrolled the page in mobile, they were on top of the themetoggle. If anything, we could give it a higher value to take precedence over Layer or others.

aries-site/src/pages/templates/login-screen.js
taysea: Right, this is not currently being used. I think it would be better to just remove the file for now. Then, it can be added back in when it will be displayed on the site.

aries-site/src/examples/templates/list-views/ListScreenExample.js
taysea: Answered above.

aries-site/src/examples/templates/SidebarExample.js
taysea: I am open to this naming -- I'm not seeing `Global` on that link however I think it does align with previous conversations of `GlobalLayout` which contains this NavBar and `AppLayout`. Thinking this naming convention (and potentially a component) would be handled by the work Shimi mentioned she'd want to explore in the development of those components. Not sure if we want to resolve naming now or let that development work happen and then we can adjust this example afterwards.

aries-site/src/examples/templates/dashboards/DashboardExample.js
halocline: Great idea. I had forgotten we had discussed that as a solution. Creating https://github.com/hpe-design/design-system/issues/419 to represent this work.

aries-site/src/examples/templates/ScreenContainer.js
```diff
@@ -0,0 +1,41 @@
...
+              pad={{ horizontal: 'medium', bottom: 'medium' }}
+              flex
+              {...rest}
+            />
+          </Box>
+          {size === 'small' && (
+            <SidebarExample
+              direction="row"
+              fill="horizontal"
+              mobile={mobile}
```
ShimiSun: Are we keeping the `mobile` calculation way? what about another screen sizing?

aries-site/src/examples/templates/ScreenContainer.js
```diff
@@ -0,0 +1,41 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Box, ResponsiveContext } from 'grommet';
+import { SidebarExample } from '.';
+
+export const ScreenContainer = ({ mobile, ...rest }) => (
+  <ResponsiveContext.Provider value={mobile && 'small'}>
+    <ResponsiveContext.Consumer>
+      {size => (
```
ShimiSun: Can we pull `size` from `useContext`?

aries-site/src/layouts/main/Layout.js
ShimiSun: I'm curious about this random number.

aries-site/src/examples/templates/ScreenContainer.js
taysea: Because we are establishing the ResponsiveContext.provider value in this file, we can't use the hook.

aries-site/src/examples/templates/ScreenContainer.js
taysea: Good catch, we don't need it anymore. I'll remove this. We are using `size` and `useContext` in Sidebar to achieve this.

aries-site/src/layouts/content/ExampleComponents.js
```diff
@@ -0,0 +1,35 @@
...
+import { SubsectionText } from '../../components';
+
+export const ExampleComponents = ({ components }) => (
+  <Subsection name="What components make this screen?" level={3}>
+    <SubsectionText>
+      To find more details about the usage of each of these components, check
+      out their documentation.
+    </SubsectionText>
+    <Box direction="row-responsive" gap="medium" wrap>
+      {components.map(component => (
```
halocline: Inconsistent formatting of buttons on mobile, multiple devices\

Screen Shot 2020-02-25 at 2 23 22 PM

taysea: Good catch -- this is because Tiles doesn't have an href because we don't have documentation currently for it anywhere. I can point it to grid for now? -- when buttons have an href the text becomes left justified.

taysea: We should have a section on the site for Tiles though, since it's an aries-core component that Shimi has already built out.

aries-site/src/layouts/main/Layout.js
taysea: The idea was to have it sit on top of page content, and without this defined, the Example was sometimes scrolling on top of the ThemeToggle button. The value `10` was chosen somewhat arbitrarily, but just to be something that would be higher than the zIndex of normal page elements. Open to changing the value to something else or hear about rationale for how to decide this value.

aries-site/src/examples/templates/dashboard/DashboardExample.js
taysea: I like this approach, but Layer needs to be adjusted so that you can choose what box the "position" should align width. Going to file as a feature request on Grommet.

Issue #416, opened 2020-02-24
Controls Page - Convert examples from <UsageExample /> to use <Example />


halocline:
For each example on the Controls page:

- Use <Example /> component
  - Ensure referenced demo code is clean
  - Add related resource links as appropriate:
    - Designer
    - Figma
    - Storybook

PR #421, opened 2020-02-25
Control page/examples

britt6612:
Related Issue: #416 

Work Included: 
* Implemented the controls page examples to use the usage examples
* Each example will now have the more to view the code and other props

Screen Shots 
Screen Shot 2020-02-25 at 12 28 11 PM

### DIFF
--- a/aries-site/src/examples/components/controls/MenuExample.js
+++ b/aries-site/src/examples/components/controls/MenuExample.js
@@ -2,8 +2,6 @@ import React from 'react';
 import { Box, Menu } from 'grommet';
 import { aries } from '../../../themes/aries';
 
-import { UsageExample } from '../../../layouts';
-
 export const MenuExample = () => {
   const items = [
     { label: 'Past 12 months' },
@@ -20,16 +18,8 @@ export const MenuExample = () => {
   const background = aries.global.colors.background.light;
 
   return (
-    <UsageExample
-      pad={{
-        horizontal: 'large',
-        vertical: 'large',
-        small: { horizontal: 'large', vertical: 'xlarge' },
-      }}
-    >
       <Box background={background} round="xsmall">
         <Menu label="Choose a timeframe" items={items} width="medium" />
       </Box>
-    </UsageExample>
   );
 };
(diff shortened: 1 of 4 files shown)


aries-site/src/pages/components/controls.js
```diff
@@ -62,6 +67,9 @@ const Controls = () => (
     </ContentSection>
     <ContentSection>
       <Subsection name="Button">
+      <SubsectionText>
```
taysea: This comment applies to all the examples on this page, but I think there had been discussion a month or so ago (before you joined on the project) where we decided the content order would go:
1. Section title
2. Example
3. SubsectionText

I realize that isn't updated in the designs but came out of discussion, and I forgot about that when we talked this morning. For all of these, can you keep the order as listed above?

halocline: @taysea This order has had additional discussion / rationale. Offline discussions with Greg, plus finding the Slack convo for reference. It starts here + next 5-6 comments: https://hpe.slack.com/archives/GPPLUK6LR/p1581632692012300

taysea: Gotcha! Thanks for pointing me to that. I missed that conversation. PR looks good to me then.

aries-site/src/pages/components/controls.js
```diff
@@ -70,29 +78,36 @@ const Controls = () => (
...
-        </SubsectionText>
       </Subsection>
     </ContentSection>
     <ContentSection>
       <Subsection name="Menu">
-        <MenuExample />
-        <SubsectionText>
+      <SubsectionText>
           Menu is used to filter or sort content on a page. It is similar to the
           select component. However, use the select component when the user must
```
halocline: Can we add an `<Anchor />` around "select component" and link it to the Select component deep link? Also capitalize Select in both instances as select is the name of the component and thus a proper noun which should be capitalized.

aries-site/src/examples/components/controls/MenuExample.js
```diff
@@ -13,23 +10,9 @@ export const MenuExample = () => {
...
   return (
-    <UsageExample
-      pad={{
-        horizontal: 'large',
-        vertical: 'large',
-        small: { horizontal: 'large', vertical: 'xlarge' },
-      }}
-    >
-      <Box background={background} round="xsmall">
+      <Box round="xsmall">
```
halocline: good catch

aries-site/src/pages/components/controls.js
britt6612: good thinking! I added that in

PR #423, opened 2020-02-26
Template controls v1

taysea:
Updates `Example` to place all controls at bottom right hand corner -- similar to a video player. When the Example is expanded, a Layer opens full screen to give more real estate to consume the desktop example in a more realistic dimension. Here, the user can also toggle between desktop and mobile. These controls are found at the bottom right hand corner of the Layer as well.

For the mobile icon, I am pulling it from flat icon because grommet-icons doesn't currently have a mobile phone. I assume we would design our own and replace this with that.

Some updates were made to List to make it format better on mobile and style more closely to example designs we have.


taysea:
Closing because feedback/work has become focused on the v2 template controls.

PR #428, opened 2020-02-26
Feedback updates v1

taysea:
One approach to the feedback options proposed here:
- https://designer.grommet.io/?id=design-system-feedback-eric-soderberg-hpe-com
- https://docs.google.com/document/d/1spSsKk3nkift8iwFG5i8QHE1V-6BY1q4YUoYRMW220w/edit

Having all of the options live inline on the page felt expansive but jumping to a completely new page felt unwanted if I was still interested in consuming the content on the page I was on.

The subscribe button has not been linked up yet, this will require some more digging into -- depending on timing this week this section could be changed to a simple 'mailto:' for now.

### DIFF
--- a/aries-site/src/tests/utils.js
+++ b/aries-site/src/tests/utils.js
@@ -4,7 +4,7 @@ import { ReactSelector } from 'testcafe-react-selectors';
 export const baseUrl = 'http://localhost:3030';
 
 // Selector name found using Chrome React dev tools on prod mode of website
-export const Search = 'D Box';
+export const Search = 'Search__StyledTextInput';
 
 export const getLocation = ClientFunction(() => document.location.href);
 
--- a/aries-site/src/components/content/SubmitFeedback.js
+++ b/aries-site/src/components/content/SubmitFeedback.js
@@ -1,23 +1,149 @@
 import React from 'react';
-import { Anchor } from 'aries-core';
+import {
+  Box,
+  Button,
+  Form,
+  FormField,
+  Grid,
+  Layer,
+  MaskedInput,
+  ResponsiveContext,
+  Text,
+} from 'grommet';
+import { ChatOption, MailOption, Close, Github } from 'grommet-icons';
 import { ContentSection, Subsection } from '../../layouts/content';
 import { SubsectionText } from '.';
 
 export const SubmitFeedback = () => {
-  return (
-    <ContentSection lastSection>
-      <Subsection name="Still have questions?" level={2}>
-        <SubsectionText>
-          Something missing or looking for more information? Edit or open an
-          issue on Github to help us make the HPE Design-System better.
-        </SubsectionText>
-        <Anchor
+  const size = React.useContext(ResponsiveContext);
+  const [open, setOpen] = React.useState();
+  const onOpen = () => setOpen(true);
+  const onClose = () => setOpen(undefined);
+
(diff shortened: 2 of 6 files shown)


ericsoderberghp:
One aspect I was considering was whether people might ask in a communication channel how to get connected, if there was a deep link to the Feedback page, we could just include that in a message. If it's not a deep link, we'd have to include a URL but then also say something like "scroll to the bottom and click the 'feedback' link at the bottom", which feels a bit heavy.

ericsoderberghp:
Nice Layer though, has a bit of a grommet-designer feel to it :)

taysea:
> One aspect I was considering was whether people might ask in a communication channel how to get connected, if there was a deep link to the Feedback page, we could just include that in a message. If it's not a deep link, we'd have to include a URL but then also say something like "scroll to the bottom and click the 'feedback' link at the bottom", which feels a bit heavy.

Very true, that's a good point. Technically the "feedback" section at the bottom of those pages is a deeplink, but it might feel arbitrary to get a deeplink to a random content page.

However, there is also the "feedback" link in the footer of the home page. Would it make sense in that case to link to a separate feedback page (leave the other pages as the Layer)? in which case, this feedback page could be used as a shareable link? Understanding we'd be having two feedback behaviors, and not sure if that is ideal or not.

taysea:
@ericsoderberghp Added approach of having shareable link, but maintained the layer style for content pages:
https://deploy-preview-428--keen-mayer-a86c8b.netlify.com/feedback

ericsoderberghp:
It looks like the /feedback path isn't tied in to the routing as the page seems to be reloading. Otherwise, I think it's looking good.

aries-site/src/tests/utils.js
```diff
@@ -4,7 +4,7 @@ import { ReactSelector } from 'testcafe-react-selectors';
 export const baseUrl = 'http://localhost:3030';
 
 // Selector name found using Chrome React dev tools on prod mode of website
-export const Search = 'D Box';
+export const Search = 'Search__StyledTextInput';
```
britt6612: was this causing issues for you?

britt6612:
LGTM spacing and resizing looks better! 👍

aries-site/src/tests/utils.js
taysea: Yes, when I inspected the DOM, search was no longer `D Box` but instead `B Box`, so that selector felt unstable. Changing to the specific text input to be more reliable in future changes.

taysea:
> Looks good. I like the spacing/layout updates.
> 
> A couple general comments:
> 
> * Logged bug for this: #431
Thank you!

> * Can you remind me what we determined about Layer not respecting the the ThemeMode? Do we have an issue for that?
We would need to add a piece to the theme (included in the expand screen pr) that adds a background color to layer in the theme -- by making it `background` it allows the layer to respect the thememode. If we want the feedback layer to respect thememode, it will automatically happen when the expand screen template merges. If we decide we want the white background always, we can set that explicitly on a Box inside the Layer.

> * Something for us to keep eyes on: Join Slack, then I am joining "Grommet" slack came off a little weird. May need to watch and/or adjust copy to "Join design-system channel on Grommet Slack"

Agreed on this disjoint, we can adjust the copy for now and then think more on it if needed.

Issue #418, opened 2020-02-25
hovering over footer anchors causes text to jump


ericsoderberghp:
### Is this issue related to the design system core library or design system website?

design-system-website

### Expected Behavior

Anchor text shouldn't jump left and right as I hover across them in the footer.

### Actual Behavior

jumpy text

PR #430, opened 2020-02-27
fix hover text

britt6612:
Related Issue #418 

## Work Included 

* Added a box around the NavLinks in order to set a width so that when the text changed to bold it 
would expand which was causing the text to move

* Moved footerLinks above to an array to clean up some code then mapped through that array in the NavLinks

* Added Margin for horizontal to footer otherwise since I added in the width to the box it was being expanded in the footer example

### Mobile
Screen Shot 2020-02-27 at 12 38 54 PM
 
### Full Screen

Screen Shot 2020-02-27 at 12 38 19 PM


britt6612:
@taysea thanks for catching that! After trying the different approaches I think this is something to be discussed with the team and designers about other approaches

ericsoderberghp:
I'm concerned that the use of bold text as a hover indicator for Anchor tends to cause issues like this and that we should discuss whether that's the best style to use in the theme. Agreed on wanting a discussion. :)

britt6612:
@ericsoderberghp I agree 100%. Especially if were trying to provide an out of the box solution for users to have. Setting a fixed width in a footer example is not something we would want. It was beneficial for me to try and explore different approaches for us to then go back to the design to discuss.

britt6612:
> I'm concerned that the use of bold text as a hover indicator for Anchor tends to cause issues like this and that we should discuss whether that's the best style to use in the theme. Agreed on wanting a discussion. :)

@L0ZZI @gregfuruiye

aries-site/src/layouts/main/Footer.js
```diff
@@ -1,12 +1,24 @@
...
-import { NavLink } from 'aries-core';
+import { FooterLink } from 'aries-core';
 import { Box, Footer as GrommetFooter, ResponsiveContext, Text } from 'grommet';
 import { nameToPath } from '../../utils';
 
 export const Footer = () => {
   const size = useContext(ResponsiveContext);
   const year = new Date().getFullYear();
+
+  const footerLinks = [
```
taysea: Should we rename this to be `externalFooterLinks` to be clear why `Feedback` isn't part of this?

aries-core/src/js/components/helpers/Anchors/FooterLink.js
```diff
@@ -0,0 +1,40 @@
+import React, { forwardRef, useState } from 'react';
```
taysea: This is maybe a bigger discussion for after the PR, but I think we need to assess all of the anchor treatments that currently exist in aries-core and make sure that we want the design system to support all of them and then potentially revisit earlier discussions about having them live under a single component called `Anchor` with prop `type`. Thoughts? @halocline

aries-site/src/layouts/main/Footer.js
britt6612: yes good suggestion Ill change that

aries-core/src/js/components/helpers/Anchors/FooterLink.js
halocline: @taysea Yes, I think the scope is larger than this PR. I added detail to this issue https://github.com/hpe-design/design-system/issues/308. Let's get on a call to talk through.

Issue #445, opened 2020-03-02
Update Header with code example, desktop/mobile controls 


MikeWalrath:
![image](https://user-images.githubusercontent.com/17712467/75711346-816b8580-5c83-11ea-9499-80a1e8ae1382.png)

Related to: https://github.com/hpe-design/design-system/issues/440

britt6612:
## Discuss if we want to add the desktop and mobile view to the header and footer examples in components page 

* Add component to be able to switch views from desktop to mobile
* Change the Header example to include hidden text when mobile view

halocline:
Needs discussion / resolution as to whether this work makes sense / fits / is consistent with the rest of the examples on this page.

halocline:
Closing due staleness.

PR #460, opened 2020-03-03
Layout page/examples

### LINKED ISSUE #440: Layout Page Convert examples to usage examples
For each example on the Layout page:

- Use <Example /> component
  - Ensure referenced demo code is clean
  - Add related resource links as appropriate:
    - Designer
    - Figma
    - Storybook

britt6612:
Related Issue: #440 

Closes #440 

Work Included: 
* Implemented the controls page examples to use the usage examples
* Each example will now have the more to view the code and other props

### DIFF
--- a/aries-site/src/examples/components/layouts/FooterExample.js
+++ b/aries-site/src/examples/components/layouts/FooterExample.js
@@ -8,11 +8,6 @@ export const FooterExample = () => {
     <Footer 
       background="background-front"
       direction={size !== 'small' ? 'row' : 'column-reverse'}
-      pad={{
-        vertical: size !== 'small' ? 'small' : 'large',
-        // Match horizontal padding of aries-core Nav
-        horizontal: size !== 'small' ? 'medium' : 'large',
-      }}
       align={size !== 'small' ? 'center' : undefined}
     >
       <Box
--- a/aries-site/src/examples/components/layouts/HeaderExample.js
+++ b/aries-site/src/examples/components/layouts/HeaderExample.js
@@ -8,7 +8,7 @@ export const HeaderExample = () => {
   const [focused, setFocused] = useState(false);
 
   return (
-      <Header pad="none" background='background-front'>
+      <Header fill="horizontal" pad="none" background='background-front'>
         <AppIdentity
           brand="hpe"
           logoOnly={size === 'small'}
--- a/aries-site/src/examples/components/layouts/HeaderExample.js
+++ b/aries-site/src/examples/components/layouts/HeaderExample.js
@@ -3,15 +3,15 @@ import { AppIdentity, Header } from 'aries-core';
 import { Box, ResponsiveContext, TextInput } from 'grommet';
 import { Search as SearchIcon } from 'grommet-icons';
 
-export const HeaderExample = ({ background, ...rest }) => {
+export const HeaderExample = () => {
   const size = useContext(ResponsiveContext);
   const [focused, setFocused] = useState(false);
 
   return (
-      <Header pad="none" background={background} {...rest}>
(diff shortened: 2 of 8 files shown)


aries-site/src/examples/components/layouts/LayerExample.js
```diff
@@ -1,15 +1,14 @@
...
-import { UsageExample } from '../../../layouts';
 
 export const LayerExample = () => {
   const [open, setOpen] = useState(false);
   const size = useContext(ResponsiveContext);
   const onOpen = () => setOpen(true);
   const onClose = () => setOpen(undefined);
   return (
-    <UsageExample>
+    <Box>
```
taysea: Could this outer box just be a <> and then the closing would be </> ?

halocline: Since this is going to be an example demonstrating best practices, I think it is a good idea to implement @taysea 's suggestion.

britt6612:
> Looking good. Two small changes for polish:
> 
> 1. [#460 (comment)](https://github.com/hpe-design/design-system/pull/460#discussion_r387275597)
> 2. The padding on the FooterExample is inconsistent (larger) with the rest of the examples on this page. Can we update this example for consistency?
> 
> Screen Shot 2020-03-03 at 1 55 53 PM

I agree the padding is off the initial code was to match the padding of the aries-core nav

```

      pad={{
        vertical: size !== 'small' ? 'small' : 'small',
        // Match horizontal padding of aries-core Nav
        horizontal: size !== 'small' ? 'medium' : 'large',
      }}

```

so not sure if there was a discussion as to why this had to match ?

britt6612:
Found another issue

Screen Shot 2020-03-03 at 2 32 35 PM

aries-site/src/examples/components/layouts/FooterExample.js
```diff
@@ -8,11 +8,6 @@ export const FooterExample = () => {
     <Footer 
       background="background-front"
       direction={size !== 'small' ? 'row' : 'column-reverse'}
-      pad={{
-        vertical: size !== 'small' ? 'small' : 'large',
-        // Match horizontal padding of aries-core Nav
-        horizontal: size !== 'small' ? 'medium' : 'large',
-      }}
       align={size !== 'small' ? 'center' : undefined}
```
taysea: Trying to comment something farther down this file, but the styling of the Footer Links doesn't match the most recent Footer styles (it's still using the bolded on hover).

aries-site/src/examples/components/layouts/HeaderExample.js
```diff
@@ -8,41 +8,40 @@ export const HeaderExample = () => {
...
-      >
-        {size !== 'small' || focused ? (
-          <TextInput placeholder="Search HPE Design System" plain />
-        ) : (
-          undefined
-        )}
+      <Header fill="horizontal" pad="none" background='background-front'>
+        <AppIdentity
+          brand="hpe"
+          logoOnly={size === 'small'}
```
taysea: I think we'd still want the HPE Design System text to appear unless the TextInput is focused, this can be in a separate PR though if you'd prefer.
Screen Shot 2020-03-06 at 10 18 50 AM

britt6612: yes I agree I think since Ill have to look at the logic of how that is hidden in the search component so I think this will be better to put in another PR I did add this to issue #445

aries-site/src/examples/components/layouts/FooterExample.js
britt6612: oh yes I changed that in the footer PR I just wasn't up to date with master should be good now!

PR #465, opened 2020-03-05
Form SettingsExample Pattern

halocline:
Related Issues: https://github.com/hpe-design/design-system/issues/464

[Preview](https://deploy-preview-465--keen-mayer-a86c8b.netlify.com/templates/forms#settings)

[Design](https://www.figma.com/file/K0PppsSh6aQSszqlwKAekg/hpe-design-system-stickers?node-id=267%3A77)

Work Included:
- Added `SettingsExample` to `Templates/Forms`
- Extended `aries` theme to support Toggle and RangeInput styling (changes here will eventually need to be ported over to `grommet-theme-hpe` and then pruned from `aries` theme)

Screen Shot 2020-03-05 at 9 54 42 AM

### DIFF
--- a/aries-site/src/examples/templates/forms/SettingsExample.js
+++ b/aries-site/src/examples/templates/forms/SettingsExample.js
@@ -1,0 +1,105 @@
+import React from 'react';
+import {
+  Box,
+  Button,
+  CheckBox,
+  Form,
+  FormField,
+  Header,
+  Heading,
+  Main,
+  MaskedInput,
+  RangeInput,
+  Text,
+  TextInput,
+} from 'grommet';
+
+import { FormContainer } from '.';
+
+export const SettingsExample = () => {
+  const [formValues, setFormValues] = React.useState({
+    doNotDisturbHours: '9:00pm - 5:00am',
+    notifications: true,
+    notificationsVolume: 0.7,
+  });
+
+  // eslint-disable-next-line no-unused-vars
+  const onSubmit = ({ value, touched }) => {
+    // Your submission logic here
+  };
+
+  return (
+    <FormContainer width="medium">
+      <Box gap="medium">
+        <Header
+          direction="column"
+          align="start"
+          gap="xxsmall"
+          pad={{ horizontal: 'xxsmall' }}
+        >
+          <Heading level={3} margin="none">
+            Settings
+          </Heading>
+          <Text>for HPE Service</Text>
+        </Header>
+        <Main
+          // Padding used to prevent focus from being cutoff
+          pad={{ horizontal: 'xxsmall' }}
+        >
+          <Form
+            value={formValues}
+            onChange={setFormValues}
+            onSubmit={({ value, touched }) => onSubmit({ value, touched })}
+          >
+            <Box margin={{ bottom: 'small' }}>
+              <Text margin={{ left: 'small' }}>Key feature</Text>
+              <FormField
+                name="notifications"
+                label="Notifications"
+                component={CheckBox}
+                reverse
+                help="Description of feature"
+                toggle
+              />
+              <FormField
+                htmlFor="notificationsVolume"
+                name="notificationsVolume"
+                label="Notifications Volume"
+              >
+                <RangeInput
+                  id="notificationsVolume"
+                  name="notificationsVolume"
+                  max={1}
+                  min={0}
+                  pad
+                  step={0.1}
+                />
+              </FormField>
+            </Box>
+            <Box margin={{ bottom: 'small' }}>
+              <Text margin={{ left: 'small' }}>Key feature</Text>
+              <FormField
+                name="doNotDisturb"
+                label="Do Not Disturb"
+                component={CheckBox}
+                help="Description of feature"
+                reverse
+                toggle
+              />
+              <FormField htmlFor="doNotDisturbHours" name="doNotDisturbHours">
+                <TextInput
+                  id="doNotDisturbHours"
+                  name="doNotDisturbHours"
+                  disabled={!formValues.doNotDisturb}
+                />
+              </FormField>
+            </Box>
+            <Box align="start" margin={{ top: 'medium', bottom: 'small' }}>
+              <Button label="Apply Settings" primary type="submit" />
+            </Box>
+          </Form>
+        </Main>
+      </Box>
+    </FormContainer>
+  );
+};
--- a/aries-site/src/examples/templates/forms/SettingsExample.js
+++ b/aries-site/src/examples/templates/forms/SettingsExample.js
@@ -7,8 +7,6 @@ import {
   FormField,
   Header,
   Heading,
-  Main,
-  MaskedInput,
   RangeInput,
   Text,
   TextInput,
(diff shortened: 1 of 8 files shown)


aries-site/src/examples/templates/forms/SettingsExample.js
```diff
@@ -0,0 +1,105 @@
...
+          align="start"
+          gap="xxsmall"
+          pad={{ horizontal: 'xxsmall' }}
+        >
+          <Heading level={3} margin="none">
+            Settings
+          </Heading>
+          <Text>for HPE Service</Text>
+        </Header>
+        <Main
```
taysea: Not sure we should be using `Main` here because there should only be one <main> per document which we are using for the page content.

https://www.w3schools.com/tags/tag_main.asp

aries-site/src/examples/templates/forms/SettingsExample.js
```diff
@@ -0,0 +1,105 @@
...
+            value={formValues}
+            onChange={setFormValues}
+            onSubmit={({ value, touched }) => onSubmit({ value, touched })}
+          >
+            <Box margin={{ bottom: 'small' }}>
+              <Text margin={{ left: 'small' }}>Key feature</Text>
+              <FormField
+                name="notifications"
+                label="Notifications"
+                component={CheckBox}
```
taysea: Noticed the "on" toggle shouldn't be filed in black from what I see in the sticker sheet. Should look the same as "off" toggle

Expected look:
Screen Shot 2020-03-05 at 9 03 04 AM

aries-site/src/examples/templates/forms/SettingsExample.js
```diff
@@ -0,0 +1,105 @@
...
+          pad={{ horizontal: 'xxsmall' }}
+        >
+          <Form
+            value={formValues}
+            onChange={setFormValues}
+            onSubmit={({ value, touched }) => onSubmit({ value, touched })}
+          >
+            <Box margin={{ bottom: 'small' }}>
+              <Text margin={{ left: 'small' }}>Key feature</Text>
+              <FormField
```
taysea: Do we want to adjust the left padding/margin (not sure which it is) of label?

Current:
Screen Shot 2020-03-05 at 9 03 33 AM

Design:
Screen Shot 2020-03-05 at 9 07 37 AM

aries-site/src/examples/templates/forms/SettingsExample.js
```diff
@@ -0,0 +1,105 @@
...
+                  max={1}
+                  min={0}
+                  pad
+                  step={0.1}
+                />
+              </FormField>
+            </Box>
+            <Box margin={{ bottom: 'small' }}>
+              <Text margin={{ left: 'small' }}>Key feature</Text>
+              <FormField
```
taysea: In general, what should focus behavior be for form fields with border? Right now focus is appearing on the outer box and the checkbox itself. This is probably on the grommet side and shouldn't inhibit this PR but I think this keyboard behavior should be discussed.

Screen Shot 2020-03-05 at 9 08 21 AM

aries-site/src/examples/templates/forms/SettingsExample.js
halocline: @taysea I could use a second set of eyes on applying that styling. It's a bit of a "rabbit hole" and ended up eating a lot of time.  Perhaps we can get on a call?

aries-site/src/examples/templates/forms/SettingsExample.js
halocline: Alignment of Labels is consistent throughout all the examples. This discrepancy between design and theme is an item to be discussed in: https://github.com/hpe-design/design-system/issues/458

aries-site/src/examples/templates/forms/SettingsExample.js
halocline: I will add this point to: https://github.com/hpe-design/design-system/issues/458

aries-site/src/examples/templates/forms/SettingsExample.js
```diff
@@ -0,0 +1,105 @@
...
+                reverse
+                help="Description of feature"
+                toggle
+              />
+              <FormField
+                htmlFor="notificationsVolume"
+                name="notificationsVolume"
+                label="Notifications Volume"
+              >
+                <RangeInput
```
taysea: When I hover over rangeinput knob, it creates a white border around the knob that's subtle. Is this the hover behavior we want? I'm fine with it but just couldn't tell from designs.
Screen Shot 2020-03-05 at 9 40 52 AM

aries-site/src/examples/templates/forms/SettingsExample.js
halocline: Good catch.

aries-site/src/examples/templates/forms/SettingsExample.js
taysea: Yeah, let's do it !

aries-site/src/examples/templates/forms/SettingsExample.js
halocline: Added to: https://github.com/hpe-design/design-system/issues/458

aries-site/src/examples/templates/forms/SettingsExample.js
halocline: Looking at this further, we can apply hover styling in the theme, but it gets [overwritten in Grommet's implementation](https://github.com/grommet/grommet/blob/3fe8dde589f8543a9d00cf4f9951b252f8d78398/src/js/components/RangeInput/StyledRangeInput.js#L78-L91).  The style from our theme is being applied, but then gets further CSS box-shadow applied, which trumps any previous box-shadow styling.

Possible solutions:
1. Add `!important` -- which I don't recommend
2. Modify Grommet
3. Leave as is and treat as discussion point

Thoughts?

ericsoderberghp: Probably #2 but worth #3, not #1

taysea: I would say we should leave as is for now and add as discussion point. I think the discussion should include the consideration to modify Grommet -- it seems if the user does something to the theme that they should see the effect of that.

I agree we should avoid `!important`

halocline: Adding to discussion points in: https://github.com/hpe-design/design-system/issues/458#issue-574790708

aries-site/src/examples/templates/forms/SignUpExample.js
```diff
@@ -41,7 +40,7 @@ export const SignUpExample = () => {
           </Heading>
           <Text>for a Hewlett Packard Enterprise account</Text>
         </Header>
-        <Main
```
britt6612: was there a particular reason we switched out `Main` with `Box` ?

britt6612:
Code wise it looks good to me Im still not able to see the figma file so I can't really compare to designs. Ill see if I can get that resolved then take another look

aries-site/src/examples/templates/forms/SignUpExample.js
halocline: Main should be reserved for the primary content container on a document. Per spec, a document should only have one Main.

halocline:
> Code wise it looks good to me Im still not able to see the figma file so I can't really compare to designs. Ill see if I can get that resolved then take another look

You'll need to have Chris add you as a viewer of the file.

PR #469, opened 2020-03-06
Form templates shipping

halocline:
Related Issues: https://github.com/hpe-design/design-system/issues/411

[Preview](https://deploy-preview-469--keen-mayer-a86c8b.netlify.com/templates/forms#shipping)
[Design](https://www.figma.com/file/K0PppsSh6aQSszqlwKAekg/hpe-design-system-stickers?node-id=251%3A189)

Screen Shot 2020-03-06 at 12 59 19 PM


aries-site/src/examples/templates/forms/ShippingExample.js
```diff
@@ -0,0 +1,126 @@
...
+            value={formValues}
+            onChange={setFormValues}
+            onSubmit={({ value, touched }) => onSubmit({ value, touched })}
+          >
+            <Box>
+              <Heading level={4} margin={{ bottom: 'small' }}>
+                Shipping Information
+              </Heading>
+              <Text>Shipping Address</Text>
+              <FormField htmlFor="firstName" name="firstName">
```
taysea: This is an issue on grommet side I think, but the formfield is not associating the label with the input field for any of these.

aries-site/src/examples/templates/forms/ShippingExample.js
```diff
@@ -0,0 +1,126 @@
...
+                  name="state"
+                  dropHeight="small"
+                  options={states}
+                  placeholder="Select State"
+                />
+              </FormField>
+              <FormField htmlFor="zipcode" name="zipcode">
+                <TextInput id="zipcode" name="zipcode" placeholder="Zipcode" />
+              </FormField>
+              <CheckBox name="isBusiness" label="This is a business" />
```
taysea: I'm noticing in the designs, a new checkbox design is being introduced that doesn't match the others. This is a comment of reconciling the theme (which design are we supporting, would we support both?)
Current implementation (matches what is in the theme now): 
Screen Shot 2020-03-06 at 12 12 12 PM

This design calls for:
Screen Shot 2020-03-06 at 12 12 36 PM

aries-site/src/examples/templates/forms/ShippingExample.js
halocline: Actually, none of these have labels specified 😬

aries-site/src/examples/templates/forms/ShippingExample.js
```diff
@@ -0,0 +1,126 @@
...
+          <Heading level={3} margin="none">
+            Shipping
+          </Heading>
+          <Text>for your HPE products</Text>
+        </Header>
+        <Main
+          // Padding used to prevent focus from being cutoff
+          pad={{ horizontal: 'xxsmall' }}
+        >
+          <Form
```
taysea: Noticing the formfields in the design have a black border as opposed to grey (I'm noticing a mix of grey and black formfield borders across the sticker sheet). For separate, we should reconcile which is the intended formfield border color.

aries-site/src/examples/templates/forms/ShippingExample.js
```diff
@@ -0,0 +1,126 @@
...
+          align="start"
+          gap="xxsmall"
+          pad={{ horizontal: 'xxsmall' }}
+        >
+          <Heading level={3} margin="none">
+            Shipping
+          </Heading>
+          <Text>for your HPE products</Text>
+        </Header>
+        <Main
```
taysea: Main needs to be removed here

aries-site/src/examples/templates/forms/ShippingExample.js
```diff
@@ -0,0 +1,126 @@
...
+        <Header
+          direction="column"
+          align="start"
+          gap="xxsmall"
+          pad={{ horizontal: 'xxsmall' }}
+        >
+          <Heading level={3} margin="none">
+            Shipping
+          </Heading>
+          <Text>for your HPE products</Text>
```
taysea: The gap between this text and the start of the form feels a bit too spacious in comparison to the design

aries-site/src/examples/templates/forms/ShippingExample.js
```diff
@@ -0,0 +1,126 @@
...
+          pad={{ horizontal: 'xxsmall' }}
+        >
+          <Form
+            validate="blur"
+            value={formValues}
+            onChange={setFormValues}
+            onSubmit={({ value, touched }) => onSubmit({ value, touched })}
+          >
+            <Box>
+              <Heading level={4} margin={{ bottom: 'small' }}>
```
taysea: This heading level isn't bolded in the design -- but semantically I think it makes sense to be a header.

aries-site/src/examples/templates/forms/ShippingExample.js
halocline: 3rd bullet here: https://github.com/hpe-design/design-system/issues/458

Also, the checkbox styling has changed over the last two days, this one got missed.

aries-site/src/examples/templates/forms/ShippingExample.js
halocline: I will fix.

taysea: Mmm then makes sense why no label is being found haha -- if labels aren't being displayed, we need to think about how accessibility can still be met. I can do some reading. I've read some things about having label still in the DOM but visually hidden.

Using placeholder text in place of a label doesn't meet accessibility on its own.

I don't know if this should stop the PR because it falls in line more with your other form discussion/question of how we design accessible forms.

aries-site/src/examples/templates/forms/ShippingExample.js
britt6612: I think since your using Heading its the margin thats already built in making that space the gap is the same so maybe give `margin top= 'none`

britt6612:
Screen Shot 2020-03-06 at 1 29 11 PM
Screen Shot 2020-03-06 at 1 29 16 PM

### Shipping address does not match with the margin as the other labels are. For example Phone number had that margin around the label again cant see designs but feels off to me.

britt6612:
code wise looks good besides a couple gap/margin clean up. still cant access designs so going to trust @taysea on her comparison until I get permissions

aries-site/src/examples/templates/forms/ShippingExample.js
halocline: I hadn't caught this one yet! Added `- FormField
  - Border color (light #CCCCCC vs. #EEEEEE)` to https://github.com/hpe-design/design-system/issues/458

aries-site/src/examples/templates/forms/ShippingExample.js
halocline: Good calls. Thank you.

halocline:
> Screen Shot 2020-03-06 at 1 29 11 PM
> Screen Shot 2020-03-06 at 1 29 16 PM
> 
> Shipping address does not match with the margin as the other labels are. For example Phone number had that margin around the label again cant see designs but feels off to me.

Good catch. Fixed.

aries-site/src/examples/templates/forms/ShippingExample.js
halocline: Added the following to: https://github.com/hpe-design/design-system/issues/458
- Form section headings
  - Design calls for 'normal' weight. Should we be treating these section headings semantically as headings? If so, theme has headings bold vs. normal.

aries-site/src/examples/templates/forms/ShippingExample.js
halocline: fixed

aries-site/src/examples/templates/forms/ShippingExample.js
halocline: fixed

aries-site/src/examples/templates/forms/ShippingExample.js
halocline: @taysea This [commit](https://github.com/hpe-design/design-system/pull/469/commits/cf52c7e3832507b1754a0aaec726c8c4796b3c86) proposes an approach for having field labels visually hidden, yet still available to screen readers. It implements this [suggestion from W3C](https://www.w3.org/WAI/tutorials/forms/labels/#note-on-hiding-elements) through a styled component.

aries-core/src/js/components/core/FormField/FormField.js
```diff
@@ -0,0 +1,31 @@
...
+      height: 1px;
+      margin: -1px;
+      overflow: hidden;
+      padding: 0;
+      position: absolute;
+      width: 1px;
+    }`};
+`;
+
+export const FormField = ({ visuallyHidden, ...rest }) => {
```
taysea: wondering if the prop should be called `labelHidden` or `hiddenLabel` since it's not the whole formfield becoming visually hidden, but just the label right?

Also, hesitant to pull this into the aries-core now that teams may be using the design system. I'm wondering if this prop would be valuable to have in grommet as well. If so, this wouldn't cause any import errors from teams (or us) having to swap out every import instance. Maybe it'd be good to discuss with @ericsoderberghp and see if a prop like this would have value in Grommet and if so we could implement there.

ericsoderberghp: Agreed on putting this in grommet.

What about: `label={{ text, hidden }}`, to make it clear it's about `label`?

taysea: I like that.

halocline: Agree. This was intended as more of a way to propose an approach. I can move it to be an `aries-site` component to not muddy `aries-core`.

halocline: > wondering if the prop should be called `labelHidden` or `hiddenLabel` since it's not the whole formfield becoming visually hidden, but just the label right?

good call.

halocline: 👍

PR #484, opened 2020-03-10
Update Search to use TextInput with Icon

### LINKED ISSUE #479: Update Search on design system site to use TextInput with icon

taysea:
This PR updates our Search component to utilize the `icon` prop that was recently added to Grommet's TextInput.

There is one remaining outstanding piece of this which I'll open an issue on, but is dependent on this grommet PR: https://github.com/grommet/grommet/pull/3830
- I will be adding a condition that closes the textinput when it's not in focus on mobile (so only the search icon appears), but I need the `onSuggestionsClose` prop to work to do that. Once the grommet PR merges, I'll update our Search. This has been an outstanding item on the Search component since we started, so it doesn't introduce any regression in functionality.

Closes #479

### DIFF
--- a/aries-site/src/layouts/navigation/Search.js
+++ b/aries-site/src/layouts/navigation/Search.js
@@ -1,10 +1,4 @@
-import React, {
-  useContext,
-  useRef,
-  useEffect,
-  useState,
-  useCallback,
-} from 'react';
+import React, { useContext, useRef, useEffect, useState } from 'react';
 import { useRouter } from 'next/router';
 import styled from 'styled-components';
 import PropTypes from 'prop-types';
@@ -28,14 +22,6 @@ export const Search = ({ focused, setFocused }) => {
   const [value, setValue] = useState('');
   const [suggestions, setSuggestions] = useState(allSuggestions);
   const inputRef = useRef();
-  // Needed so that boxRef.current is not undefined. Allows suggestions drop
-  // to match width of containing box as opposed to just width of text input
-  const [, updateState] = useState();
-  const forceUpdate = useCallback(() => updateState({}), []);
-
-  useEffect(() => {
-    forceUpdate();
-  }, [forceUpdate]);
 
   useEffect(() => {
     if (focused && inputRef.current) {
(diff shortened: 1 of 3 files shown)


aries-site/src/layouts/navigation/Search.js
```diff
@@ -27,8 +27,6 @@ export const Search = ({ focused, setFocused }) => {
   const size = useContext(ResponsiveContext);
   const [value, setValue] = useState('');
   const [suggestions, setSuggestions] = useState(allSuggestions);
-
-  const boxRef = useRef();
   const inputRef = useRef();
```
britt6612: Since you took out `const boxRef = useRef();` was this still needed ? maybe just change up comment

aries-site/src/tests/utils.js
```diff
@@ -30,7 +30,7 @@ export const getTabCount = ClientFunction(expectedPath => {
   let hrefIndex;
   for (let i = 0; i < tabbableElements.length; i += 1) {
     if (tabbableElements[i].getAttribute('href') === expectedPath) {
-      hrefIndex = i;
+      hrefIndex = i + 1;
```
britt6612: What was the reasoning behind having the `index + 1` ?

taysea: I believe previously when we had the tabbable box around the search input it was working, but logically off by one in terms of how a user would tab. The reason you add one is because array indexes start at 0. So, if the item I want to tab to is the first item in the array (index = 0), then I'd want to tab 1 time to get there (index + 1) -- maybe it's make more sense for the function to return hrefIndex + 1 instead of reassigning it there. I'll update and add a clarifying comment for why this is the case.

aries-site/src/layouts/navigation/Search.js
taysea: Good catch, I can actually remove a few functions there.

aries-site/src/tests/utils.js
britt6612: sounds good thank you! everything else looks good!

halocline:
It looks like we have lost our focus border color. Is that intentional?

Screen Shot 2020-03-10 at 11 17 38 AM

taysea:
> It looks like we have lost our focus border color. Is that intentional?
> 
> Screen Shot 2020-03-10 at 11 17 38 AM

That is due to the fact that we applied `plain` on the textinput to align with the designs (and now we don't have a bounding onClick box. I think this would live on the grommet side that `plain` should remove the border but still show focus indicator when using keyboard for navigation. What do you think about filing and issue on grommet separately for this?

PR #494, opened 2020-03-11
Form templates pay

### LINKED ISSUE #475: Form Template - Pay
Design - https://www.figma.com/file/K0PppsSh6aQSszqlwKAekg/hpe-design-system-stickers?node-id=256%3A0

britt6612:
Related Issues: #475 

closes #475 

[Preview](https://deploy-preview-494--keen-mayer-a86c8b.netlify.com/templates/forms#pay)
[Figma](https://www.figma.com/file/K0PppsSh6aQSszqlwKAekg/hpe-design-system-stickers?node-id=1%3A2)

Screen Shot 2020-03-11 at 10 22 37 AM


aries-site/src/examples/templates/forms/PayExample.js
```diff
@@ -0,0 +1,108 @@
...
+                placeholder="Jane Smith"
+              />
+            </FormField>
+            <FormField name="cardNumber" label="Credit Card Number">
+              <MaskedInput
+                id="cardNumber"
+                name="cardNumber"
+                reverse
+                mask={creditCardMask}
+                icon={<CreditCard color="status-disabled" />}
```
taysea: Should the color of this match the placeholder text color?

britt6612: right I wasn't sure about that since our discussion yesterday with how the theme may change the placeholder text. Ill look at the notes from yesterday

britt6612:
put a limit on the years for the `maskedInput` on the expiration. From 2020-2025 let me know if this should not be the case

aries-site/src/examples/templates/forms/PayExample.js
```diff
@@ -0,0 +1,108 @@
...
+              Credit Card Information
+            </Heading>
+            <FormField name="cardName" htmlFor="cardName" label="Name on Card">
+              <TextInput
+                id="cardName"
+                name="cardName"
+                placeholder="Jane Smith"
+              />
+            </FormField>
+            <FormField name="cardNumber" label="Credit Card Number">
```
halocline:
```suggestion
            <FormField htmlFor="cardNumber" name="cardNumber" label="Credit Card Number">
```

We need the `htmlFor` prop in order to properly tie the label to the input. It should be the same value in the input's id

aries-site/src/examples/templates/forms/PayExample.js
```diff
@@ -0,0 +1,108 @@
...
+                id="cardNumber"
+                name="cardNumber"
+                reverse
+                mask={creditCardMask}
+                icon={<CreditCard color="status-disabled" />}
+              />
+            </FormField>
+            <Box direction="row" gap="medium">
+              <Box flex={false}>
+                <FormField name="expiration" label="Expires on">
```
halocline:
```suggestion
                <FormField htmlFor="expiration" name="expiration" label="Expires on">
```

aries-site/src/examples/templates/forms/PayExample.js
```diff
@@ -0,0 +1,108 @@
...
+                  <MaskedInput
+                    id="expiration"
+                    name="expiration"
+                    mask={dateMask}
+                  />
+                </FormField>
+              </Box>
+              <Box fill>
+                <FormField name="cvv" label="CVV">
```
halocline:
```suggestion
                <FormField htmlFor="cvv" name="cvv" label="CVV">
```

aries-site/src/examples/templates/forms/formHelpers.js
```diff
@@ -15,6 +15,58 @@ export const emailMask = [
...
+    length: [1, 2],
+    options: Array.from({ length: 12 }, (v, k) => k + 1),
+    regexp: /^1[0,1-2]$|^0?[1-9]$|^0$/,
+    placeholder: 'MM',
+  },
+  { fixed: '/' },
+  {
+    length:  2,
+    options: Array.from({ length: 6 }, (v, k) => 20 + k),
+    regexp: /^2[0,1-5]$|^2?$/,
```
halocline: Saw your comment about restricting dates. I like the forward thinking and inspiration. In this case, I would recommend we think of the `mask` as specifying the "types" of values, but not restrict the actual values being entered. For the actual validation of the input, I would use a validation rule (though that might be overkill for this example case -- but could be fun to implement).

Recommendation:
- Remove the 20-25 restriction on the mask
- Play with a validation rule which is a function which compares the value and ensures the value is >= to today's date, but less than today's date plus years (in a real world implementation we'd likely lean on whatever payment service's API, but supplement it with some guardrails aiming to limit any erroneous data prior to the fetch call to the API)

aries-site/src/examples/templates/forms/formHelpers.js
```diff
@@ -15,6 +15,58 @@ export const emailMask = [
...
+    placeholder: 'YY',
+  },
+];
+
+export const cvvValidation = [
```
halocline: Is this "validation" or a "mask"? If mask, I'd suggest:

```suggestion
export const cvvMask = [
```

aries-site/src/examples/templates/forms/formHelpers.js
```diff
@@ -15,6 +15,58 @@ export const emailMask = [
...
+  {
+    length: 3,
+    regexp: /^[0-9]{1,3}$/,
+    placeholder: '123',
+  },
+];
+
+const creditCard =  /^[0-9]{1,4}$/;
```
halocline: I like the DRY principle here! Nice.

Recommend changing variable name to be more explicit - e.g. "creditCardRegExp" or similar

aries-site/src/examples/templates/forms/PayExample.js
halocline: There is a color named `placeholder` in the [theme](https://github.com/grommet/grommet-theme-hpe/blob/2f070ee90ebaf3cbe8d0fa876efe9476195ad06e/src/js/index.js#L105). If we use it, whenever the theme gets updated, the update will automatically flow through and remain consistent.



```suggestion
                icon={<CreditCard color="placeholder" />}
```

aries-site/src/examples/templates/forms/formHelpers.js
britt6612: nice catch!

britt6612:
I added some validation for the dates. I left the date mast but then added a way to get the date that was entered and reverse so that was it was yy/mm and had to then add the 20 to which ever was selected for the YY then put that full date 2021 fro example in the formate yyyy/mm to compare the current date

aries-site/src/examples/templates/forms/PayExample.js
```diff
@@ -0,0 +1,122 @@
...
+                    mask={cvvMask}
+                    id="cvv"
+                    name="cvv"
+                    placeholder="123"
+                  />
+                </FormField>
+              </Box>
+            </Box>
+            <Box align="start" margin={{ top: 'medium', bottom: 'small' }}>
+              <Button label="Check out" primary type="submit" />
```
taysea:
```suggestion
              <Button label="Checkout" primary type="submit" />
```

PR #495, opened 2020-03-11
Fixing orphaned select form field label

halocline:
Resolves WAVE accessibility warning where select in a form field is reported as an orphaned label

**Before**
Screen Shot 2020-03-11 at 10 39 49 AM

**After** 
Screen Shot 2020-03-11 at 10 39 28 AM

### DIFF
--- a/aries-site/src/examples/templates/forms/FilterExample.js
+++ b/aries-site/src/examples/templates/forms/FilterExample.js
@@ -6,7 +6,6 @@ import {
   FormField,
   Header,
   Heading,
-  Main,
   Select,
   Text,
   TextInput,
--- a/aries-site/src/examples/templates/forms/FilterExample.js
+++ b/aries-site/src/examples/templates/forms/FilterExample.js
@@ -77,7 +77,7 @@ export const FilterExample = () => {
               />
             </FormField>
             <FormField
-              htmlFor="locationType"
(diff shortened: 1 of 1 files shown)


aries-site/src/examples/templates/forms/FilterExample.js
```diff
@@ -77,7 +77,7 @@ export const FilterExample = () => {
               />
             </FormField>
             <FormField
-              htmlFor="locationType"
+              htmlFor="locationType__input"
```
taysea: Nice! How did you know to add the `__input`?

halocline: Inspecting the DOM:
Screen Shot 2020-03-11 at 12 02 01 PM

Feels a bit brittle though. If Grommet were ever to change its implementation of how select/drop work, I think this would be susceptible of breaking.

taysea: Gotcha -- yeah it would be interesting to see if needs to be better implemented on Grommet's side. I feel like users shouldn't need to inspect the DOM

taysea: I'm fine with this for now though since it's creating a solution with how Grommet is currently implemented.

PR #504, opened 2020-03-12
Update form label, help, info, error, and placeholder styles

### LINKED ISSUE #496: Theme Update - FormField Label Style
FormField Label should be:
- aligned with field input border
- font-size should be 'x-small' (e.g. 14px)
- font-color 'text-xweak' (check for contrast)

### LINKED ISSUE #497: Theme Update - FormField Help Text Style
FormField help text should be:
- font-color '#999999' (need to check contrast)
- font-size 'xsmall'

Notes: 
- Need to check whether the help text has margin added to the line-height. Text line-height should be enough.
- May need to revisit this one
  -  This color is for the light theme. A dark equivalent has not been specified.
  -  #999999 is not a "named" color currently existing in the light theme and a dark version. Would be nice to be able to use named colors.
    - Recommendation: Change FormField label color to be 'text-weak' and FormField help to be 'text-xweak' --- Related Issue: https://github.com/hpe-design/design-system/issues/496

taysea:
Preview: https://deploy-preview-504--keen-mayer-a86c8b.netlify.com/templates/forms

This PR updates formfield label, help, info, error, and placeholder styles.

One note: As is, FormField labels have a vertical margin of `xsmall`. I've added the same as the bottom margin to help text to create the same spacing between the text and the formfield. However, in instances where there is label + help text, the label still has the vertical margin which seems to push it a little too far away from the help text.
Just label:
 Screen Shot 2020-03-12 at 10 45 14 AM
Label + help text:
Screen Shot 2020-03-12 at 10 45 21 AM

- Ideally: We could just add top margin to the formfield component/child, then have no margin on labels or help text and let the line height create the spacing. -- however, as of now we can't

Options with current theme styling options:
- Leave as is
- Remove any margin on label and help text (would put the text essentially right on the formfield border but the spacing between the label and help text would be correct)
- Is there a way to check if there is label + help prop on formfield and if it's only label, put the bottom margin on the label?

Closes #497 
Closes #496

### DIFF
--- a/aries-site/src/themes/aries.js
+++ b/aries-site/src/themes/aries.js
@@ -3,6 +3,11 @@ import { deepMerge, normalizeColor } from 'grommet/utils';
 
 export const aries = deepMerge(hpe, {
   defaultMode: 'dark',
+  global: {
+    input: {
+      weight: 500,
+    },
+  },
   layer: {
     background: 'background',
   },
@@ -22,6 +27,37 @@ export const aries = deepMerge(hpe, {
           border: 2px solid ${theme.global.colors.text.light};
           top: 0px;
         `,
+      },
+    },
+  },
+  formField: {
+    label: {
+      size: 'xsmall',
+      color: 'text-weak',
+      margin: {
+        horizontal: 'none',
+      },
+    },
+    error: {
+      size: 'xsmall',
+      color: 'text-xweak',
+      margin: {
+        start: 'none',
+      },
+    },
+    help: {
+      size: 'xsmall',
+      color: 'text-xweak',
+      margin: {
+        start: 'none',
+        bottom: 'xsmall',
+      },
+    },
+    info: {
+      size: 'xsmall',
+      color: 'text-xweak',
+      margin: {
+        start: 'none',
       },
     },
   },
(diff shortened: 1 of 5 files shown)


aries-site/src/themes/aries.js
```diff
@@ -25,6 +30,37 @@ export const aries = deepMerge(hpe, {
       },
     },
   },
+  formField: {
```
ericsoderberghp: Shouldn't these all be in the grommet-theme-hpe?

taysea: Yes, but I think the thought was to implement it here as a staging place and allow the team to review, then at the end of each sprint (or some time frame) move anything that is placed here into the published theme.

If you'd prefer it just go straight to the HPE theme, that can work as well. Let us know your preference.

ericsoderberghp: That sounds fine as a sequencing approach. I'd just want the actual HPE theme updated as part of the card/sprint.

taysea: Great. I think Matt already created an issue for that, so I'll drag it into the sprint.

PR #528, opened 2020-03-12
Do Not Merge - Navigation Patterns for Templates

ShimiSun:
This PR is introducing Navigation patterns under Templates.
To view the changes, go to Templates -> Navigation on https://deploy-preview-528--keen-mayer-a86c8b.netlify.com/templates/navigation

Currently, I'm displaying four types of Navigation patterns, and looking for feedback:
* Navbar
* Sidebar
* Hub & Spoke (to include examples for List & Tiles)
* Navigation Cocktail (a mix of the above navigation patterns)

The current examples are just placeholders and so is the subsection text.
Looking for your feedback on the overall idea.


taysea:
I like the idea of showing different navigation patterns. However, I'm trying to think about how the examples might look. To me, it seems like these different navigational elements would often appear simultaneously within a screen (This would be your Navigation Cocktail section), but I'm wondering what the stand alone components would look like. I like the idea of showing each navigational element as an individual example as well. I'm wondering if those examples of just a Sidebar would live here or on the `Layout` Components page (similar to how Header and Footer are on the `Layout` Components page? I kind of like having all the Navigation pieces living together.

Some edits on the initial copy (I'm also no expert writer, but tried to rephrase a couple things):

A navigation pattern often includes several navigational pieces such as global, local, supplemental, and contextual navigation. All of these are vital elements under the broader topic of web navigation.

Establishing a clear hierarchy of navigation components creates a more efficient user experience and more scalable development experience. The **global navigation** of a website creates an outline for the site structure, allowing the user to easily maneuver from one area of a website to the next. On the other hand, **local navigation** helps users maneuver within a specific section of a website.

Creating clear and consistent navigation patterns allows for more efficient experiences with a website.

ericsoderberghp:
Good comments above.

I was thinking we tell these stories ...

1) navigation within an application

i) hub and spoke: tiles or list with a poker handful of choices -> more tiles (optional) -> eventually to some details. Each step down provides navigational context back up the information hierarchy. Each page has a primary focus with minimal contextual navigation chrome. A good example is: https://publisher.grommet.io/?id=HPEXS-lozzi-hpe-com

ii) persistent navigation:  Making any navigation choice still persists the navigation structure. Easy to click through the choices.

a) header navigation bar with at most a poker handful of choices. Often mixed with some other persistent elements, such as search, session, branding. For example: https://github.com

b) sidebar with at most a poker handful of choices. Often mixed with other elements, such as session. For example: server app

c) sidebar with a long set of choices. Typically used for reference documentation. For example, https://developer.mozilla.org/en-US/docs/Web/API/Element. 

2) mixed global and app navigation (cocktail?) Minimal sidebar for global context and navigation choices plus all #1 varieties of app navigation.

ShimiSun:
@halocline will be joining the effort to keep this PR moving 🙌 

For referencing a Sidebar component we can temporarily use [this Sidebar](https://github.com/hpe-design/design-system/pull/546) as a component and replace it with the Grommet Sidebar once it is released.

We will also like to provide a react-router example on a separate github repo to demonstrate how navigation would work with the react-router package.

halocline:
Navigation Detail Page:
- [ ] Should demonstrate navigation within an application (and explain when to use)
  - [ ] Should demonstrate and explain the following patterns:
    - [ ] Hub & Spoke
    - [ ] Persistent Navigation
      - [ ] Header App Bar
      - [ ] Sidebar App Bar
      - [ ] Sidebar Outline / Table of Contents
- [ ] Should demonstrate navigation within a global + app context (and explain when to use)
  - [ ] Should use Sidebar for global context w/ each variant of app context navigation
- [ ] Should provide explanation w/ reference for how routing would work with react-router

halocline:
@ShimiSun @taysea @ericsoderberghp @L0ZZI -- tagging you for feedback on the latest direction.

taysea:
Going to close this due to inactivity.

PR #538, opened 2020-03-17
Update placeholder color to text-xweak

### LINKED ISSUE #508: Theme Update - Placeholder Text Styling
Placeholder text should be:

- weight normal
- color 'text-xweak' (need to check contrast)

taysea:
Based on this issue: https://github.com/hpe-design/design-system/issues/508

Color contrast for placeholder text is not flagged for contrast -- what is important is that the context of the formfield design can be discerned via screenreader or other form context such as labels so the user doesn't have to rely solely on the placeholder text.

Closes #508

### DIFF
--- a/aries-site/src/themes/aries.js
+++ b/aries-site/src/themes/aries.js
@@ -11,6 +11,10 @@ export const aries = deepMerge(hpe, {
       weight: 500,
     },
   },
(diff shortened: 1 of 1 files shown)


britt6612:
LGTM!

PR #546, opened 2020-03-23
Added Sidebar component

ShimiSun:
#548


aries-core/src/js/components/core/Sidebar/Sidebar.js
```diff
@@ -0,0 +1,30 @@
...
+import { Box } from 'grommet';
+
+export const Sidebar = ({ children, header, ...rest }) => {
+  return (
+    <Box
+      pad="medium"
+      gap="large"
+      fill="vertical"
+      width="auto"
+      height="100%"
```
ericsoderberghp: You don't need this with `fill="vertical"`

aries-core/src/js/components/core/Sidebar/stories/Green.js
```diff
@@ -0,0 +1,67 @@
...
+  Projects,
+  StatusInfoSmall,
+} from 'grommet-icons';
+
+import { aries } from '../../../../../../../aries-site/src/themes/aries';
+import { Sidebar } from '../Sidebar';
+
+/* Avatar */
+const SidebarHeader = () => (
+  <Box
```
ericsoderberghp: Avatar!

aries-core/src/js/components/core/Sidebar/stories/Green.js
```diff
@@ -0,0 +1,67 @@
...
+    background="blue!"
+  >
+    <Text alignSelf="center" size="xlarge">
+      SY
+    </Text>
+  </Box>
+);
+
+const Navigation = () => (
+  <Box align="center" fill="vertical" justify="between">
```
ericsoderberghp: I don't think you want `fill="vertical"` here.

aries-core/src/js/components/core/Sidebar/Sidebar.js
```diff
@@ -0,0 +1,30 @@
...
+      {...rest}
+    >
+      {header}
+      {children}
```
ericsoderberghp: Since all of the Navigation examples have the same wrapper, I'm wondering if Sidebar should handle that aspect? I'm thinking that Sidebar could do this instead of just `{children}`: `<Box justify="between"><Box>{children[0 - next to last]}</Box>{children[last]}</Box>`. That would simplify the children in the examples to just `Nav` components containing `Button` and `Anchor`.

aries-core/src/js/components/core/Sidebar/Sidebar.js
ShimiSun: right, this one is a left over...

aries-core/src/js/components/core/Sidebar/Sidebar.js
ShimiSun: yea, we can let Sidebar do more work on its side.
Can we just use `<Box justify="between">{children}</Box>`? 
wrapping the children with additional internal Box may be redundant unless we want to wrap with a Nav component instead, and wrapping with a Nav felt like a too wide assumption (I had it like that at the beginning)

Note: I just verified that the Fragment renders well on the Sidebar `justify="between"` styling.

halocline:
> The blue background doesn't always fill the whole height for me:
> 
Same.

Also, noticed some funkiness with the "Green" story, but haven't had a chance to look into it yet.
Screen Shot 2020-03-23 at 4 59 09 PM

ShimiSun:
@halocline @taysea good catches - I'll take a look tonight/tomorrow.
I'm wondering if that something that should be part of the sidebar impl or on the story side.

ShimiSun:
I believe the issues are addressed. Thanks for the feedback.

aries-core/src/js/components/core/Sidebar/stories/Green.js
```diff
@@ -0,0 +1,65 @@
...
+import { Sidebar } from '../Sidebar';
+
+/* Avatar */
+const SidebarHeader = () => (
+  <Box
+    align="center"
+    round="full"
+    height="xxsmall"
+    width="xxsmall"
+    justify="center"
```
taysea: This is still getting a little squished on mine. I'd expect it to be a perfect circle:
Screen Shot 2020-03-26 at 10 33 06 AM

ShimiSun: My son in response: "squishy things are the best" 
...And sure, I'll take a look :)

aries-core/src/js/components/core/Sidebar/Sidebar.js
```diff
@@ -0,0 +1,31 @@
...
+import PropTypes from 'prop-types';
+
+import { Box } from 'grommet';
+
+export const Sidebar = ({ children, header, ...rest }) => {
+  return (
+    <Box
+      pad="small"
+      gap="large"
+      width="auto"
```
ericsoderberghp: Do we need `width="auto"`?

ShimiSun: We don't -- removed. 👍

aries-core/src/js/components/core/Sidebar/stories/Green.js
ShimiSun: I believe it's taking care of on my last commit. let me know if it is still funky on your end.

aries-core/src/js/components/core/Sidebar/Sidebar.js
```diff
@@ -0,0 +1,25 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import { Box } from 'grommet';
+
+export const Sidebar = ({ children, header, ...rest }) => {
+  return (
+    <Box pad="small" gap="large" height={{ min: '100%' }} {...rest}>
+      {header}
+      <Box justify="between" align="center" fill="vertical">
```
taysea: Is there a difference between using `fill="vertical"` and `flex="grow"` here? I am only asking because my thought would've been to use `flex="grow"`

ShimiSun: I think that `flex="grow"` means that the element can grow as long as it has available space, while we want the element to spread only vertically and take the full 100% height.

ShimiSun:
TAken care of on the Grommet side, officially released on grommet v2.13.0.

PR #565, opened 2020-03-24
Removing the floating navigation bar, aka Stan

### LINKED ISSUE #556: Detail Page - Remove sidebar nav from layout

ShimiSun:
This is the best PR I filed this month, removing this hideous component is fulfilling from both design & dev points.

Cheers for cleaner code 🥂 

Fixes #556


aries-site/src/layouts/main/Layout.js
```diff
@@ -12,13 +12,7 @@ const calcPad = size => {
...
 };
 
-export const Layout = ({
-  children,
-  descriptiveHeader,
-  title,
-  isLanding,
-  isNavPage,
-}) => {
+export const Layout = ({ children, descriptiveHeader, title, isLanding }) => {
```
taysea: I think we can get rid of isLanding as well -- now that everything is full width. Then in `pages/index.js` we can remove the `isLanding` prop from the Layout tag.

aries-site/src/layouts/main/Layout.js
```diff
@@ -12,13 +12,7 @@ const calcPad = size => {
   return val;
 };
 
-export const Layout = ({
-  children,
-  descriptiveHeader,
-  title,
-  isLanding,
-  isNavPage,
```
taysea: We can remove the `isNavPage` prop from NavPage of `Guidelines`, `Foundation`, `Templates`, `Components`, `Design`, `Develop`, and `Resources`

aries-site/src/layouts/main/Layout.js
```diff
@@ -96,13 +85,11 @@ Layout.propTypes = {
   children: PropTypes.oneOfType([PropTypes.element, PropTypes.array]),
   descriptiveHeader: PropTypes.element,
   isLanding: PropTypes.bool,
```
taysea: Remove this if we agree that we no longer need `isLanding` at all.

aries-site/src/layouts/main/Layout.js
```diff
@@ -96,13 +85,11 @@ Layout.propTypes = {
...
-  isNavPage: PropTypes.bool,
   title: PropTypes.string,
 };
 
 Layout.defaultProps = {
   descriptiveHeader: undefined,
   title: undefined,
   isLanding: false,
```
taysea: Same as above.

aries-site/src/layouts/main/Layout.js
ShimiSun: I wish, seems like isLanding is needed for Footer rendering condition.

aries-site/src/layouts/main/Layout.js
ShimiSun: Music to my ears!

aries-site/src/layouts/main/Layout.js
taysea: You're right, good call.

PR #570, opened 2020-03-25
Update Example layout

taysea:
Updates layout to match this design: https://www.figma.com/file/IvYbGVlucoUMvUgbYFSG7a/Design-System-Card-Site?node-id=42%3A1116

This affects all of our examples (not just Dashboard). The Grommet button will appear only if a designer link is provided, Figma button only if figma link is provided, and see fullscreen button only if it is a screen template.


aries-site/package.json
```diff
@@ -8,7 +8,7 @@
   "dependencies": {
     "aries-core": "*",
     "grommet": "https://github.com/grommet/grommet/tarball/stable",
-    "grommet-icons": "^4.4.0",
+    "grommet-icons": "https://github.com/grommet/grommet-icons/tarball/stable",
```
ShimiSun: 🙌

aries-site/src/components/content/CollapsibleSection.js
```diff
@@ -0,0 +1,43 @@
...
+    <>
+      <Box
+        background="background-contrast"
+        direction="row"
+        gap="small"
+        onClick={() => {
+          setOpen(!open);
+          if (onClick) onClick();
+        }}
+        margin={{ bottom: 'xxsmall' }}
```
halocline:
```suggestion
        border={open && { side: 'bottom', color: 'border' }}
```
Matches design a bit closer

aries-site/src/components/content/CollapsibleSection.js
```diff
@@ -0,0 +1,43 @@
...
+        pad="medium"
+        round={!open ? 'small' : { corner: 'top', size: 'small' }}
+      >
+        {!open ? <FormDown /> : <FormUp />}
+        <Text weight="bold">{label}</Text>
+      </Box>
+      <Collapsible open={open}>
+        <Box
+          background="background-contrast"
+          height={{ max: 'medium' }}
```
halocline: Wondering if we can just have the box size to its contents and not restrict the height because the overflow scroll behavior is a little awkward? Can always add it back in if content becomes unruly.

Soft opinion.

aries-site/src/components/content/CollapsibleSection.js
taysea: Nice, I added this into my PR.

aries-site/src/components/content/CollapsibleSection.js
taysea: Good point. I bumped up the max height to large to eliminate some of the awkwardness of things that were just a tiny bit longer than 'medium'. But I think some restriction is good otherwise code examples like Tabs get super long.

taysea:
> Good to go. Rockin'.

Just pushing one more commit to add a button for `Docs` as a stand in before we are able to discuss what would fall under the "Show Properties" collapsible in the designs. Chris has given approval of this temporary fix.

PR #572, opened 2020-03-25
Update Landing page intro

taysea:
Preview: https://deploy-preview-572--keen-mayer-a86c8b.netlify.com/

Updates the top section of the landing page. Currently the same svg is being used for dark and light mode because Chris has not provided a dark mode version yet.


aries-site/src/pages/index.js
```diff
@@ -57,10 +57,10 @@ const Index = () => {
       <Meta title={title} description={pageDetails.seoDescription} />
       <Box gap="large">
         <HomeTiles>
```
halocline: Was wondering if we should change this to be a `row-responsive Box` or a `Grid` with flexible rows? That would prevent this huge gap in single column mode: 
Screen Shot 2020-03-25 at 1 34 30 PM

britt6612: This problem is also in the current site Correct .?

halocline:
correct.

aries-site/src/pages/index.js
taysea: It is just due to the nature of the layout being created with Tiles that have consistent height, so the text content doesn't fill the height fully in this width -- Let me play with an idea I have and push a commit up

taysea: Think I fixed it!

halocline: Sweet! Thank you. I will preview as soon as the build completes.

britt6612:
Spacing looks much better!

PR #574, opened 2020-03-25
Changed Footer and ThemeToggle to accommodate new designs

ShimiSun:
#553

### DIFF
--- a/aries-site/src/components/content/ThemeModeToggle.js
+++ b/aries-site/src/components/content/ThemeModeToggle.js
@@ -17,7 +17,7 @@ export const ThemeModeToggle = () => {
       overflow="hidden"
       pad={{ vertical: 'small' }}
       responsive={false}
-      round="small"
+      round="4px"
       margin={{ right: 'medium' }}
       width="small"
       onClick={() => darkMode.toggle()}
--- a/aries-site/src/components/content/ThemeModeToggle.js
+++ b/aries-site/src/components/content/ThemeModeToggle.js
@@ -1,5 +1,5 @@
 import React from 'react';
-import { Box, Button } from 'grommet';
+import { Box } from 'grommet';
 import { IconMoon, IconSun } from '../icons';
 import { useDarkMode } from '../../utils';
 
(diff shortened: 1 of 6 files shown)


ShimiSun:
> Looks good to me. Only comment is did Chris want the color for dark mode to be `blue!` still or something else?

Since we don't have the `dark` mode designs yet, I kept it as is until further notice.

britt6612:
One thing I noticed was the `border-radius` is `4px `is designs and `12px` in preview. Unless these were changed then ignore my comment

ShimiSun:
> One thing I noticed was the `border-radius` is `4px `is designs and `12px` in preview. Unless these were changed then ignore my comment

Good catch! Fixed

aries-site/src/components/content/ThemeModeToggle.js
```diff
@@ -8,23 +8,22 @@ export const ThemeModeToggle = () => {
...
-          pad={{ vertical: 'small', horizontal: 'medium' }}
-          responsive={false}
-          round="xlarge"
-          width="small"
-        >
-          {icon}
-          {label}
-        </Box>
-      </Button>
+    <Box
```
halocline: Curious as to the rationale for eliminating the `Button`?

britt6612:
Screen Shot 2020-03-25 at 5 26 28 PM
Screen Shot 2020-03-25 at 5 26 22 PM

I did also notice the width and height were bigger then in designs. I changed to what was in the design so `width: 139px` and `height: 40px`  In the second screen shot it looks closer to designs but Im not sure if there was a reason for leaving this bigger

aries-site/src/components/content/ThemeModeToggle.js
ShimiSun: Good question, the focus was off due to the Button, it didn't adjust to the content of internal Box, so I moved the onClick inside the Box so the interactive element will be the Box and the focus will be accurate to the Box content, and then the Button didn't have any props and was just a redundant element in the DOM - so I removed the Button. 
We can take another approach of replacing the old Box rendering-approach with a Button, but I continued in the direction that was existing and usage of a single component. 
I'm willing to explore the latter.

ShimiSun:
> ...Im not sure if there was a reason for leaving this bigger

The reason is to have a constant t-shirt size value so the button size won't flip sizes between dark and light modes.

britt6612:
> > ...Im not sure if there was a reason for leaving this bigger
> 
> The reason is to have a constant t-shirt size value so the button size won't flip sizes between dark and light modes.

Makes sense! Everything else LGTM

aries-site/src/components/content/ThemeModeToggle.js
ShimiSun: I was intrigued by this question and refactored the Box to be a Button wrapped with a Box, let me know what you think.

ShimiSun:
@britt6612 thanks for the feedback, I made another commit to refactor Box to Button, and I think the spacing looks better now, at least on Desktop screen.

britt6612:
> @britt6612 thanks for the feedback, I made another commit to refactor Box to Button, and I think the spacing looks better now, at least on Desktop screen.

welcome I'll check it out with the new change!

britt6612:
@ShimiSun looks great with the refactor to `button`

PR #582, opened 2020-03-25
Update content layout

taysea:
Merges in master changes to allow full width, restricts subsection text to width="large" to maintain readability but allows other page elements to expand full width. Updates topic link on content pages to incorporate new icons (similar to content card).


taysea:
closing, opening on carte

PR #583, opened 2020-03-25
Update content page layout

taysea:
Merges in master changes to allow full width, restricts subsection text to width="large" to maintain readability but allows other page elements to expand full width. Updates topic link on content pages to incorporate new icons (similar to content card).


aries-site/src/components/content/ThemeModeToggle.js
```diff
@@ -8,23 +8,14 @@ export const ThemeModeToggle = () => {
...
-          round="xlarge"
-          width="small"
-        >
-          {icon}
-          {label}
-        </Box>
-      </Button>
+    <Box width={{ max: 'small' }}>
+      <Button
+        color={{ dark: 'blue!', light: 'white' }}
```
halocline:
```suggestion
        color="background-front"
```

taysea: I'm going to leave as is since it looks like Chris already updated how this Button appears in the Footer -- we can update in a separate PR

Issue #588, opened 2020-03-26
Create new Detail Pages as result of https://github.com/hpe-design/design-system/issues/587


halocline:
https://github.com/hpe-design/design-system/issues/587 will result in the need for new content pages to be created as the site hierarchy flattens

Pages needed to be created should be documented in this issue by assignee of https://github.com/hpe-design/design-system/issues/587

# Pages needed to be created:
## Foundation:
- [ ] /foundation/content
- [x] /foundation/branding --> /foundation/our-brand
- [x] /foundation/iconography --> /foundation/icons

## For components, all of the following need to be created under /components:
- [x]  'Accordion'
- [x]  'Anchor'
- [x]  'Box'
- [x]  'Button'
- [x]  'CheckBox'
- [x]  'Footer'
- [x]  'Grid'
- [x]  'Header'
- [x]  'Layer'
- [x]  'Main'
- [x]  'MaskedInput'
- [x]  'Menu'
- [x]  'RadioButtonGroup'
- [x]  'RangeInput'
- [x]  'Select'
- [x]  'Stack'
- [x]  'Tabs'
- [x]  'TextArea'
- [x] 'TextInput'

## Templates:
- [ ]  /templates/cards
- [x]  /templates/list-views --> /templates/lists

## Extend:
- [x] Extend NavPage
- Will these be external links?

## Pages that need to be removed:
- [x] /develop (anything here)
- [x] /resources (anything here)
- [x] /design (and anything here)
- [x] /components/controls
- [x] /components/layout
- [x] /components/input
- [x] /guidelines (and anything here)

PR #592, opened 2020-03-26
Update file structure and structure.js to match Carte designs

### LINKED ISSUE #587: structure.js - Update / re-map content hierarchy per Carte specs
The result of this work will be pages which need to be created as result of the flattened content hierarchy. The pages needed to be created should be added to https://github.com/hpe-design/design-system/issues/588 for documentation

taysea:
I have also updated #588 with the pages that need to be created.

Closes #587


aries-site/src/tests/accessibility/axe.js
```diff
@@ -52,17 +52,3 @@ test.before(async t => {
     await t.expect(violations.length === 0).ok(createReport(violations));
   }
 });
-
```
britt6612: Curious why you took this out ? Since resources is no longer a path should we just use another path and still test or ?

britt6612:
Left one comment everything else looks good also ran tests to double check and everything checks out! 👍

aries-site/src/tests/accessibility/axe.js
taysea: We don't have any page with that layout anymore since it was only appearing on NavPages with no nav items

britt6612: got it that was the coming soon page!

PR #595, opened 2020-03-27
Create Components pages

taysea:
Preview: https://deploy-preview-595--keen-mayer-a86c8b.netlify.com/components
Moves components off of "Controls" "Layout" etc. and places each one on its own page.


aries-site/src/layouts/content/Example.js
```diff
@@ -133,7 +133,7 @@ export const Example = ({
                 mobile,
               })}
           </Box>
-          {(designer || docs, figma || template) && (
```
britt6612: lol good catch on the comma

britt6612:
Before we weren't showing anything for the main just the explanation was there a direction for this?

taysea:
> Before we weren't showing anything for the main just the explanation was there a direction for this?

I just put it there because it fits the same context of what we have in the text, and it felt empty to have no visual. Happy to swap it out once we have a design from Chris.

PR #596, opened 2020-03-27
Create Extend index page

taysea:
Creates the index page for Extend. Regarding the subpages, we will need clarification if these will have their own pages or they will be external links to those sites.

### DIFF
--- a/aries-site/src/pages/extend/index.js
+++ b/aries-site/src/pages/extend/index.js
@@ -1,0 +1,56 @@
+import React from 'react';
+import Link from 'next/link';
+import { Box, Heading, Paragraph } from 'grommet';
+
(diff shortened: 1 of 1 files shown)


aries-site/src/pages/extend/index.js
```diff
@@ -0,0 +1,56 @@
...
+      <Meta
+        title={title}
+        description={pageDetails.seoDescription}
+        canonicalUrl="https://design-system.hpe.design/extend"
+      />
+      <Box gap="large">
+        <PageIntro
+          image={{
+            src: cardImage,
+            alt: 'Card deck of HPE Design System component cards',
```
halocline:
```suggestion
            alt: 'Card deck of HPE Design System extend cards',
```

PR #598, opened 2020-03-27
add textarea and textinput to their own page

britt6612:
Preview: https://deploy-preview-598--keen-mayer-a86c8b.netlify.com/

Issue #588

### DIFF
--- a/aries-site/src/pages/components/textarea.js
+++ b/aries-site/src/pages/components/textarea.js
@@ -1,0 +1,37 @@
+import React from 'react';
+import { Meta, SubsectionText } from '../../components';
+import { ContentSection, Example, Layout, Subsection } from '../../layouts';
+import { TextAreaExample } from '../../examples';
+import { getPageDetails } from '../../utils';
+
+const title = 'TextArea';
+const page = getPageDetails(title);
+const topic = 'Components';
+
+const TextArea = () => (
+  <Layout title={title}>
+    <Meta
+      title={title}
+      description={page.seoDescription}
+      canonicalUrl="https://design-system.hpe.design/components/textarea"
+    />
+    <ContentSection>
+      <Subsection name={title} level={1} topic={topic}>
+        <SubsectionText>
+        When you need to allow the user to provide longer forms of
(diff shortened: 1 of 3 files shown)


aries-site/src/pages/components/textinput.js
```diff
@@ -0,0 +1,38 @@
...
+        of data and content. Passwords and tags can also be used with
+        the TextInput component. Style can be variable, based upon
+        the use case and customer need that will elicit user
+        confidence in success.
+        </SubsectionText>
+        <Example
+          docs="https://v2.grommet.io/textinput?theme=hpe#props"
+          code="https://raw.githubusercontent.com/hpe-design/design-system/master/aries-site/src/examples/components/input/TextInputExample.js"
+        >
+          <TextInputExample />
```
taysea: Everything looks good via set-up -- I was looking at the TextInputExample code though, and it looks like it could use some cleanup -- we don't really need the abstraction of "LabelVerticalExample" -- do you think you could clean that up?

britt6612: Yes I also saw that example looked off to me but I can do that in here!

aries-site/src/pages/components/textarea.js
```diff
@@ -0,0 +1,37 @@
...
+  <Layout title={title}>
+    <Meta
+      title={title}
+      description={page.seoDescription}
+      canonicalUrl="https://design-system.hpe.design/components/textarea"
+    />
+    <ContentSection>
+      <Subsection name={title} level={1} topic={topic}>
+        <SubsectionText>
+        When you need to allow the user to provide longer forms of
```
halocline: This doesn't need to be incorporated into this PR, but I am propose that the first piece of text after the page header be `page.description` which should be a short, to the point description/definition. Any extended description should be moved to the "Show Details" portion of the page. Thoughts?

britt6612: I agree I saw `Anchor` using both the Subsection text with the details we should discuss this more next week

PR #599, opened 2020-03-27
Center examples in Example and allow Fullscreen on everything

### LINKED ISSUE #585: UsageExample - Center example in container and apply consistent sizing to container
Carte design calls for examples:
- Should be centered horizontally and vertically in UsageExample container
- UsageExample container should have a consistent, proportional height across examples
  - Consider min height as some larger examples may need greater real estate
  - If need greater real estate, do we scale to fit? Or do we use overflow?

taysea:
Preview: 
This sets a height of `576px` for the Examples (a height previously determined as optimal for aspect ratio of full-screen templates). However, in the case of non-template, we are using a min-height of 576px since some forms are longer than this and would be cut off otherwise.

Now, any example is given a fullscreen option. However, "Desktop, Mobile" toggle is still reserved for templates.

Design:
Screen Shot 2020-03-27 at 10 31 35 AM

Closes #585

### DIFF
--- a/aries-site/src/layouts/content/Example.js
+++ b/aries-site/src/layouts/content/Example.js
@@ -115,12 +115,14 @@ export const Example = ({
       <Box margin={{ vertical: 'small' }} gap="large">
         <Box>
           <Box
-            direction="row"
+            align={!template ? 'center' : undefined}
             background="background-front"
-            pad={template ? { horizontal: 'large', top: 'large' } : 'large'}
+            direction="row"
             // Height for template screen needs to be between medium and large
             // to maintain aspect ratio, so this is small + medium
-            height={template ? '576px' : undefined}
+            height={template ? '576px' : { min: '576px' }}
+            justify="center"
+            pad={template ? { horizontal: 'large', top: 'large' } : 'large'}
             round={
               designer || docs || figma || template
                 ? { corner: 'top', size: 'small' }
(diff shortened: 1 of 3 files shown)


aries-site/src/layouts/content/Example.js
```diff
@@ -115,12 +115,14 @@ export const Example = ({
...
           <Box
-            direction="row"
+            align={!template ? 'center' : undefined}
             background="background-front"
-            pad={template ? { horizontal: 'large', top: 'large' } : 'large'}
+            direction="row"
             // Height for template screen needs to be between medium and large
             // to maintain aspect ratio, so this is small + medium
-            height={template ? '576px' : undefined}
+            height={template ? '576px' : { min: '576px' }}
```
halocline: Rather than hard code the `576px`, can we do the calculation for small + medium? That way if the theme changes base spacing, definitions of medium, small, etc. this would automatically adjust as well.

```
import {defaultProps} from 'grommet'
const { size } = defaultProps.theme.global;
const aspectHeight = `${parseInt(size.medium, 10) +
  parseInt(size.small, 10)}px`;
```

```
height={template ? aspectHeight : { min: aspectHeight }}
```

taysea: Nice, yes let me adjust.

taysea: Done!

PR #613, opened 2020-03-31
Create placholder pages for Extend section

taysea:
Preview: https://deploy-preview-613--keen-mayer-a86c8b.netlify.com/
This creates placeholder pages so we have no dead end links on the site. It leverages the "ComingSoon" component we had made prior with a small adjustment so the feedback button was not redundant on the page.

### DIFF
--- a/aries-site/src/components/content/ComingSoon.js
+++ b/aries-site/src/components/content/ComingSoon.js
@@ -26,7 +26,7 @@ export const ComingSoon = () => {
           <Text size={textSize}>Carte Design System</Text>
         </Box>
         <SubsectionText>
-          The HPE Design System is the way Hewlett Packard Enterprise’s brand,
+          The HPE Carte Design System is the way Hewlett Packard Enterprise’s brand,
           technology, and its partners share a single language for application,
           web, and digitial experiences to answer your customers needs—Look
           behind the element!
--- a/aries-site/src/components/content/ComingSoon.js
+++ b/aries-site/src/components/content/ComingSoon.js
@@ -23,7 +23,7 @@ export const ComingSoon = () => {
           <Text size={textSize} weight="bold">
             HPE
           </Text>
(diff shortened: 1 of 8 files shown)


aries-site/src/components/content/ComingSoon.js
```diff
@@ -18,32 +18,22 @@ export const ComingSoon = () => {
           This page appears to be empty, but content will be coming soon!
         </Text>
       </ContentSection>
-      <ContentSection lastSection>
+      <ContentSection>
         <Box direction="row" gap="small">
           <Text size={textSize} weight="bold">
             HPE
           </Text>
           <Text size={textSize}>Design System</Text>
```
halocline: Do you think we should be referring to "Carte" now? I don't know how all in we are ready to be. (pun unintentional). If so,
```suggestion
          <Text size={textSize}>Carte Design System</Text>
```

aries-site/src/components/content/ComingSoon.js
```diff
@@ -18,32 +18,22 @@ export const ComingSoon = () => {
...
         </Box>
-        <Text>
+        <SubsectionText>
           The HPE Design System is the way Hewlett Packard Enterprise’s brand,
```
halocline:
```suggestion
          The HPE Carte Design System is the way Hewlett Packard Enterprise’s brand,
```

aries-site/src/components/content/ComingSoon.js
taysea: Let's do it to be consistent. Easy to remove if we change it later on.

britt6612:
Looks good. I did see some places in which. this should also go unless you left it out for a reason ?

Under Foundation 
There should be Content, Cards are both 404 page also Icons doesnt have anything so maybe putting this there 

Under Templates the Cards is also a 404 page

PR #615, opened 2020-03-31
Detail Page - Add related content section

halocline:
Related Issues: https://github.com/hpe-design/design-system/issues/555

[Preview](https://deploy-preview-615--keen-mayer-a86c8b.netlify.com/)

Screen Shot 2020-03-31 at 10 03 40 AM

Work Included:
- Added related content section to 3 example detail pages:
  - `button.js`
  - `dashboards.js`
  - `principles.js` (note: principles does not have a `relatedContent` property on its object in `structure.js` -- this is purposeful to demonstrate how the detail page renders if related content is `undefined` or is and empty `[]`
- Refactored `<CardGrid>` to reduce boiler plate, automatically incorporate Next Links, etc.
  - Applied refactor to home, foundation, components, templates, and extend
- Added relatedContent query function to `search.js`


aries-site/src/data/structure.js
```diff
@@ -146,6 +146,7 @@ export const structure = [
     seoDescription:
       'HPE Design System dashboard template for screens featuring content within tiles.',
     sections: [],
+    relatedContent: ['Cards', 'Lists', 'Grid'],
```
taysea: I wonder if we want to drive relatedContent via tags such as 'Inputs', 'Controls', 'Layout', etc. Then, we'd get the related content cards based on other pages with the same tags. This might be more efficient for cross linking between pages. We could even have it support both specific page names (in case we wanted to hand pick a page to show up in related) as well as the tags. Thoughts?

I'd say it could be an optimization to this set up, but I think for the first pass we could move forward like this.

aries-site/src/pages/components/button.js
```diff
@@ -30,6 +31,14 @@ const Button = () => (
           <ButtonExample />
         </Example>
       </Subsection>
+      {relatedContent.length > 0 ? (
+        <Subsection name="Related">
```
taysea: Can we make this a `RelatedContentSection` component if the text will be consistent across pages?

aries-site/src/components/cards/LinkedCard.js
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import Link from 'next/link';
+
+import { nameToPath } from '../../utils';
+import { ContentCard } from '.';
+
+export const LinkedCard = ({ topic }) => {
```
taysea: If CardGrid is enforcing the usage of a linked card, do we need this component? Maybe we could just do the wrapping of a ContentCard within CardGrid directly.

aries-site/src/components/cards/CardGrid.js
```diff
@@ -12,11 +14,16 @@ export const CardGrid = ({ children, ...rest }) => {
       fill
       {...rest}
     >
-      {children}
+      {cards &&
+        cards.map(topic => <LinkedCard key={topic.name} topic={topic} />)}
```
taysea: Can we just use ContentCard here and wrap the Link around it? Feels like `LinkedCard` is an over abstraction.

aries-site/src/pages/components/button.js
halocline: The H2 will likely be consistent, but I think we will need flexibility in the subsection text. I originally headed that direction, but couldn't make text that works for a component like button, also work for something more abstract like principles or topics in the Extend section.

aries-site/src/components/cards/LinkedCard.js
halocline: Good call. Will re-evaluate.

aries-site/src/components/cards/CardGrid.js
halocline: Good call. Will re-evaluate.

aries-site/src/data/structure.js
halocline: I'm open to approach like this. I was thinking flexibility to allow for use-case patterns emerge and then address as those patterns appear.

aries-site/src/components/cards/CardGrid.js
halocline: Implemented.

aries-site/src/components/cards/LinkedCard.js
halocline: Implemented.

aries-site/src/pages/foundation/index.js
```diff
@@ -36,18 +35,7 @@ const Foundation = () => {
             <Paragraph fill>{pageDetails.description}</Paragraph>
           </Box>
         </PageIntro>
-        <CardGrid>
```
britt6612: Good Clean up

aries-site/src/utils/search.js
```diff
@@ -70,3 +74,14 @@ export const getCards = cardCategory => {
         : page.parent.name === cardCategory,
     );
 };
+
+/*
```
britt6612: curious about how you came up with this?

britt6612:
LGTM just one comment/question

aries-site/src/utils/search.js
halocline: about the getRelatedContent function?

britt6612: yes

halocline: The idea is that: 
1. Each object in the structure array represents a page
2. Need a flexible place to store related content
- That page may have a variety of related content that might be a direct sibling, parent, child or distant. 
- For example, the Dashboards page's related content might consist of content related to alternate layouts, other page templates, components which compose a dashboard, may be even link to a page on charting or colors used in charting. 
- With that need for flexibility to fit a variety of cases, I propose we create a `relatedContent` property on the page object which is an array of page names (right now, page names are assumed to be unique and are being treated as a unique key -- if this content lived in a database or content management system, we'd define these relationships via a unique id -- eventually we need to transition these references to ids, rather than names).
3. We can't assume that content provided for a page is 100% complete. Therefore, we should plan for cases where that property is not present, is empty, or has 1-n related pages.

With those assumptions, the approach to the function says:
1. Find the first object in the `structure` array which has a name equal to the page in which I am interested
2. Try to get the relatedContent property for that object, if exists use it; if not, assign an empty array.
3. Loop through that array, finding each page contained in that array
4. Return the resulting array of page objects

Hopefully that explanation is sound and helpful. LMK.

britt6612: That helped alot thank you!

PR #618, opened 2020-03-31
Use Stack layout to allow intro images to span behind page content

### LINKED ISSUE #605: Allow heading section images to extend fully
Screen Shot 2020-03-27 at 3 28 31 PM

instead of:

Screen Shot 2020-03-27 at 3 29 22 PM

taysea:
Preview: https://deploy-preview-618--keen-mayer-a86c8b.netlify.com/

This is a stab at the Stacked Layout. I am hoping to get some feedback/new eyes on this to help clean it up.
Screen Shot 2020-03-31 at 10 21 04 AM
## What this PR does
It uses `Stack` to place the cards images behind the page layout. On desktop layout, the Tiles are primarily two-column, so the background image should only fill just beyond half-way. However, once the layout switches to single-column, we would want the background image to span full width.

## Some notes
1. Because Responsive Context is established inside the `Layout` file, we can't use Responsive context with the NavPages. For example, if we want to style margin differently when screen size is `small`, we can't access `size` directly. Because of this, in the backgroundImage prop of `Layout`, you'll see I've created a piece of the object called `size` which leverages ResponsiveContext downstream in `PageBackground`.
2. Because each image is very specific, we can't rely on t-shirt sizing for the styling and have to use px.
3. We reach a really awkward layout width where the Layout falls to single column before the screen width hits small, but at this point we'd really want the background image to be utilizing the `size === small` styling. I'm wondering if this warrants us creating an additional custom breakpoint for the site or adjusting the value for the small breakpoint to this window window where the tiles break to single column. The header is also struggling at this width.
Screen Shot 2020-03-31 at 9 31 05 AM
4. I had to disable an accessibility rule called "region" which says that all page content must be contained within a landmark (nav, main, footer, etc.). The page element being flagged is a Stack div created by grommet that the user does not have access to, so I couldn't add a `role` for accessibility even if I wanted to. This may need to be address on the grommet side.

Closes #605


aries-site/src/components/content/PageBackground.js
```diff
@@ -0,0 +1,51 @@
...
+          : backgroundImage.pad
+      }
+      responsive={false}
+      width={size !== 'small' ? '65%' : 'fill'}
+    >
+      <Image
+        src={backgroundImage.src[themeMode]}
+        alt={backgroundImage.alt}
+        fit={backgroundImage.fit}
+        fill={backgroundImage.fill}
```
ShimiSun: we can deconstruct these variables on line 9 and use them directly.

aries-site/src/components/content/PageBackground.js
```diff
@@ -0,0 +1,51 @@
...
+          ? backgroundImage[size].margin
+          : backgroundImage.margin
+      }
+      pad={
+        backgroundImage[size] && backgroundImage[size].pad
+          ? backgroundImage[size].pad
+          : backgroundImage.pad
+      }
+      responsive={false}
+      width={size !== 'small' ? '65%' : 'fill'}
```
ShimiSun: 65% sounds random, where is this value coming from?

aries-site/src/layouts/main/Layout.js
```diff
@@ -24,51 +29,59 @@ export const Layout = ({ children, descriptiveHeader, title, isLanding }) => {
...
-                  top: 'medium',
-                }}
-              >
-                {children}
-                {!isLanding && <FeedbackSection />}
-              </Box>
-            </Main>
-            <Footer />
-          </Box>
+          <Stack fill guidingChild={backgroundImage && 'last'}>
```
ShimiSun: can you explain `guidingChild={backgroundImage && 'last'}` ?  
if `backgroundImage` exists the guiding child is the last?

aries-site/src/pages/components/index.js
```diff
@@ -4,53 +4,55 @@ import { Box, Heading, Paragraph } from 'grommet';
...
+      src: {
+        dark: '/carte-components-dark.svg',
+        light: '/carte-components-light.svg',
+      },
+      alt: 'HPE Carte Design System',
+      fit: 'contain',
+      margin: { top: '-200px', left: '-50px' },
+      small: {
+        margin: { top: '-250px', left: '0' },
+      },
```
ShimiSun: I'm not sure about those random numbers. can you back those numbers up?

aries-site/src/tests/accessibility/axe.js
```diff
@@ -23,7 +23,12 @@ test('should check Home page', async t => {
     const axeContext = {
       exclude: [],
     };
-    const axeOptions = { rules: { 'color-contrast': { enabled: false } } };
+    const axeOptions = {
+      rules: {
+        'color-contrast': { enabled: false },
```
ShimiSun: should we add a comment to enable it once the colors are approved?

ShimiSun:
Attaching an image from the site that doesn't look right:
image

And this might be a question to @L0ZZI , but the asymmetric behavior is bothering me on the left side of the page, the way the image overflows outside of the App Header limits, feels wrong
image

aries-site/src/pages/components/index.js
taysea: Trying to align as closely with the placement of the images as in the designs from Chris. T-shirt sizing doesn't give enough specificity to control how the images overflow of the edge of the page.

aries-site/src/components/content/PageBackground.js
taysea: Same as below, trying to just align as closely to how the designs overlap the text.

aries-site/src/layouts/main/Layout.js
taysea: On the pages when we have a backgroundImage, the page content becomes the `last` child in the Stack and we want that to drive the layout.

For details pages, we don't have a background image. So, the only child of the Stack would be the Box containing the page content (now the first child), in which case we would want that to guide. Therefore, this condition checks for that. We need this order of children within the Stack because otherwise the backgroundImage sits on top of the text.

taysea: I can add a comment in the file.

aries-site/src/tests/accessibility/axe.js
taysea: Yes, I'll add one into the file.

taysea:
> Attaching an image from the site that doesn't look right:
> image
> 
> And this might be a question to @L0ZZI , but the asymmetric behavior is bothering me on the left side of the page, the way the image overflows outside of the App Header limits, feels wrong
> image

Good screen shots. Regarding your second point, that would be a good question for @L0ZZI. For your first screen shot, that's one I'm still trying to fix. I agree it looks wrong.

aries-site/src/components/content/PageBackground.js
taysea: Done.

aries-site/src/layouts/main/Layout.js
taysea: Added comment.

aries-site/src/tests/accessibility/axe.js
taysea: Done.

PR #620, opened 2020-03-31
New/footer

britt6612:
Preview: https://deploy-preview-620--keen-mayer-a86c8b.netlify.com/

Designs: https://www.figma.com/file/IvYbGVlucoUMvUgbYFSG7a/Design-System-Card-Site?node-id=3%3A0

## Update Footer

* I have updated the footer to match the carte designs
* The order of the footer links have now changed
* light/dark mode button has changed
* Using `NavButton` with hover styles and not `Anchor` underlines


aries-site/src/components/content/ThemeModeToggle.js
```diff
@@ -10,9 +10,10 @@ export const ThemeModeToggle = () => {
   return (
     <Box width={{ max: 'small' }}>
       <Button
-        color={{ dark: 'blue!', light: 'white' }}
+        color='background-back'
         label={label}
         icon={icon}
+        plain
```
taysea: Can we style this the same way as NavButtons instead of using plain? I assumed that on hover it would have a background to allow it to appear more "clickable".

aries-site/src/layouts/main/Footer.js
```diff
@@ -30,29 +36,36 @@ export const Footer = () => {
...
+          <Link
+            key={item.name}
+            href={item.href}
+            passHref
+          >
+            <NavButton
+              item={item.name}
+              active={router.pathname === (item.href)}
+              target="_blank"
+              href={item.href}
```
taysea: Since we are passing the Href above, I'm not sure that we need href on the NavButton anymore. Can you try locally and see if it still works properly without the href here?

aries-site/src/layouts/main/Footer.js
```diff
@@ -30,29 +36,36 @@ export const Footer = () => {
         horizontal: size !== 'small' ? 'xlarge' : 'large',
       }}
     >
-      <Box direction="row-responsive">
-        <ThemeModeToggle />
-      </Box>
       <Box direction="row-responsive" gap="medium" align="center">
```
taysea: I think we should decrease the gap to match what is in the Header because it feels like there is too much space now between buttons since they have their own pad also.

aries-site/src/layouts/main/Footer.js
```diff
@@ -30,29 +36,36 @@ export const Footer = () => {
...
         <Box>
-          <Text size="small">&copy; {year} Hewlett Packard Enterprise</Text>
+          <Text size="medium">&copy; {year} Hewlett Packard Enterprise</Text>
```
taysea: Since "medium" is default, we don't need the size prop here.

aries-site/src/components/content/ThemeModeToggle.js
britt6612: awe yes that was before I was using `NavButton`

aries-site/src/layouts/main/Footer.js
britt6612: lol I have no idea why this was here! my bad its fine just above

britt6612:
> I think the ThemeToggle is still missing hover behavior of NavButton, but other than that looks good.

Yeah I didnt wrap the toggle in the NavButton I can do that

aries-site/src/components/button/NavButton.js
```diff
@@ -0,0 +1,33 @@
...
+  return (
+      <Button
+        onMouseOver={() => setHover(true)}
+        onFocus={() => setHover(true)}
+        onMouseOut={() => setHover(false)}
+        onBlur={() => setHover(false)}
+        ref={ref}
+        {...rest}
+      >
+        <Box
```
taysea, line 16: On mobile, I don't think the box should be spanning the full width -- if we apply `align="start"` does that fix it? Or if it doesn't work here maybe doing that on the bounding box of the NavButtons?
Screen Shot 2020-04-02 at 10 17 42 AM

aries-site/src/components/content/ThemeModeToggle.js
```diff
@@ -1,21 +1,42 @@
...
 import { useDarkMode } from '../../utils';
 
-export const ThemeModeToggle = () => {
+export const ThemeModeToggle = ({ active }) => {
+  const [hover, setHover] = useState(false);
   const darkMode = useDarkMode();
   const label = darkMode.value ? 'Light Mode' : 'Dark Mode';
   const icon = darkMode.value ? <IconSun /> : <IconMoon />;
   return (
     <Box width={{ max: 'small' }}>
```
taysea, line 13: Same as above.

PR #626, opened 2020-04-01
DO NOT MERGE - Create dynamic routes for NavPages

taysea:
This PR utilizes NextJS dynamic routing for the NavPages. The layout of these pages was identical across the various topics, so having multiple files for them was adding more maintenance.

## What is new?
- Created a page called /[topic].js which serves as a template for any route at /... As a result, all of the index.js files within pages/componentes, etc. have been removed. For cases like /feedback which doesn't follow the same structure as the NavPages, as long as we create its own file of feedback.js then that takes precedence. So we don't need to worry about layout conflicts. 
- An intro on dynamic routing can be found here: https://nextjs.org/docs/routing/dynamic-routes
- Info on the caveats mentioned above can be found here: https://nextjs.org/docs/routing/dynamic-routes#caveats

To optimize performance, we still want to leverage static HTML generation at build time since all of our data is known. So we do the following:
- Utilizing getStaticPaths and getStaticProps to generate the pages. From a performance perspective, this still creates static HTML pages on build time (the same way that our other files are as well). However, it also creates a JSON file that holds each routes specific data. More on that can be found here: https://nextjs.org/docs/basic-features/pages#scenario-2-your-page-paths-depend-on-external-data

- Some small adjustments were made to the links to leverage the template but instruct the browser what the true URL of the page is. Resources can be found here for usage of `as` prop on next Link: https://nextjs.org/docs/api-reference/next/link


aries-site/src/pages/[topic].js
```diff
@@ -0,0 +1,61 @@
...
+
+import { CardGrid, Meta } from '../components';
+import { Layout, PageIntro } from '../layouts';
+import { getCards, getPageDetails, titleCase } from '../utils';
+
+const NavPage = ({ topic: topicProp }) => {
+  const topic = titleCase(topicProp);
+  const pageDetails = getPageDetails(topic);
+  const cards = getCards(topic);
+  React.useEffect(() => console.log(pageDetails), [pageDetails]);
```
Comment (halocline) [inline aries-site/src/pages/[topic].js:13]: Still need? Delete if not.

aries-site/src/pages/[topic].js
```diff
@@ -0,0 +1,61 @@
...
+            <Paragraph fill>{pageDetails.description}</Paragraph>
+          </Box>
+        </PageIntro>
+        <CardGrid cards={cards} />
+      </Box>
+    </Layout>
+  );
+};
+
+// This function gets called at build time
```
Comment (halocline) [inline aries-site/src/pages/[topic].js:38]: Worth including this url in the comment?
https://nextjs.org/docs/basic-features/pages#scenario-2-your-page-paths-depend-on-external-data

aries-site/src/pages/index.js
```diff
@@ -10,20 +10,7 @@ const pageDetails = getPageDetails(title);
...
-      style: { transform: 'scale(1.4)', transformOrigin: 'top left' },
-      small: {
-        margin: { left: '-75px', top: '-75px' },
-      },
-      useTiles: true,
-    }}
-    title={title}
-    isLanding
-  >
+  <Layout backgroundImage={pageDetails.backgroundImage} title={title} isLanding>
```
halocline, line 13: I am getting a console warning related to the backgroundImage when I load a page in light mode. It seems as if the staticProps are being initialized in one themeMode on the Server, but the Client receives a different Prop

 "index.js:1 Warning: Prop `src` did not match. Server: "/carte-cards-dark.svg" Client: "/carte-cards.svg""

Screen Shot 2020-04-08 at 2 54 43 PM

halocline: Do we need to modify the `getStaticProps` function in `[topic].js` to accommodate for these assets which are dependent on `themeMode`?

halocline: Disregard. Looks like this error is on Master and unrelated to the work here.

halocline:
> There's a warning related to the `backgroundImage` for the NavPage that would be good to get cleaned up prior to merging.

Disregard. Issue seems to be coming from elsewhere.

aries-site/src/layouts/content/Subsection.js
```diff
@@ -78,7 +78,7 @@ export const Subsection = ({
           <Header>
             <Box align="start" gap="small">
               {level === 1 && topic && (
-                <Link href={`/${topic.toLowerCase()}`} passHref>
+                <Link href="/[topic]" as={`/${topic.toLowerCase()}`} passHref>
```
ShimiSun, line 81: Not sure I understand how this change works, can you elaborate?

taysea: The href tells text JS that we are going to use a dynamic route where the template is [topic].js, but we supply `as` so the browser knows what the value of the path will be. Described here: https://nextjs.org/docs/api-reference/next/link#dynamic-routes

aries-site/src/pages/[topic].js
Comment (taysea) [reply aries-site/src/pages/[topic].js:38]: Done.

taysea:
not currently relevant, closing for now.

PR #629, opened 2020-04-01
Variants section added to button page

halocline:
Related Issues: https://github.com/orgs/hpe-design/projects/1#card-35491931

[Preview](https://deploy-preview-629--keen-mayer-a86c8b.netlify.com/components/button)

Work Included:
- Added Variants section to `button.js`, plus added button variant examples
- Simplified `<ContentSection>` as new designs have reduced the nuance. Recommend keeping as component for spacing and scanability reasons.
- Minor tweaks to `<Example>` to adjust style on smaller screens
- Allow for subsection heading text to be semantically correct but render at a different size

For Discussion:
- Variants seemed buried below Guidance and may get lost
- Button examples do not demonstrate variants such as:
  - disabled
  - icon
  - sizes
- Sizing of the `<Example>` container seems quite large and introduces a lot of scrolling, especially if there are additional variants we'd like to show (e.g. disabled, icon, sizing). I am going to create another branch with smaller height <Example> so we can contrast with this branch.


aries-site/src/examples/components/button/ButtonExample.js
```diff
@@ -0,0 +1,7 @@
+import React from 'react';
+import { Button } from 'grommet';
+import { Next } from 'grommet-icons';
+
+export const ButtonExample = () => {
+  return <Button label="Button" icon={<Next />} reverse onClick={() => {}} />;
```
taysea, line 6: It doesn't feel "default" if we have to use `reverse` on it. I know this is adhering to designs but I'm wondering if it's not the best example for default.

halocline:
agree. I flip-flopped on it. Will take it out.

aries-site/src/layouts/content/ContentSection.js
```diff
@@ -1,37 +1,16 @@
...
+  <Box
+    as="section"
+    align="start"
+    gap="medium"
+    pad={{
+      bottom: 'medium',
+      horizontal: 'xxsmall',
+      top: 'large',
+    }}
+    {...rest}
```
taysea, line 14: Love seeing the cleanup in this.

aries-site/src/layouts/content/Example.js
```diff
@@ -93,6 +94,7 @@ export const Example = ({
   ...rest
 }) => {
   const theme = React.useContext(ThemeContext);
+  const viewport = React.useContext(ResponsiveContext);
```
taysea, line 97: Is there a reason we are using `viewport` as opposed to `size` like we usually have been doing?

halocline:
size was already declared as a variable elsewhere

aries-site/src/layouts/content/Example.js
halocline: size was already declared as a variable elsewhere, line 104 I believe

aries-site/src/examples/components/button/ButtonExample.js
halocline: agree. I flip-flopped on it. Will take it out.

aries-site/src/layouts/content/Example.js
taysea: What about destructing that line to be: 
```
const {
    size: { medium, small },
  } = defaultProps.theme.global;
```
Then, we can reserve `size` for ResponsiveContext you're adding

taysea: or maybe better:
```
  const { medium, small } = defaultProps.theme.global.size;
```

halocline: 👍

aries-site/src/layouts/content/ExampleControls.js
```diff
@@ -8,6 +8,7 @@ const ControlButton = ({ children, ...rest }) => (
     {({ hover }) => (
       <Box
         direction="row"
+        align="center"
```
taysea, line 11: On mobile, I think I'd expect the "see full screen" button to be at the top of the right. What do you think?
Screen Shot 2020-04-01 at 3 45 34 PM

PR #631, opened 2020-04-02
revert grommet version

halocline:
Setting Grommet version 2.11.3 due to failed builds on Grommet/Stable and Grommet 2.12.0. This will get our builds back up and running.

Error in build stated that the 'document is not available' meaning with SSR there is no client/document.


ShimiSun:
Hi @halocline @taysea, This change won't work since the design system is now using the Avatar component which isn't available on 2.11.3. I want to hear why we are reverting.

PR #632, opened 2020-04-02
Button Page Examples & Smaller Example Height 

halocline:
[Preview](https://deploy-preview-632--keen-mayer-a86c8b.netlify.com/components/button)

This branch presents a slightly different approach to the Variants section of the details pages. This is intended to compare to the initial PR which stays true to the Figma design.

### DIFF
--- a/aries-site/src/examples/components/button/ButtonIconExample.js
+++ b/aries-site/src/examples/components/button/ButtonIconExample.js
@@ -1,16 +1,21 @@
 import React from 'react';
-import { Box, Button } from 'grommet';
+import { Box, Button, ResponsiveContext } from 'grommet';
 import { FormPrevious, FormNext } from 'grommet-icons';
 
 export const ButtonIconExample = () => {
+  const size = React.useContext(ResponsiveContext);
+
   return (
-    <Box direction="row" gap="large">
-      <Box align="center" gap="medium">
+    <Box
+      direction={size !== 'small' ? 'column' : 'row'}
+      gap={size !== 'small' ? 'small' : 'medium'}
+    >
+      <Box direction={size !== 'small' ? 'row' : 'column'} gap="medium">
         <Button label="Button" icon={<FormPrevious />} onClick={() => {}} />
         <Button label="Button" icon={<FormNext />} reverse onClick={() => {}} />
         <Button icon={<FormNext />} onClick={() => {}} />
       </Box>
-      <Box align="center" gap="medium">
+      <Box direction={size !== 'small' ? 'row' : 'column'} gap="medium">
         <Button
           label="Button"
           icon={<FormPrevious />}
--- a/aries-site/src/examples/components/button/ButtonIconExample.js
+++ b/aries-site/src/examples/components/button/ButtonIconExample.js
@@ -1,25 +1,30 @@
 import React from 'react';
 import { Box, Button } from 'grommet';
-import { Previous, Next } from 'grommet-icons';
+import { FormPrevious, FormNext } from 'grommet-icons';
 
 export const ButtonIconExample = () => {
   return (
     <Box direction="row" gap="large">
       <Box align="center" gap="medium">
-        <Button label="Button" icon={<Previous />} onClick={() => {}} />
-        <Button label="Button" icon={<Next />} reverse onClick={() => {}} />
-        <Button icon={<Next />} onClick={() => {}} />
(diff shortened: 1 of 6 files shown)


aries-site/src/pages/components/button.js
```diff
@@ -92,6 +103,99 @@ const Button = () => (
...
+        <SubsectionText>fdsfsaf</SubsectionText>
+        <Example
+          docs="https://v2.grommet.io/button?theme=hpe#props"
+          code="https://raw.githubusercontent.com/hpe-design/design-system/master/aries-site/src/examples/components/button/ButtonSizingExample.js"
+          height={{ min: 'small' }}
+        >
+          <ButtonStatesExample />
+        </Example>
+      </Subsection>
+      <Subsection name="Button Sizing" level={3}>
```
taysea, line 184:
```suggestion
      <Subsection name="Button Sizes" level={3}>
```

aries-site/src/examples/components/button/ButtonSizingExample.js
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+import { Box, Button } from 'grommet';
+
+export const ButtonSizingExample = () => {
+  return (
+    <Box direction="row" align="center" pad="large" gap="medium">
+      <Button label="Default" />
+      <Button size="small" label="Small" />
+      <Button size="medium" label="Medium" />
```
taysea, line 9: I'm wondering if there's a better way to demonstrate that "medium" is the default size. Maybe by ordering them as
`small` `default` `large` ? The display made me wonder "is default the same as medium ..? I think so but I can't tell"

aries-site/src/examples/components/button/ButtonIconExample.js
```diff
@@ -0,0 +1,36 @@
...
+
+export const ButtonIconExample = () => {
+  const size = React.useContext(ResponsiveContext);
+
+  return (
+    <Box
+      direction={size !== 'small' ? 'column' : 'row'}
+      gap={size !== 'small' ? 'small' : 'medium'}
+    >
+      <Box direction={size !== 'small' ? 'row' : 'column'} gap="medium">
```
taysea, line 13: The buttons seem like they aren't respecting the default height (they're all growing to match the icon only size) -- maybe we need an `align="center"` when they are in a row? Also, wondering if Chris intended for them all to be in a single row across. That was my impression.

aries-site/src/examples/components/button/ButtonStatesExample.js
```diff
@@ -0,0 +1,24 @@
...
+
+export const ButtonStatesExample = () => {
+  const size = React.useContext(ResponsiveContext);
+
+  return (
+    <Box
+      direction={size !== 'small' ? 'column' : 'row'}
+      gap={size !== 'small' ? 'small' : 'medium'}
+    >
+      <Box direction={size !== 'small' ? 'row' : 'column'} gap="medium">
```
taysea, line 12: Same as above re: single row.

aries-site/src/pages/components/button.js
```diff
@@ -146,10 +151,53 @@ const Button = () => (
...
+          docs="https://v2.grommet.io/button?theme=hpe#props"
+          code="https://raw.githubusercontent.com/hpe-design/design-system/master/aries-site/src/examples/components/button/ButtonSizingExample.js"
+          height={{ min: 'small' }}
+        >
+          <ButtonSizingExample />
+        </Example>
+      </Subsection>
     </ContentSection>
     <ContentSection>
       {relatedContent.length > 0 ? (
```
taysea, line 203: The Related Content isn't displaying the parent category

halocline: Hmmm... it's that way on Carte branch too. Must be regression outside of this PR. Will look into after we get these merged in.

aries-site/src/examples/components/button/ButtonIconExample.js
halocline: I'll fix the height. Recommend we get the main stuff out the door, then can tweak individual examples as necessary. Okay with that approach?

britt6612:
I just had one comment besides what @taysea pointed out!

This may be something we look at later but maybe looking at how these would be listed in mobile view. 
If `full-screen` could just go at the top of its own column or be on the bottom of the other column
Screen Shot 2020-04-03 at 2 55 57 PM

PR #634, opened 2020-04-02
Formfield error theme changes

taysea:
Changes to formField error state based on discussions from today's call.


ericsoderberghp:
LGTM :)

PR #642, opened 2020-04-06
Fix alignment of NavButton

taysea:
Notice that the `alignStart` was causing the NavButton to align incorrectly in the header. Resolved, then wrapped footer links with a box with align=start to maintain the correct footer alignment.

### DIFF
--- a/aries-site/src/layouts/main/Footer.js
+++ b/aries-site/src/layouts/main/Footer.js
@@ -1,12 +1,7 @@
 import React, { useContext } from 'react';
 import Link from 'next/link';
 import { useRouter } from 'next/router';
-import {
-  Box,
-  Footer as GrommetFooter,
-  ResponsiveContext,
-  Text,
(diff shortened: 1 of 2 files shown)


aries-site/src/layouts/main/Footer.js
```diff
@@ -40,28 +35,26 @@ export const Footer = () => {
...
-              target="_blank"
-              rel="noreferrer noopener"
-            />
-          </Link>
-        ))}
-        {/* Need to pass href because of:
+        <Box direction="row" gap="xsmall">
+          {externalFooterLinks.map(item => (
+            <Box align="start">
+              <Link key={item.name} href={item.href} passHref>
```
halocline, line 41: @taysea I think we need to move this `key` up to the parent Box. Now getting a console warning:
Screen Shot 2020-04-08 at 3 02 41 PM

PR #648, opened 2020-04-08
Do Not Merge - Tile previews

britt6612:
## Work Completed

* Added the images into the cards for both light and dark mode 
* Cleaned up old files that are no longer going to be used


aries-site/public/extend-preview-chomp.svg
```diff
@@ -0,0 +1,14 @@
+<svg width="336" height="192" viewBox="0 0 336 192" fill="none" xmlns="http://www.w3.org/2000/svg">
```
halocline, line 1: Thinking we should move all of these images into a subdirectory like `public/images/card-previews` or something along those lines. Having them all at the public root is a little overwhelming.

britt6612: lol sounds good to me it is a little much!

aries-site/package.json
```diff
@@ -10,6 +10,9 @@
     "grommet": "https://github.com/grommet/grommet/tarball/stable",
     "grommet-icons": "https://github.com/grommet/grommet-icons/tarball/stable",
     "grommet-theme-hpe": "https://github.com/grommet/grommet-theme-hpe/tarball/NEXT-stable",
+    "imagemin-svgo": "^7.1.0",
+    "next-compose-plugins": "^2.2.0",
+    "next-optimized-images": "^2.5.7",
```
ShimiSun, line 15: Can you explain at the PR description why you added those packages and how they help to achieve the PR goal?

britt6612:
## Issue on adding Images to cards on home page 

- After watching the. network traffic it was observed that web-pack is stalling. 
- Tested the network after removing the logic for light and dark images and network traffic is still stalling. 

Screen Shot 2020-04-20 at 11 12 00 AM

## Started to explore other libraries and issues that were related to optimizing imaging as well as looking at the webpack config in Nextjs. 

- https://github.com/cyrilwanner/next-optimized-images

Looked through how this library was optimizing images and after installing and setting up next config file I found that there was no change in the time that the images are being loaded. 

- All of the images are svgs (should not be taking up much space) 
- Also page will load about half way through then stall and take about a min-2 to load all of the images

ShimiSun:
@britt6612 what are the resolutions from using:
```
  "imagemin-svgo": "^7.1.0",
    "next-compose-plugins": "^2.2.0",
```
Did it have any impact?

aries-site/src/examples/cardPreviews/textinput 2.js
```diff
@@ -0,0 +1,6 @@
+import React from 'react';
+import { TextInput } from 'grommet';
+
+export const TextInputPreview = () => {
+  return <TextInput focusIndicator placeholder="Placeholder text" />;
```
ShimiSun, line 5: why do we need an external component that only performs one line without a logic? and why does the file called `textinput 2.js`?
why number two? isn't it a component and it should be using upper case?

ShimiSun: the same comment for all the files that use the number two.

ShimiSun:
Out of 86 file changes, I think that this PR needs a massive clean up so I'd be able to dive into the root cause of the problem.

britt6612:
https://github.com/hpe-design/design-system/pull/647/files link to image components PR

britt6612:
opened a new PR #679

PR #659, opened 2020-04-09
Fix key warning -- move footer link keys up

halocline:
Related: https://github.com/hpe-design/design-system/pull/642#discussion_r405813574

We were getting a console warning wanting a unique "key" prop for the Links in our Footer. This adds the key and fixes the warning.

Moving key up from Link to bounding Box.


britt6612:
Thanks for fixing this I saw this as well on the cardImage PR!

PR #669, opened 2020-04-17
Fix hpe icon size and update aruba color

taysea:
- The HPE brand icon was too big (Eric and I had changed some calculations a month or so ago to that HPE icon wouldn’t need extra sizing to match the default sizes of other icons, but I never stripped that out of the AppIdentity)
-  The aruba icon wasn’t using aruba orange either. Need to use color="plain" to allow built-in orange color to be used.

### DIFF
--- a/aries-site/src/examples/foundation/branding/HpeElementExample.js
+++ b/aries-site/src/examples/foundation/branding/HpeElementExample.js
@@ -7,7 +7,7 @@ export const HpeElementExample = () => {
 
   return (
     <Box direction="row" align="center" gap="medium">
-      <Hpe color="brand" size="large" />
+      <Hpe color="brand" />
       <Box direction="row" gap="xsmall">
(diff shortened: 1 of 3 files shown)


halocline:
@taysea Your changes look good. I added a small commit for the logos on the branding page. Can you review?

aries-site/src/examples/foundation/branding/HpeElementExample.js
```diff
@@ -7,7 +7,7 @@ export const HpeElementExample = () => {
 
   return (
     <Box direction="row" align="center" gap="medium">
-      <Hpe color="brand" size="66px" />
+      <Hpe color="brand" size="large" />
```
taysea, line 10: Think we don't want any size here, but want to use the default size
```suggestion
      <Hpe color="brand" />
```

PR #679, opened 2020-04-20
Add images card tiles

britt6612:
## Work Completed

* Added the images into the cards for both light and dark mode 
* Cleaned up old files that are no longer going to be used 

### Check out [preview link](https://deploy-preview-648--keen-mayer-a86c8b.netlify.app/) to view the issue 

- After watching the. network traffic it was observed that web-pack is stalling. 
- Tested the network after removing the logic for light and dark images and network traffic is still stalling. 

Screen Shot 2020-04-20 at 11 12 00 AM

## Started to explore other libraries and issues that were related to optimizing imaging as well as looking at the webpack config in Nextjs. 

- https://github.com/cyrilwanner/next-optimized-images

Looked through how this library was optimizing images and after installing and setting up next config file I found that there was no change in the time that the images are being loaded. 

- All of the images are svgs (should not be taking up much space) 
- Also page will load about half way through then stall and take about a min-2 to load all of the images


ShimiSun:
Can you attach the svg package that Chris sent you?

britt6612:
## Files from Chris for svgs 


[carte-card-previews 2.zip](https://github.com/hpe-design/design-system/files/4506497/carte-card-previews.2.zip)

ShimiSun:
I think we close this one, its taken care of #686

PR #682, opened 2020-04-21
Update designer links to master file

taysea:
Point all of the designer links to the master file that uses HPE email. This should be used for all future designer examples as well.


taysea:
> This looks good to go. Question though, do you have any insight as to why Designer doesn't seem to be honoring routing? When I follow a link it resolves the last component I looked at, not what is specified in the url.

Hmm, I'm not sure on that. Would be a good question for Eric.

PR #686, opened 2020-04-21
Researched and Fixed nextjs performance issue for image loading

ShimiSun:
On the Carte designs, we are loading images for each card, and overall +40 cards on the landing page of the application.

The Carte landing page is running more than 30 SVG, i.e. ~30M of resources, that are translated to images as the page initially loads.

Due to the massive resource load, the application is experiencing the following different malfunction:
1. Blank white screen as the user scroll.
2. The browser is blocked and sending a timeout alert.
3. Interactive elements are not responding to events.

I explored common related NEXTJS loading issues of svgs, and the structure of assets on the public vs static directories and came to the conclusion that we are currently structuring the assets correctly on our NEXTjs package.

Next steps, ways to resolve the problem:
1. Allow lazy loading (i.e Infinitscroll) - but I didn't feel like it is the right direction before understanding the root cause of the problem.
2. Compression of the svg images.

I went down the rabbit hole of compressing the images programmatically, to make sure the compression occurs automatically during build time. 
After researching the most popular NextJS compression packages out there, the NextJS users seem to have positive experiences with `next-optimized-images` package, which had a high amount of downloads and seems to be well maintained.
I applied the package to my branch, configured it on next.config.js with an additional plugin package, and programmatically let it run on the SSR project. 
This package allows the svg compressions during build time, and lowered our resource load but not in a significant manner and still didn't cut our requirements for smooth scrolling, loading, and interactive elements interactions.

So instead of trusting automated tools for compressions during build time, I switched to a more aggressive way of compressing the svgs [manually](https://vecta.io/nano), and even allowing losing the image quality for the sake of testing, to see if that will work.
And guess what?! it still didn't work and the page still got a timeout!

So I went ahead exploring the image sizes after mega-compressions and one of the svgs seemed to be the root cause of the problem having more than 20M size, and no matter how far it went on compression it still didn't resize as expected. From here, Chris helped with editing the svg to png file and taking down the size of the image significantly.

After this change, the behavior worked as expected (Yay)! So I decided to reverse engineer the optimization packages and see if we can get away without adding additional dependencies and having our code cleaner, and it all worked, we didn't need a manual or automated compressing, and we all lived happy ever after!

Part of the investigation, I've done clean-up on the public assets you can see on this PR. and it all works as expected.

Resources used in the research process:
https://nextjs.org/docs/advanced-features/automatic-static-optimization
https://spectrum.chat/next-js/general/pre-loading-images~b4dc9e19-6457-4ee0-95ed-cae8498b7d91
https://github.com/cyrilwanner/next-compose-plugins


britt6612:
Curious was it the HPE Images svg ?

aries-site/src/components/cards/ContentCard.js
```diff
@@ -37,6 +37,7 @@ export const ContentCard = forwardRef(({ topic, ...rest }, ref) => {
                 fit={preview.image.fit || 'cover'}
               />
             ) : (
+              preview &&
```
halocline, line 40: Is this check redundant to line 28?

aries-site/src/data/structure.js
```diff
@@ -403,7 +403,7 @@ export const structure = [
       'Need a place to host images in the cloud for use by your HPE services and apps?',
     preview: {
       image: {
```
britt6612, line 405: Screen Shot 2020-04-20 at 8 14 48 PM
 Nice catch! I saw this but wasnt sure if there was a way around it

aries-site/src/components/cards/ContentCard.js
ShimiSun: Good observation, thanks. 
Fixed.

PR #690, opened 2020-04-23
Update link and validate example

taysea:
- Updates text input link to use master file
- Updates TextArea validation example to match how TextInput/Select work

### DIFF
--- a/aries-site/src/pages/components/textinput.js
+++ b/aries-site/src/pages/components/textinput.js
@@ -28,7 +28,7 @@ const TextInput = () => (
       <Subsection name={title} level={1} topic={topic}>
         <SubsectionText>{page.description}</SubsectionText>
         <Example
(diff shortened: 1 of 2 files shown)


aries-site/src/pages/components/textinput.js
```diff
@@ -28,7 +28,7 @@ const TextInput = () => (
       <Subsection name={title} level={1} topic={topic}>
         <SubsectionText>{page.description}</SubsectionText>
         <Example
-          designer="https://designer.grommet.io/textinput?id=HPE-design-system-taylor-seamans-hpe-com&mode=edit"
```
britt6612, line 31: should we switch out links ? for the main designer

britt6612:
Looks good thanks for switching this out!

aries-site/src/pages/components/textinput.js
taysea: It should be updated

britt6612: okay good to know!

PR #718, opened 2020-04-30
remove extra boxes

britt6612:
Did a first pass through the components we have put up to remove any extra `boxes`

I found that In `TextInput` there was `box` with `width='small'` but without that there wasnt much change. I did leave the `box` around `search` of `width='medium'` otherwise the search took up majority of the space. 

There is `boxs` around `select` in which they have a `width='medium'` without it the `select` component does get noticeably smaller so if we leave them there thats probably fine or can take a couple out to show the actual size without `box`

### DIFF
--- a/aries-site/src/examples/components/menu/MenuExample.js
+++ b/aries-site/src/examples/components/menu/MenuExample.js
@@ -1,5 +1,5 @@
 import React from 'react';
-import { Box, Menu } from 'grommet';
+import { Menu } from 'grommet';
 
 export const MenuExample = () => {
   const items = [
@@ -9,8 +9,6 @@ export const MenuExample = () => {
   ];
 
   return (
-    <Box round="xsmall">
       <Menu label="Manage Account" items={items} width="medium" />
-    </Box>
   );
 };
--- a/aries-site/src/examples/components/search/SearchExample.js
+++ b/aries-site/src/examples/components/search/SearchExample.js
@@ -1,6 +1,6 @@
 import React from 'react';
 import styled from 'styled-components';
-import { TextInput, Box } from 'grommet';
+import { TextInput } from 'grommet';
 import { Search as SearchIcon } from 'grommet-icons';
 
 // Inputs should always be accompanied by labels for accessibility. An icon
@@ -16,14 +16,12 @@ export const SearchExample = () => {
   const [value, setValue] = React.useState();
 
   return (
-    <Box width="medium">
-      <StyledTextInput
-        icon={<SearchIcon id="search-icon" />}
-        placeholder="Search placeholder"
-        reverse
-        value={value}
-        onChange={event => setValue(event.target.value)}
-      />
-    </Box>
+    <StyledTextInput
+      icon={<SearchIcon id="search-icon" />}
+      placeholder="Search placeholder"
+      reverse
+      value={value}
+      onChange={event => setValue(event.target.value)}
+    />
   );
 };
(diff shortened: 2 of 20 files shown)


halocline:
@taysea @britt6612 -- FormFields scale to fit their bounding container. Placing a Box around a FormField helps maintain a consistent FormField width throughout a page.  What is our opinion on the use of Box in these cases? Clean, consistent UI presentation vs. clean code examples?

I am good either way, but just want us to have an understanding that these are the tradeoffs and know that we are making a conscientious decision either way. Thoughts?

halocline:
If we are touching/updating a file that has a FormField, we should also make sure that FormField is contained in a Form. I would like for these updates to include adding Form.

britt6612:
> If we are touching/updating a file that has a FormField, we should also make sure that FormField is contained in a Form. I would like for these updates to include adding Form.

Since this was a separate issue I was going to do this in another PR to keep things small and easy to merge unless you guys don't mind them on this PR

britt6612:
> @taysea @britt6612 -- FormFields scale to fit their bounding container. Placing a Box around a FormField helps maintain a consistent FormField width throughout a page. What is our opinion on the use of Box in these cases? Clean, consistent UI presentation vs. clean code examples?
> 
> I am good either way, but just want us to have an understanding that these are the tradeoffs and know that we are making a conscientious decision either way. Thoughts?

Maybe we can bring this up tomorrow in our stand up or ask in the channel. But I think having cleaner code is the way I would go but I did leave some `boxs` because I wanted to see your guys thoughts. I think that depending on the users placement of these components their widths will change so I think adding in maybe in comments or guidance that a `box` can be used to help size `select` for example.

halocline:
> > If we are touching/updating a file that has a FormField, we should also make sure that FormField is contained in a Form. I would like for these updates to include adding Form.
> 
> Since this was a separate issue I was going to do this in another PR to keep things small and easy to merge unless you guys don't mind them on this PR

I'd prefer to include them if they are minimal and easy to understand. Will be more efficient to fix and review once, rather than go through all files multiple times, fix multiple times, and review multiple times.

britt6612:
> > > If we are touching/updating a file that has a FormField, we should also make sure that FormField is contained in a Form. I would like for these updates to include adding Form.
> > 
> > 
> > Since this was a separate issue I was going to do this in another PR to keep things small and easy to merge unless you guys don't mind them on this PR
> 
> I'd prefer to include them if they are minimal and easy to understand. Will be more efficient to fix and review once, rather than go through all files multiple times, fix multiple times, and review multiple times.

sounds good I have them in my git stash so I can add them to this PR.

halocline:
> > @taysea @britt6612 -- FormFields scale to fit their bounding container. Placing a Box around a FormField helps maintain a consistent FormField width throughout a page. What is our opinion on the use of Box in these cases? Clean, consistent UI presentation vs. clean code examples?
> > I am good either way, but just want us to have an understanding that these are the tradeoffs and know that we are making a conscientious decision either way. Thoughts?
> 
> Maybe we can bring this up tomorrow in our stand up or ask in the channel. But I think having cleaner code is the way I would go but I did leave some `boxs` because I wanted to see your guys thoughts. I think that depending on the users placement of these components their widths will change so I think adding in maybe in comments or guidance that a `box` can be used to help size `select` for example.

Let's err on the clean-code side for now. Still can bring up in standup, but don't want to let waiting get in the way of your progress.

PR #719, opened 2020-04-30
update component for textinput on preview

britt6612:
The image on `textinput` preview was not as correct. Originally I wanted to force the focus but that would throw off the accessibility  so I decided to just go with the inline style to match up better with how the focus is in `formfield`

### DIFF
--- a/aries-site/src/examples/cardPreviews/textinput.js
+++ b/aries-site/src/examples/cardPreviews/textinput.js
@@ -4,14 +4,16 @@ import { TextInput, FormField } from 'grommet';
 export const TextInputPreview = () => {
   return (
     <FormField
-      style={{ borderRadius: '4px' }}
-      fill="horizontal"
-      border={{ color: 'focus', size: 'small' }}
+      style={{
+        boxShadow: '0 0 2px 2px #00E8CF',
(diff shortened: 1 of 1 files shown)


aries-site/src/examples/cardPreviews/textinput.js
```diff
@@ -4,14 +4,16 @@ import { TextInput, FormField } from 'grommet';
 export const TextInputPreview = () => {
   return (
     <FormField
-      style={{ borderRadius: '4px' }}
-      fill="horizontal"
-      border={{ color: 'focus', size: 'small' }}
+      style={{
+        boxShadow: '0 0 2px 2px #00E8CF',
+        borderRadius: '4px',
+      }}
```
taysea, line 10: Think we need htmlFor="focus-id" on the FormField

britt6612: good catch!

halocline: Plus wrap in Form?

britt6612: right! how did I fix all these and miss my own

PR #722, opened 2020-04-30
Allow Example to restrict width of child

taysea:
Preview: 

After cleaning up boxes from the examples, the Search component was expanding too wide, and this could be an issue for future examples as well. This adds a `width` prop to `Example` that allows us to restrict the width of the Example child without having to clutter the specific example code itself.


britt6612:
Thanks for doing this!

PR #724, opened 2020-04-30
Flip logic for width restriction

taysea:
Flips the logic -- setting the width by 100% by default was causing elements to stretch wider than desired. Instead, no width is set but for elements meant to span full width (Header, Footer, Dashboards, etc.) we can specify a width of 100%.


britt6612:
Nice and this fixes what we were looking at

Issue #649, opened 2020-04-08
Update patterns to not use internal components 


MikeWalrath:
We would like to update the pattern example code to work as easily as possible. To copy/paste the code and have it work.

britt6612:
Some of the code that is used in our Patterns is from `aries-core` since we have not published those.  we should get clarification here on approach for if this would be okay to show these or not. 

Also `ScreenContainer` code is used but it is more of a template for us to showcase what things look like in full screen and mobile screen. If someone were to copy and paste they would not fully understand the context behind this code. We can maybe wrap this within our example page and use as a prop that way its still getting wrapped in the `ScreenContainer` but this would not be needed.

PR #730, opened 2020-05-01
Example wcode

britt6612:
closes issue #649 

we were calling other files in our example. 
for now I brought them all to the one file for the user to read. Some of the examples are really long so I think we should brainstorm another approach or explore other options on how to show templates with all the code we use but in a clean way.


aries-site/src/examples/templates/dashboards/DashboardExample.js
```diff
@@ -1,6 +1,212 @@
-import React from 'react';
-import { DashboardTiles } from '.';
-import { AppHeaderExample, PageHeaderExample, ScreenContainer } from '..';
+import React, { useContext } from 'react';
+import { AppIdentity, Identifier, Tiles, Tile } from 'aries-core';
```
taysea, line 2: I think we are steering away from using any 'aries-core' components since the library hasn't been published yet. We might need to create components here call "AppIdentity", etc. -- but would be good to get clarification here on approach

britt6612: yeah I wasnt to sure about that as well. I think we need more clarification on what we should show or make the same type of component with grommet

aries-site/src/examples/templates/dashboards/DashboardExample.js
```diff
@@ -1,6 +1,212 @@
...
+            icon={<StatusUnknown color="text-xweak" />}
+          />
+        </Box>
+      )}
+    </Box>
+  );
+};
+
+const ScreenContainer = ({ mobile, ...rest }) => (
+  <ResponsiveContext.Provider value={mobile && 'small'}>
```
taysea, line 157: This code is really only relevant to the way that the docs site toggles between desktop and mobile view for template examples. For someone who copy and pastes this out to their code, it would be irrelevant and probably cause confusion.

It feels like we will need to address this in the future.

taysea: Or maybe, this ScreenContainer should live inside Example.js instead? and if a `template` prop is given to an Example, then we wrap the example in this? I would have to look more into implementation but could be an option.

aries-site/src/examples/templates/forms/CustomizeExample.js
```diff
@@ -1,4 +1,5 @@
 import React from 'react';
+import { Tile } from 'aries-core';
```
taysea, line 2: Same question here for use of aries-core components. Would be an important discussion point.

aries-site/src/examples/templates/dashboards/DashboardExample.js
halocline: ☝️

aries-site/src/examples/templates/dashboards/DashboardExample.js
```diff
@@ -1,6 +1,210 @@
...
+        </Tile>
+      ))}
+    </Tiles>
+  );
+};
+
+const SidebarExample = ({ ...rest }) => {
+  const size = useContext(ResponsiveContext);
+  return (
+    <Box
```
taysea, line 117: This should probably be updated to use Sidebar

aries-site/src/examples/templates/dashboards/DashboardExample.js
```diff
@@ -1,6 +1,210 @@
...
+            a11yTitle="Help"
+            icon={<StatusUnknown color="text-xweak" />}
+          />
+        </Box>
+      )}
+    </Box>
+  );
+};
+
+const ScreenContainer = ({ mobile, ...rest }) => (
```
taysea, line 156: We can remove "mobile" here

aries-site/src/examples/templates/dashboards/DashboardExample.js
```diff
@@ -1,6 +1,237 @@
...
+            <Box gap="medium">
+              <Box>
+                <Text size="medium" weight="bold">
+                  {value.title}
+                </Text>
+                <Text size="medium">{value.subTitle}</Text>
+              </Box>
+            </Box>
+          </Box>
+          <Box flex />
```
taysea, line 112: Is there any way to get the same spacing without the empty box?

taysea: maybe some kind of justify between somewhere in a parent box? not sure, but just first thing that came to mind

aries-site/src/examples/templates/dashboards/DashboardExample.js
```diff
@@ -1,6 +1,237 @@
...
+
+const DashboardTiles = () => {
+  const size = useContext(ResponsiveContext);
+  return (
+    <Grid
+      gap="medium"
+      columns={size !== 'small' ? 'small' : 'auto'}
+      margin={size === 'small' ? { bottom: 'xlarge' } : undefined}
+    >
+      {data.map(value => (
```
taysea, line 84: Could we give value a more descriptive name such as `dashboardItem`

aries-site/src/examples/templates/dashboards/DashboardExample.js
```diff
@@ -1,6 +1,237 @@
...
+const PageHeaderExample = ({ title }) => (
+  <Header>
+    <Heading margin={{ vertical: 'medium' }} size="small">
+      {title}
+    </Heading>
+    <Button label="Manage" primary />
+  </Header>
+);
 
 export const DashboardExample = ({ ...rest }) => {
```
taysea, line 236: What is the ...rest here -- users won't have access directly to that would they? if we're able to remove it we should.

aries-site/src/examples/templates/forms/ChangePasswordExample.js
```diff
@@ -9,8 +9,30 @@ import {
...
+    status: 'error',
+  },
+  {
+    regexp: new RegExp('(?=.*?[#?!@$ %^&*-])'),
+    message: 'At least one special character or space',
+    status: 'error',
+  },
+];
+
+const FormContainer = ({ ...rest }) => {
```
taysea, line 25: Same as above, what is ..rest here and can it be removed or what value does it hold

taysea: Or in this case, rest may be "children" right? this might be slightly different than my above comment.

aries-site/src/examples/templates/dashboards/DashboardExample.js
britt6612: done good point!

aries-site/src/examples/templates/forms/ChangePasswordExample.js
britt6612: yes the rest here is children for the `formcontainer`

aries-site/src/examples/templates/dashboards/DashboardExample.js
britt6612: right here yes we dont need rest

aries-site/src/examples/templates/ScreenContainer.js
```diff
@@ -4,33 +4,31 @@ import { Box, ResponsiveContext } from 'grommet';
...
-              flex
-              {...rest}
-            />
-          </Box>
-          {size === 'small' && (
-            <SidebarExample
-              style={{ position: 'absolute', bottom: 0, left: 0 }}
-            />
-          )}
+  <ResponsiveContext.Consumer>
```
halocline, line 7: Is this file being used anywhere? Can it be deleted?

aries-site/src/examples/templates/dashboards/DashboardExample.js
```diff
@@ -1,13 +1,254 @@
...
+            icon={<StatusUnknown color="text-xweak" />}
+          />
+        </Footer>
+      )}
+    </Sidebar>
+  );
+};
+
+const ScreenContainer = ({ ...rest }) => (
+  <ResponsiveContext.Consumer>
```
halocline, line 171: For consistency, should this be refactored to useContext hook?

aries-site/src/examples/templates/list-views/ListScreenExample.js
```diff
@@ -1,6 +1,229 @@
...
+          { label: 'TRON_SERV8r' },
+          { label: 'HPE-Server-101' },
+        ]}
+      />
+    </Box>
+  </Header>
+);
+
+const ScreenContainer = ({ mobile, ...rest }) => (
+  <ResponsiveContext.Consumer>
```
halocline, line 201: For consistency, should this be refactored to useContext hook?

aries-site/src/examples/templates/ScreenContainer.js
britt6612: I think it can be deleted now thanks

halocline: File still exists? Not deleted?

halocline: nvm. I linked to the wrong view. Disregard.

PR #733, opened 2020-05-01
Form guidance

### LINKED ISSUE #593: Form
Acceptance criteria is based on alignment across the below 
- [x] Figma 
- [x] Theme
- [x] Grommet
- [ ] Code Examples
- [ ] Guidance  

We need to add some textual guidance explaining some of our recommendations for forms. For example:

- as few fields as possible
- default as much as possible
- progressive disclosure of fields
- how to handle required/optional fields
- variations if providing help guidance text
- the placement of the submit and cancel buttons
- language of buttons, "Add" vs. "Add User"
- how to show errors that aren't tied to a particular field
- how to handle results? ideally result is quick enough to show result immediately in outer context, what about things that take a long time? notifications? always leave the user in control of the UI, don't lock the user to wait for something to complete.

We should at least provide this sort of guidance in the design system site as there will be teams without designers who will not know how to navigate Figma.

**Figma File**
https://www.figma.com/file/3fkwBelW5UsCbfwdDCJkT8/HPE-Form-Templates?node-id=0%3A1

taysea:
Preview: https://deploy-preview-733--keen-mayer-a86c8b.netlify.app/templates/forms
Filled in form guidance for topics of:
- Usage
- Required vs optional fields
- Button labeling/placement
- Help text
- Form submission
- Accessibility

Help needed:
- how to show errors that aren't tied to a particular field
- "We should add guidance and an example illustrating "progressive disclosure" for things like "Advanced Options".
We don't currently have examples for the above options.

Additionally, I made some tweaks to form examples to remove color of Anchor and a couple other pieces to align with the latest Designs.

To avoid repetition of DOM ids, I created a 'SimpleSignUpExample' that mimics the sign up example, minus password help text. I couldn't use the same example twice because then multiple DOM elements would have the same id.

Closes #593

### DIFF
--- a/aries-site/src/examples/templates/forms/SettingsExample.js
+++ b/aries-site/src/examples/templates/forms/SettingsExample.js
@@ -50,7 +50,7 @@ export const SettingsExample = () => {
             onSubmit={({ value, touched }) => onSubmit({ value, touched })}
           >
             <Box margin={{ bottom: 'small' }}>
-              <Text color="text-weak" size="xsmall">
+              <Text color="text" size="xsmall" weight={600}>
                 Key feature
               </Text>
               <FormField
@@ -77,7 +77,7 @@ export const SettingsExample = () => {
               </FormField>
             </Box>
             <Box margin={{ bottom: 'small' }}>
-              <Text color="text-weak" size="xsmall">
+              <Text color="text" size="xsmall" weight={600}>
                 Key feature
               </Text>
               <FormField
--- a/aries-site/src/examples/templates/forms/SettingsExample.js
+++ b/aries-site/src/examples/templates/forms/SettingsExample.js
@@ -50,17 +50,18 @@ export const SettingsExample = () => {
             onSubmit={({ value, touched }) => onSubmit({ value, touched })}
           >
             <Box margin={{ bottom: 'small' }}>
-              <Text color="text" size="xsmall" weight={600}>
-                Key feature
-              </Text>
               <FormField
                 name="notifications"
-                label="Notifications"
-                component={CheckBox}
-                reverse
+                label="Key feature"
                 help="Description of feature"
-                toggle
-              />
+              >
+                <CheckBox
+                  name="notifications"
+                  label="Notifications"
+                  toggle
+                  reverse
+                />
+              </FormField>
               <FormField
                 htmlFor="notificationsVolume"
                 name="notificationsVolume"
@@ -77,17 +78,18 @@ export const SettingsExample = () => {
               </FormField>
             </Box>
             <Box margin={{ bottom: 'small' }}>
-              <Text color="text" size="xsmall" weight={600}>
-                Key feature
-              </Text>
               <FormField
                 name="doNotDisturb"
-                label="Do Not Disturb"
-                component={CheckBox}
+                label="Key Feature"
                 help="Description of feature"
-                reverse
-                toggle
-              />
+              >
+                <CheckBox
+                  name="doNotDisturb"
+                  label="Do Not Disturb"
+                  reverse
+                  toggle
+                />
+              </FormField>
               <FormField htmlFor="doNotDisturbHours" name="doNotDisturbHours">
                 <TextInput
                   id="doNotDisturbHours"
--- a/aries-site/src/pages/templates/forms.js
+++ b/aries-site/src/pages/templates/forms.js
@@ -119,8 +119,10 @@ const Forms = () => {
           <Text weight="bold">On submission, your Form should:</Text>
           <BulletedList
             items={[
-              `Display any relevant errors (if applicable) with text that 
-              explains the necessary criteria to fix errors`,
+              `In the case of an error, focus should be placed 
+              on the input element where the error occurred.`, 
+              `Display all relevant errors accompanied by text 
+               explaining actions needed to resolve.`,
               `Provide user with indication that form was submitted 
               successfully (e.g. notification, success page, etc.)`,
               `Allow user to continue navigating through the application 
--- a/aries-site/src/pages/templates/forms.js
+++ b/aries-site/src/pages/templates/forms.js
@@ -107,7 +107,7 @@ const Forms = () => {
             have a button with label "Update Password".
           </SubsectionText>
           <SubsectionText>
-            Specifity in button labels gives the user confidence in the action
+            Button label specificity provides users confidence in the action
             they are performing.
           </SubsectionText>
         </Subsection>
--- a/aries-site/src/pages/templates/forms.js
+++ b/aries-site/src/pages/templates/forms.js
@@ -120,11 +120,11 @@ const Forms = () => {
           <BulletedList
             items={[
               `In the case of an error, focus should be placed 
-              on the input element where the error occurred.`, 
-              `Display all relevant errors accompanied by text 
-               explaining actions needed to resolve.`,
+              on the input element where the error occurred.`,
               `Provide user with indication that form was submitted 
               successfully (e.g. notification, success page, etc.)`,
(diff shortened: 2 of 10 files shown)


aries-site/src/examples/templates/forms/SettingsExample.js
```diff
@@ -77,7 +77,7 @@ export const SettingsExample = () => {
               </FormField>
             </Box>
             <Box margin={{ bottom: 'small' }}>
-              <Text color="text-weak" size="xsmall">
+              <Text color="text" size="xsmall" weight={600}>
```
halocline, line 80: This input is a CheckBox. With the recent CheckBox work, I believe we can eliminate this Text, while relying on FormField label + moving CheckBox to be a child rather than specified as a prop.

aries-site/src/pages/templates/forms.js
```diff
@@ -29,11 +31,118 @@ const Forms = () => {
...
+          </Example>
+        </Subsection>
+      </ContentSection>
+      <ContentSection>
+        <Subsection name="Guidance">
+          <SubsectionText>
+            Forms provide a way to collect important data from a user. However,
+            this data is only collected effectively when Forms are built in a
+            way that users are inclined to fill out in full.
+          </SubsectionText>
```
halocline, line 50:
```suggestion
          <SubsectionText>
            Forms provide a way to collect important data from a user. However,
            this data is only collected effectively when Forms are built in a
            way in which users are inclined to fill out in full.
          </SubsectionText>
```

aries-site/src/pages/templates/forms.js
```diff
@@ -29,11 +31,118 @@ const Forms = () => {
...
+            Button labels should be specific to the action of the button. For
+            example, a <Anchor label="sign-up form" href="#sign-up" /> may have
+            a submit button with label "Sign Up". Alternately,{' '}
+            <Anchor label="password change form" href="#change-password" /> may
+            have a button with label "Update Password".
+          </SubsectionText>
+          <SubsectionText>
+            Specifity in button labels gives the user confidence in the action
+            they are performing.
+          </SubsectionText>
```
halocline, line 112:
```suggestion
          <SubsectionText>
            Specificity in button labels provide users confidence in the action
            they are performing.
          </SubsectionText>
```
OR
```suggestion
          <SubsectionText>
            Button label specificity provides users confidence in the action
            they are performing.
          </SubsectionText>
```

aries-site/src/pages/templates/forms.js
```diff
@@ -29,11 +31,118 @@ const Forms = () => {
...
+            Visual feedback is important to instill confidence in a user after
+            form submission.
+          </SubsectionText>
+          <Text weight="bold">On submission, your Form should:</Text>
+          <BulletedList
+            items={[
+              `Display any relevant errors (if applicable) with text that 
+              explains the necessary criteria to fix errors`,
+              `Provide user with indication that form was submitted 
+              successfully (notification, success page, etc.)`,
```
halocline, line 125:
```suggestion
              successfully (e.g. notification, success page, etc.)`,
```

halocline, line 125:
```suggestion
              `Provide user with indication that form was submitted 
              successfully (notification, success page, etc.)`,
              `Display any relevant errors (if applicable) with text that 
              explains the necessary criteria to fix errors`,    
```

aries-site/src/pages/templates/forms.js
```diff
@@ -29,11 +31,118 @@ const Forms = () => {
...
+        <Subsection name="Form submission" level={3}>
+          <SubsectionText>
+            Visual feedback is important to instill confidence in a user after
+            form submission.
+          </SubsectionText>
+          <Text weight="bold">On submission, your Form should:</Text>
+          <BulletedList
+            items={[
+              `Display any relevant errors (if applicable) with text that 
+              explains the necessary criteria to fix errors`,
```
halocline, line 123:
```suggestion
              `In the case of an error, focus should be placed 
              on the input element where the error occurred.`, 
              `Display all relevant errors accompanied by text 
               explaining actions needed to resolve.`,
```

aries-site/src/pages/templates/forms.js
```diff
@@ -29,11 +31,118 @@ const Forms = () => {
...
+          </SubsectionText>
+        </Subsection>
+        <Subsection name="Accessibility" level={3} gap="small">
+          <SubsectionText>
+            All Form input elements need to be contained within a FormField.
+          </SubsectionText>
+          <Text weight="bold">FormFields need to have:</Text>
+          <BulletedList
+            items={[
+              'An htmlFor property that matches the id of its child input',
```
halocline, line 142: `htmlFor` is Grommet specific, resolving to label's `for` attribute in the DOM. I'm not sure which direction we want to go... provide general guidance vs. this guidance assumes Grommet...

taysea: I will modify the copy to elaborate a little here.

aries-site/src/examples/templates/forms/SettingsExample.js
taysea: Great, let me look into that.

taysea:
> I think the guidance here hit on some real good points. Minor suggestions to consider and then good to merge.
> 
> I think that when we make a second pass on how Form guidance is presented, we should audit the examples more closely -- especially those examples relying on components which have/will change in the near future.

I'd agree that the presentation could benefit from a second pass, along with the specifics on examples as the implementation is further fleshed out.

aries-site/src/examples/templates/forms/SettingsExample.js
taysea: Made some changes to implement this. FormField will need a bit of discussion on the Grommet side to look at how labels are handled when Checkbox is the input -- leads to a situation where multiple labels are present for a single input since Checkbox and FormField need label props.

aries-site/src/pages/templates/forms.js
```diff
@@ -29,11 +31,121 @@ const Forms = () => {
...
+            possible when building a Form. Making Forms simple to complete leads
+            to increase in form completion.
           </SubsectionText>
+          <Text weight="bold">
+            Follow these guidelines when creating a Form:
+          </Text>
+          <BulletedList
+            items={[
+              'Use as few fields as possible',
+              'Wrap all input elements in a FormField',
```
britt6612, line 65:
```suggestion
              'Wrap all input elements within a FormField',
```

aries-site/src/pages/templates/forms.js
```diff
@@ -29,11 +31,121 @@ const Forms = () => {
...
+        <Subsection name="Accessibility" level={3} gap="small">
+          <SubsectionText>
+            All Form input elements need to be contained within a FormField.
+          </SubsectionText>
+          <Text weight="bold">FormFields need to have:</Text>
+          <BulletedList
+            items={[
+              `A "for" property that matches the id of its child input. If 
+              using Grommet, this is applied with the htmlFor prop.`,
+              'A clear, concise label',
```
britt6612, line 146: They should also have a name property as well...

aries-site/src/pages/templates/forms.js
```diff
@@ -29,11 +31,121 @@ const Forms = () => {
...
+            to increase in form completion.
           </SubsectionText>
+          <Text weight="bold">
+            Follow these guidelines when creating a Form:
+          </Text>
+          <BulletedList
+            items={[
+              'Use as few fields as possible',
+              'Wrap all input elements in a FormField',
+              'Fill in default values for as many fields as possible',
```
britt6612, line 66: I would also mention using pre filled inputs to reduce error. Like we have jane.smith@hpe.com in the signup example

PR #735, opened 2020-05-01
RadioButtonGroup Examples and Guidance

halocline:
[Preview](https://deploy-preview-735--keen-mayer-a86c8b.netlify.app/components/radiobuttongroup)

- [x] Figma
- [x] Guidance
- [x] Examples
- [x] Theme -- [PR here](https://github.com/grommet/grommet-theme-hpe/pull/75)
- [ ] Grommet -- Will have several issues created to represent outstanding work

Updates to `aries.js` to be removed once https://github.com/grommet/grommet-theme-hpe/pull/75 is merged


vavalos5:
First pass on RadioButton Group:

### **Radio Button Group**

A Radio Button Group is a set of related options within an element that allows only one entry selected at a time.

### **Guidance**

Keep labels concise and simple
Avoid long text descriptions on labels
If there are more than 5 options, or if the default value is reasonable most of the time, consider using Select instead, to avoid cluttering the experience 
Only use when a user needs to see all related options
Radio Buttons should always be listed vertically with one choice per line
Never ad sub-categories to a Radio Button Group


### **When to use Radio Button Group:**

Radio Button Group is used to lay out a small list of mutually related options that are easily visible. Radio Button Groups always require one choice, so it is important to be clear of what you ask from the user on the label.

If you would like the user to check more than one option, use Checkbox Group instead.

### **Variants**
Radio Button Groups are primarily used to select one single item from a list of two or more options. Here are some ways to use them:

**Validation**
Validation styling indicates that a selection was not made. An error message is accompanied by the styling that explains how to resolve the problem.

**Disabled**
Indicated that the Radio Button Group cannot be interacted with.

halocline:
Great. @vavalos5, I will drop into the page for folks to review.

aries-site/src/examples/components/radiobuttongroup/RadioButtonGroupValidationExample.js
```diff
@@ -0,0 +1,46 @@
...
+
+  const onChange = value => {
+    setPowerRegulation(value);
+    if (typeof powerRegulation === 'undefined') {
+      setMessage();
+    }
+  };
+
+  return (
+    <Form>
```
taysea, line 27: When I click on one of these radio buttons, the page goes to an error screen.

aries-site/src/pages/components/radiobuttongroup.js
```diff
@@ -21,13 +35,106 @@ const RadioButtonGroup = () => {
...
             docs="https://v2.grommet.io/radiobuttongroup?theme=hpe#props"
+            width="medium"
           >
             <RadioButtonGroupExample />
           </Example>
         </Subsection>
       </ContentSection>
+      <ContentSection>
+        <Subsection name="Guidance">
+          <SubsectionText>Keep labels concise and simple.</SubsectionText>
```
taysea, line 50: This feels a little awkward/orphaned to have separated from the other best practices.

halocline: Good call. Will move into lead of best practices.

aries-site/src/examples/components/radiobuttongroup/RadioButtonGroupValidationExample.js
halocline: will fix. Thanks.

halocline: Fixed.

aries-site/src/pages/components/radiobuttongroup.js
halocline: implemented

PR #742, opened 2020-05-04
take out any reference to carte

britt6612:
closes issue #740 

I commented one sentence out since we can reuse once 'Carte' gets approved. Then changed paths and alt names to take out Carte.


aries-site/src/pages/index.js
```diff
@@ -32,10 +32,11 @@ const Index = () => (
...
           </Heading>
           <Paragraph size="large" fill>
-            The Carte Design System was created to empower designers, developer,
-            and others to contribute in making great experiences for the
-            customer. Carte means “card game” in Italian. Pick your cards and
-            see what games you can play.
+            The Design System was created to empower designers,
+            developer, and others to contribute in making
+            great experiences for the customer.{' '}
+            {/* Carte means “card game” in Italian. Pick your cards and
```
taysea, line 38: Is there benefit to leaving this commented out? To keep the code clean, should we just remove this for now?

britt6612: I was just leaving it commented out in case we wanted to use the same sentence but we can take it out either way is fine with me

taysea: I'd say let's take it out for now since we can always go back in the commits/PRs to get that sentence if we want it

britt6612: sounds good!

aries-site/src/pages/index.js
```diff
@@ -32,10 +32,8 @@ const Index = () => (
             Mix, match, and stack cards to change the game
           </Heading>
           <Paragraph size="large" fill>
-            The Carte Design System was created to empower designers, developer,
-            and others to contribute in making great experiences for the
-            customer. Carte means “card game” in Italian. Pick your cards and
-            see what games you can play.
+            The Design System was created to empower designers, developer, and
+            others to contribute in making great experiences for the customer.
```
halocline, line 36: Headline references mix and match cards, we should retain that concept (i.e. don't lose the card reference altogether).
```suggestion
            The Design System was created to empower designers, developer, and
            others to contribute in making great experiences for the customer. Pick 
            your cards, then mix, match, and stack to see what experiences you 
            can create.
```
OR
```suggestion
            The Design System was created to empower designers, developer, and
            others to contribute in making great experiences for the customer. Pick 
            your cards, then mix, match, and stack to see what games you can play.
```

PR #754, opened 2020-05-06
Display status on page

### LINKED ISSUE #731: Add an indicator to each component and template indicating availability
The motivation behind this is to be reduce confusion about alignment between figma and theme.

The idea is to add something to each component and template page, perhaps even per example, indicating whether the figma, grommet, and theme are aligned or are still in progress. When aligned, indicate which versions of figma, grommet, and the theme are required.

Leaving the design details of how to visualize this up the team.

taysea:
Preview: https://deploy-preview-754--keen-mayer-a86c8b.netlify.app/components/button

Allow figma and grommet status to be displayed on details page. I've created a reusable component called <Status /> that takes a status prop. In structure.js we can define and object called status for any page.

For pages that are complete on both the grommet and figma side, no status is displayed. However, for those are are in progress either on grommet or figma or both, a status should be shown. I've added statuses to the pages I think need them based on this criteria, but if I've missed any let me know.

Closes #731

### DIFF
--- a/aries-site/src/components/content/Status.js
+++ b/aries-site/src/components/content/Status.js
@@ -1,0 +1,78 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Box, Text } from 'grommet';
+import { Figma, Grommet } from 'grommet-icons';
+import { IconCircle, IconTriangle } from '../icons';
+
+const types = {
+  figma: {
+    name: 'Figma',
+    icon: <Figma color="plain" />,
+  },
+  grommet: {
+    name: 'Grommet',
+    icon: <Grommet color="plain" />,
+  },
+};
+
+const statuses = {
+  'In Progress': {
+    icon: <IconTriangle size="small" color="status-warning" />,
+  },
+  Complete: {
+    icon: <IconCircle size="small" color="status-ok" />,
+  },
+};
+
+const StatusBadge = ({ status, type }) => {
+  return (
+    <Box direction="row">
+      <Box
+        background="background-front"
+        pad={{ vertical: 'xsmall', horizontal: 'small' }}
+        round={{ corner: 'left', size: 'xsmall' }}
+        direction="row"
+      >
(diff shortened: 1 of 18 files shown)


aries-site/src/components/content/Status.js
```diff
@@ -0,0 +1,78 @@
...
+    icon: <IconTriangle size="small" color="status-warning" />,
+  },
+  Complete: {
+    icon: <IconCircle size="small" color="status-ok" />,
+  },
+};
+
+const StatusBadge = ({ status, type }) => {
+  return (
+    <Box direction="row">
```
britt6612, line 29: can we do `row-responsive` ? 
Screen Shot 2020-05-07 at 8 41 41 AM

britt6612:
Looks good! Curious did you try putting it on the other side to be right aligned ?

britt6612:
something we can change later as well but to me it looks less busy..
Screen Shot 2020-05-07 at 8 51 57 AM

taysea:
> something we can change later as well but to me it looks less busy..
> Screen Shot 2020-05-07 at 8 51 57 AM

I had tried this as well, but as soon as I started to shrink the screen down, the badges got crunched -- let me try with row-responsive though. That might alleviate the issue.

aries-site/src/components/content/Status.js
taysea: Yes, good call.

britt6612:
> > something we can change later as well but to me it looks less busy..
> > Screen Shot 2020-05-07 at 8 51 57 AM
> 
> I had tried this as well, but as soon as I started to shrink the screen down, the badges got crunched -- let me try with row-responsive though. That might alleviate the issue.
> 
> Update: Even with row-responsive, the badges start to get crunched really early. I do agree that having them on the right side is less busy, but implementing this where the badges don't get squished is a bit tricky

oh okay yeah I didnt shrink the page I wasnt sure if you had tried it and had your reasoning. Let me look at what your referring to and see

PR #765, opened 2020-05-08
Updates Button usage to use default button

taysea:
Now that new theme styling is in, we don't need to use custom Box styling to achieve the desired Button appearance/behavior.


britt6612:
Screen Shot 2020-05-08 at 12 22 27 PM
This contrast seems off should this be white ? It is not reading the dark color

britt6612:
Is footer in figma updated to match button styling in here ?

taysea:
> Is footer in figma updated to match button styling in here ?

It should be no border, are you looking at the deploy or on local? If local, you may be on cached version of grommet

taysea:
> Screen Shot 2020-05-08 at 12 22 27 PM
> 
> This contrast seems off should this be white ? It is not reading the dark color

Good catch -- this should probably be an issue filed on grommet.

britt6612:
> > Screen Shot 2020-05-08 at 12 22 27 PM
> > This contrast seems off should this be white ? It is not reading the dark color
> 
> Also due to cached grommet I think.

ignore I didnt have latest grommet on my local this is fine in the deploy

aries-site/src/examples/templates/list-views/ListScreenExample.js
```diff
@@ -167,7 +167,7 @@ const PageHeaderExample = ({ title }) => (
 
 const AppHeaderExample = () => (
   <Header pad={{ vertical: 'small' }}>
-    <Button>
+    <Button plain>
```
halocline, line 170: For the <Text> instances on lines 180 & 183, I think we need to add color="text" -- right now coming across as text-strong

aries-site/src/examples/templates/dashboards/DashboardExample.js
```diff
@@ -194,7 +194,7 @@ const ScreenContainer = ({ ...rest }) => {
 
 const AppHeaderExample = () => (
   <Header pad={{ vertical: 'small' }}>
-    <Button>
+    <Button plain>
```
halocline, line 197: lines 207 & 210 need color="text"

PR #773, opened 2020-05-12
Use Markdown for docs pages -- new version using MDX in progress

taysea:
# What does this PR do?
Explores potential to use markdown for docs pages and statically build the site with this content.

## File structure and data flow
1. The `js` file in the `pages` directory (eg. `src/pages/compontents/button.js`) is modified to call `prepareMarkdown` in getStaticProps. Essentially, `button.js` is providing the markdown file path to `prepareMarkdown`.
2. `prepareMarkdown` handles parsing the markdown. It separates out any inline "example" objects from the markdown strings and extracts any frontmatter.
3. `prepareMarkdown returns two values: docs (an array containing markdown and examples separated from each other), and data (containing frontmatter information).
4. In `button.js`, `data` and `docs` get passed as the static props to the page. These are passed to `MarkdownDocs` which handles the page layout and styling given the data it has been provided.

**All the `.md` files live in `src/docs/....` where the structure follows that structure of the pages directory. So, `button.md` would live in `src/docs/components/button.md`.


taysea:
Closing because MDX route will be much more flexible. This first approach relied heavily on template strings which wasn't very flexible for the inline components we may want to include on a page.

PR #778, opened 2020-05-14
replace button for figma

### LINKED ISSUE #777: Remove Link from Color page
Remove "Use the colors" button from the Design System Color page 
https://design-system.hpe.design/foundation/color

britt6612:
[Preview](https://deploy-preview-778--keen-mayer-a86c8b.netlify.app/foundation/color)
closes #777 

There was 2 links for colors one figma and one designer. Figma button should be updated to how we have the other figma Buttons in components. As of now not sure of what the benefits of having Designer link for colors. 
We can have a colors screen. in designer but that would match the figma page which may be overkill

### DIFF
--- a/aries-site/src/pages/foundation/color.js
+++ b/aries-site/src/pages/foundation/color.js
@@ -1,5 +1,6 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+import { Figma } from 'grommet-icons';
 import { Anchor, Box, Button, ResponsiveContext } from 'grommet';
 
 import {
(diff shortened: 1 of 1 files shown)


halocline:
> I would make the button primary since it's the page's main CTA
🤔 
Is that the primary action we want users to take? To me, it is supplementary. I'd rather have folks consuming the content of the page, gaining exposure to color concepts and proper applications of color.

If a page has a single button, I don't think it has to get primary treatment. I think primary should be reserved for explicitly communicating "This is the step I want you to take next" or "If you were to only do one thing, do this." It should depend on use case, and I'd argue that "Open in Figma" is a supplementary action and not our primary call to action.

halocline:
> I would make the button primary since it's the page's main CTA

If we do go with primary, can we use a mono color for the Figma logo instead of the color? The colored logo gets lost on the green background. Probably would want to do a quick check on Figma's brand guidelines too.

britt6612:
> > I would make the button primary since it's the page's main CTA
> 
> If we do go with primary, can we use a mono color for the Figma logo instead of the color? The colored logo gets lost on the green background. Probably would want to do a quick check on Figma's brand guidelines too.

I wonder if we could make this a secondary with a border. I feel like with how it is without primary it doesnt really stand out as a button until you hover over it

taysea:
> > > I would make the button primary since it's the page's main CTA
> > 
> > 
> > If we do go with primary, can we use a mono color for the Figma logo instead of the color? The colored logo gets lost on the green background. Probably would want to do a quick check on Figma's brand guidelines too.
> 
> I wonder if we could make this a secondary with a border. I feel like with how it is without primary it doesnt really stand out as a button until you hover over it

I think that logic makes sense @halocline. We can now use `secondary` on the Button @britt6612 and that should apply the outline

PR #790, opened 2020-05-19
Fix deeplinking - don't merge, exploring performance issue

### LINKED ISSUE #320: Fix deeplinking to ensure page scroll is at correct section

taysea:
Wrapping the ThemeMode around the root component does something in the order of operations that makes sure the page doesn't rerender again after navigating which allows the correct scroll position to a hash to persist.

To test this out, navigate to `background colors` from the search. the page should bring you to the appropriate location on the page.

Resource: https://nextjs.org/docs/advanced-features/custom-app

Closes #320


aries-site/src/pages/_app.js
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+import { ThemeMode } from '../layouts';
+
```
halocline, line 3:
```suggestion
/* _app.js allows for customizing Next.js's default <App> component
 * Details: https://nextjs.org/docs/advanced-features/custom-app
 * 
 * The `Component` prop is the active `page`, so whenever you 
 * navigate between routes, `Component` will change to the new `page`. 
*/
```

britt6612:
Looks good do you want to comment out the test we have in` naviigation/deeplinks` ?

taysea:
I'm actually noticing a performance drop due to the increase in the `First Load JS shared by all` -- going to dive into that before merging this

taysea:
After looking into things further, the performance does not seem to be affected by this PR specifically. Instead, there are a few areas where performance could be improved and is being noted in #795

PR #808, opened 2020-05-22
Add variant for CheckBoxGroup with array of objects for options

taysea:
Adds a variant demonstrating that options can be an array of objects as opposed to just an array of strings.

### DIFF
--- a/aries-site/src/examples/components/checkboxgroup/CheckBoxGroupObjectExample.js
+++ b/aries-site/src/examples/components/checkboxgroup/CheckBoxGroupObjectExample.js
@@ -1,0 +1,29 @@
+import React from 'react';
+import { CheckBoxGroup, Form, FormField } from 'grommet';
+
+const objectOptions = [];
+for (let i = 1; i <= 3; i += 1) {
+  objectOptions.push({
+    label: `Option ${i}`,
+    val: i,
+  });
+}
+
+export const CheckBoxGroupObjectExample = () => {
+  return (
+    <Form>
+      <FormField
+        name="checkboxgroup-objectoptions"
+        fill
+        htmlFor="objectoptions-checkboxgroup"
+        label="Label"
+      >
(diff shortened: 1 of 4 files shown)


aries-site/src/examples/components/checkboxgroup/CheckBoxGroupObjectExample.js
```diff
@@ -0,0 +1,29 @@
...
+    <Form>
+      <FormField
+        name="checkboxgroup-objectoptions"
+        fill
+        htmlFor="objectoptions-checkboxgroup"
+        label="Label"
+      >
+        <CheckBoxGroup
+          options={objectOptions}
+          name="checkboxgroup-objectoptions"
```
britt6612, line 23: They shouldn't all be checked when you click on one ?

britt6612, line 23: `valueKey="val"`

britt6612:
```
    valueKey: PropTypes.string.description(
      `When the options array contains objects, this property indicates how
        to determine the value of each option. If a string is provided, 
        it is used as the key to retrieve each option's value.`,
    ),
  };
```

From Grommet docs then I think it should be good to go!

taysea:
> ```
>     valueKey: PropTypes.string.description(
>       `When the options array contains objects, this property indicates how
>         to determine the value of each option. If a string is provided, 
>         it is used as the key to retrieve each option's value.`,
>     ),
>   };
> ```
> 
> From Grommet docs then I think it should be good to go!

Thank you!

aries-site/src/examples/components/checkboxgroup/CheckBoxGroupObjectExample.js
taysea: Good catch, thanks! added it in now.

PR #810, opened 2020-05-27
Use text for color of Extend icon

### LINKED ISSUE #775: Extend icon in Dark Mode stays 'dark'
### Is this issue related to the design system core library or design system website?

design-system-core and design-system-website

### Expected Behavior
Extend icon goes to dark mode color

### Actual Behavior
Extend icon stays in light mode color when switching to dark mode
Screen Shot 2020-05-12 at 3 45 48 PM

### URL, screen shot, or CodeSandbox exhibiting the issue
https://design-system.hpe.design/extend

### Steps to Reproduce

1. switch to dark mode 

### Your Environment

- Design System version:
- Grommet version:
- Browser Name and version:
- Operating System and version (desktop or mobile):

taysea:
Use 'text' so that in dark mode, the text color will change appropriately.

Closes #775

### DIFF
--- a/aries-site/src/data/structure.js
+++ b/aries-site/src/data/structure.js
@@ -138,7 +138,7 @@ export const structure = [
   },
   {
     name: 'Extend',
-    color: 'black',
+    color: 'text',
     description:
       'Why does HPE have a design system? All the aesthetics, best practices, and information about the platform and how to wield it.',
     icon: (size, color) => <IconExtend size={size} color={color} />,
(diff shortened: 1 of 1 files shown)


aries-site/src/data/structure.js
```diff
@@ -138,7 +138,7 @@ export const structure = [
   },
   {
     name: 'Extend',
-    color: 'black',
+    color: 'text',
```
halocline, line 141:
```suggestion
    color: 'text-strong',
```

taysea: I'm good with that. The design calls for #000 in light mode, which is text-strong. So it seems fitting to use text-strong.

PR #811, opened 2020-05-27
Card sort order

halocline:
Closes https://github.com/hpe-design/design-system/issues/710

[Preview](https://deploy-preview-811--keen-mayer-a86c8b.netlify.app/)

**What does this PR do?**
Allows for custom sorting of cards based on category in which the fall and aligns with [sort order described here](https://github.com/hpe-design/design-system/issues/710#issuecomment-631779829).

**Work included**
- For easier object manipulation/maintainability, moves Foundation, Components, Templates, and Extend pages into their own files
- Creates a Structure class which holds some utility functions assist with common sorting patterns and allow for function chaining
- Modifies index.js for Home, Foundation, Components to facilitate specific sort orders


aries-site/src/data/structures/Structure.js
```diff
@@ -0,0 +1,27 @@
...
+
+  sortByCardOrder() {
+    return this.sort((a, b) => a.cardOrder - b.cardOrder);
+  }
+
+  sortByCategory(weights) {
+    return this.sort((a, b) => weights[a.category] - weights[b.category]);
+  }
+
+  sortByName() {
```
taysea, line 20: Is this alphabetizing? If so, can we just do `this.sort()` since sort by default will alphabetize?

halocline: For an array of strings we could. Since we are sorting objects by a property, I think we have to go long form, right?

taysea: Ahh, yes I think you're right there

aries-site/src/data/structures/Structure.js
```diff
@@ -0,0 +1,27 @@
+/* Utility functions to support page manipulations */
+
+export class Structure extends Array {
```
taysea, line 3: File structure wise, would it make more sense for these functions to live in utils? I could go either way here, because the proximity is nice, but since we're calling them "utility functions" my first thought is that it'd live in utils.

aries-site/src/pages/components/index.js
```diff
@@ -40,7 +40,19 @@ const Components = () => (
           </Paragraph>
         </Box>
       </PageIntro>
-      <CardGrid cards={cards} />
+      <ContentSection>
+        <Subsection name="Layouts">
```
taysea, line 44: Really like having these h2s on the page. helps a lot to feel less overwhelmed by cards

aries-site/src/data/structures/Structure.js
halocline: I modified the description a bit. If it's okay, I'd like to leave them as is for the time being since they are particularly related to the stucture file. Can migrate later if we see additional use cases.

taysea: Sounds good to me.

halocline:
> LGTM!
> 
> With the Templates page, I could see in the future there could be benefit to grouping the into categories as well -- "Navigation", "Page Layouts", etc. but for now I think it's fine as is.

Good call. I agree.

PR #825, opened 2020-06-01
Remove statuses from checkbox/checkboxgroup

taysea:
As of now, checkbox and checkbox group have been marked as "complete" in design and dev. This removes the status badges.

### DIFF
--- a/aries-site/src/data/structures/components.js
+++ b/aries-site/src/data/structures/components.js
@@ -144,10 +144,6 @@ export const components = [
       component: () => <CheckBoxPreview />,
       background: 'background-front',
     },
-    status: {
-      figma: statuses.complete,
-      grommet: statuses.inProgress,
-    },
   },
   {
     name: 'CheckBoxGroup',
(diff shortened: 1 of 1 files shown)


halocline:
> As of now, checkbox and checkbox group have been marked as "complete" in design and dev. This removes the status badges.

So, if a page does not have status badges, that means the design and dev assets are complete? As opposed to setting both to complete and display green for both?

taysea:
> > As of now, checkbox and checkbox group have been marked as "complete" in design and dev. This removes the status badges.
> 
> So, if a page does not have status badges, that means the design and dev assets are complete? As opposed to setting both to complete and display green for both?

Yes, I think the pattern we're trying to have is that status badges are only on pages where there is a discrepancy between design and dev states.

halocline:
> Yes, I think the pattern we're trying to have is that status badges are only on pages where there is a discrepancy between design and dev states.

Good to know. I missed that somewhere along the line. Thanks.

PR #874, opened 2020-06-08
Designer button label

halocline:
Closes https://github.com/hpe-design/design-system/issues/863

**Changes**
- Changed designer button label to read "Open in Grommet Designer"
- Changed Example control buttons to be lain out in Grid for better responsiveness and button alignment/justification
  - When reviewing, please include Safari - as Grid can be more fussy there ;).

### DIFF
--- a/aries-site/src/layouts/content/ExampleControls.js
+++ b/aries-site/src/layouts/content/ExampleControls.js
@@ -44,7 +44,7 @@ export const ExampleControls = ({ designer, docs, figma, setShowLayer }) => (
         <Button
           href={designer}
           icon={<Grommet color="plain" />}
-          label="Open in Grommet"
+          label="Open in Grommet Designer"
           target="_blank"
         />
       )}
--- a/aries-site/src/layouts/content/ExampleControls.js
+++ b/aries-site/src/layouts/content/ExampleControls.js
@@ -1,6 +1,6 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import { Box, Button } from 'grommet';
+import { Box, Button, Grid } from 'grommet';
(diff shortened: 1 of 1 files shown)


aries-site/src/layouts/content/ExampleControls.js
```diff
@@ -39,12 +39,18 @@ export const ExampleControls = ({ designer, docs, figma, setShowLayer }) => (
     pad={{ horizontal: 'medium', vertical: 'small' }}
     round={{ corner: 'bottom', size: 'small' }}
   >
-    <Box direction="row-responsive" gap="small">
+    <Grid
+      columns={{ count: 'fill', size: ['small', 'auto'] }}
```
taysea, line 43: Screen Shot 2020-06-08 at 4 02 07 PM

The "see fullscreen" text is getting forced to two lines even on desktop where there's enough space. Also, on mobile it doesn't seem like there will ever be enough space for both so maybe the entire thing should be single column?
Screen Shot 2020-06-08 at 4 01 49 PM

aries-site/src/layouts/content/ExampleControls.js
```diff
@@ -1,77 +1,73 @@
...
+        side: 'top',
+        color: 'background-back',
+        size: 'xsmall',
+      }}
+      justify="between"
+      pad={{ horizontal: 'medium', vertical: 'small' }}
+      round={{ corner: 'bottom', size: 'small' }}
+    >
+      <Grid
+        columns={{ count: 'fill', size: ['small', 'auto'] }}
```
taysea, line 25: The responsiveness is better on the most recent, but the spacing between the buttons on the left isn't uniform:
Screen Shot 2020-06-09 at 9 21 06 AM

"Open docs" is further away from the others

PR #907, opened 2020-06-16
POC Page Layouts

### LINKED ISSUE #887: Create initial page layout concept
Extension of the initial page layout exploration #797 

Create an initial concept of the "Page Layouts" page on the Design System site incorporating course graphic views, components and related content.

Success: to complete an initial concept of the "Page layouts" page on the website that we can meet and discuss for further refinement. 

### Deliverables
- [ ] update design system site with a single page that has an initial version, enough for team review

expected to take 1-2 days.

taysea:
This deploy is scheduled to be discussed on Thursday's meeting, but any feedback prior is welcomed!

Preview: https://deploy-preview-907--keen-mayer-a86c8b.netlify.app/templates/page-layouts

This takes discussions from last week's meetings and implements them as a first pass at a Page Layouts page. It seemed the conclusion from the meeting was that Page Layouts would be more coarse/high-level demonstration of how content areas layout on a page, but that content-rich examples would be housed on separate template pages.

After a follow-up call with Maggie and Vicky, we decided to focus on 4 variations from [this Figma file](https://www.figma.com/file/aGU4dgPXgnxuSKkoxPU8x4/Page-layout?node-id=0%3A1). If other layouts feel necessary, please provide feedback of which should be added.
This PR implements the high-level content areas and links to the relevant component pages (each containing rich examples of how components could will these layout areas). Once full-page templates have been created we can add them in.

TBD: How do we want to present scrolling? Do we want to show examples with sticky headers? etc.

### DIFF
--- a/aries-site/src/examples/templates/index.js
+++ b/aries-site/src/examples/templates/index.js
@@ -2,4 +2,5 @@ export * from './dashboards';
 export * from './forms';
 export * from './hub-spoke-navigation';
 export * from './list-views';
+export * from './page-layouts';
 export * from './persistent-navigation';
--- a/aries-site/src/examples/templates/page-layouts/HeaderFooterExample.js
+++ b/aries-site/src/examples/templates/page-layouts/HeaderFooterExample.js
@@ -1,0 +1,47 @@
+import React from 'react';
+import { Header, Box, Footer, Main, ResponsiveContext, Text } from 'grommet';
+
+export const HeaderFooterExample = () => {
+  return (
+    <AppContainer>
+      <Box flex overflow="auto">
+        <Box height={{ min: '100%' }} gap="small">
+          <Header
+            fill="horizontal"
+            height="xxsmall"
+            background="green!"
+            pad="small"
+          >
+            <Text weight="bold" color="text-strong">
+              Header
+            </Text>
+          </Header>
+          <Main flex="grow" fill={undefined} background="orange" pad="small">
+            <Text weight="bold">Main</Text>
+          </Main>
+          <Footer
+            fill="horizontal"
+            height="xxsmall"
+            background="blue"
+            pad="small"
+          >
+            <Text weight="bold">Footer</Text>
+          </Footer>
+        </Box>
+      </Box>
+    </AppContainer>
+  );
+};
+
+const AppContainer = ({ ...rest }) => {
+  const size = React.useContext(ResponsiveContext);
+  return (
+    <Box
+      direction={size === 'small' ? 'column-reverse' : 'row'}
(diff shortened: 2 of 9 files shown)


halocline:
I think the items which add the greatest value are those that demonstrate the interplay of the sections. To me those interactions are:
- How layouts translate across various screen sizes.
  - We have the "See Fullscreen" which does allow for desktop/mobile representations. However, is this effective? Is it findable? Do people even know this feature exists?
  - Are there alternatives to demonstrating breakpoints/responsiveness? Does a .gif or some other motion convey this more effectively?
- The need to demonstrate scrolling behavior
  - For example, if a Sidebar is present, I should be able to see that it is ever-present and that all content to its right scrolls vertically as the sidebar remains fixed.
- What is the behavior for when a Sidebar goes from narrow/minimized state to an expanded state? Does it cause the other layout elements to reduce width? Does the sidebar layer atop that content? (I think it should be the former, but just listing out questions/options)
- Advisement on container "max-width"s? Or does this live in the next fidelity layer of content?

> TBD: How do we want to present scrolling? Do we want to show examples with sticky headers? etc.
I'd advocate for 'yes' we should show sticky headers, but do so with guidance and recommendations (e.g. on smaller screens sticky headers occupy valuable real estate, however the need for a header's stickiness can inferred from and be set by a user's actions, such as scroll direction).

@ericsoderberghp's Designer example demonstrates some options for how to present scrolling: https://designer.grommet.io/fixed-header?id=page-layouts-eric-soderberg-hpe-com&mode=edit

halocline:
Other tangential notes:

- Square vs. rounded content containers?
- Consistency in color choices between site and Figma
- Order of examples?
  - Simple --> Complex? vs. Most interesting --> ?
- I like the idea of having the intra-page "jump-to" links at the top of the page.
  - This provides users with a preview of the content contained in the page, which facilitates scanability (does this page contain content in which I am interested?), navigation, and anticipation
  - Visually, Is this the best way to display this preview/table of contents? Alternatives to consider?
- Taxonomy for layouts? Eric has [labeled some layouts here](https://designer.grommet.io/fixed-header?id=page-layouts-eric-soderberg-hpe-com&mode=edit). Having a name is nice, but am wondering about the benefits of using straight-forward labels as currently shown. Kind of leaning toward the latter and only introducing names if they are needed to address a taxonomy issue in the future.

taysea:
Thanks for all the feedback, I agree with these points. As a quick feedback re: intra-page jump to links, I think it's an important point that extends to other pages. I believe there is a backlog issue regarding how to make the "Variants" more accessible. Maybe we could re-use the solution that comes from this page on those as well.

taysea:
Closing this PR since feedback is being incorporated into v2

PR #931, opened 2020-06-23
POC wide layout display

### LINKED ISSUE #919: Add wide-screen guidance and examples to Page Layouts
Related to https://github.com/hpe-design/design-system/issues/917. Add any additional guidance regarding how the layouts work in wide screen resolutions. For instance, anything without a sidebar could have a maximum width of content that is centered on the page. This would be a guidance, not a requirement, and we might explain that wide tables or data visualizations could take advantage of the wider screen real-estate. For sidebar layouts, ???

### Deliverables

- [ ] Figma screens showing wide resolution behavior
- [ ] Guidance text in Page Layouts
- [ ] Examples (we can split this part into a separate issue if needed)

expected to take 2-3 days, including collaboration between designers and developers.

taysea:
Preview: https://deploy-preview-931--keen-mayer-a86c8b.netlify.app/templates/page-layouts

This PR takes a pass at showing page layouts on wide monitors. Based on the designs we have, there is a max-width for content that then gets centered once the width has been exceeded. The goal is to allow the user to visualize this behavior by giving a toggle between "Laptop" (an average size screen), "Desktop" (a large monitor), and mobile.

Closes #919


aries-site/src/layouts/content/Example.js
```diff
@@ -88,6 +83,53 @@ const syntax = {
   `,
 };
 
+export const screens = {
+  desktop: 'desktop',
+  mobile: 'mobile',
+  laptop: 'laptop',
+};
+
+const BrowserWrapper = ({ screen, ...rest }) => {
```
halocline, line 92: This results of this are super nice. Great work. 

I do think Example needs to graduate to its own directory and have `Example`, `ExampleControls`, `ResponsiveControls`, `BrowserWrapper` as components which live within.  (The ResponsiveControls modification is perfect)

Probably the `syntax` style definition as well.

Probably the details, code, relevantComponents show/hide block as well.

aries-site/src/layouts/content/Example.js
```diff
@@ -96,6 +138,7 @@ export const Example = ({
   docs,
   figma,
   height,
+  layout,
```
halocline, line 141: Can we name this to be a bit more explicit? 'layout' just seems too generic and applicable too broadly. 

screenContainer? device? formFactor? screenWidth? viewPort?

aries-site/src/layouts/content/Example.js
taysea: I was thinking the same, that it's time to break this out. I'll do some restructuring in this PR to clean things up.

aries-site/src/layouts/content/Example.js
taysea: Yes, I like `screenContainer` or maybe `browserContainer`

halocline: `browserContainer` +1 👍

PR #938, opened 2020-06-24
Add related content to page layouts

taysea:
Add Related Content section to Page Layouts page with:
- Dashboards
- Lists
- Navigation
- Grid
- Header
- Sidebar
- Footer

### DIFF
--- a/aries-site/src/data/structures/templates.js
+++ b/aries-site/src/data/structures/templates.js
@@ -108,6 +108,15 @@ export const templates = [
         fit: 'contain',
       },
     },
+    relatedContent: [
+      'Navigation',
+      'Dashboards',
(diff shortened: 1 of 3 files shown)


aries-site/src/data/structures/templates.js
```diff
@@ -108,6 +108,15 @@ export const templates = [
         fit: 'contain',
       },
     },
+    relatedContent: [
```
halocline, line 111: Can we update `available: false` to be `available: true`? This will allow PageLayouts card to be prioritized in the card sort displays.

taysea: Good call, thanks.

PR #939, opened 2020-06-25
Add build test

taysea:
Before merging a PR, we will check that the site successfully builds. This will be beneficial once pages have shifted to MDX format and team members are editing copy via github as opposed to locally. Currently the build check runs pre-commit, but this will allow it to run when a PR is opened on github.


taysea:
Closing since build has been incorporated in the testcafe github actions PR

PR #943, opened 2020-06-25
Run testcafe tests on PR

taysea:
Before merging a PR, we will check that the site successfully builds and e2e tests run. This will be beneficial once pages have shifted to MDX format and team members are editing copy via github as opposed to locally. Currently this check runs pre-commit, but this will allow it to run when a PR is opened on github.

The main difference between this script and the local script is that we run the tests in headless chrome since we are in the CI environment. A script called `test:ci` has been added and is triggered when a PR is opened.

Prior to running the tests, we install the latest dependencies via `yarn upgrade`. This is needed as opposed to just `yarn` because in my setup I ran into an issue where weren't in sync with the latest grommet / grommet-theme-hpe. Running `yarn upgrade` makes sure we're in sync.


.github/workflows/testcafe.yml
```diff
@@ -0,0 +1,19 @@
...
+    steps:
+      - uses: actions/checkout@v2
+      - name: Use Node.js
+        uses: actions/setup-node@v1
+        with:
+          node-version: '12.x'
+      - name: Install latest dependencies
+        run: yarn upgrade
+        working-directory: aries-site
+      - name: Build and run e2e tests
```
taysea, line 17: clarify that only running chrome

.github/workflows/testcafe.yml
```diff
@@ -0,0 +1,19 @@
+name: Run e2e tests
+on: pull_request
```
taysea, line 2: only run on PRs to master

.github/workflows/testcafe.yml
```diff
@@ -0,0 +1,19 @@
...
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+      - name: Use Node.js
+        uses: actions/setup-node@v1
+        with:
+          node-version: '12.x'
+      - name: Install latest dependencies
+        run: yarn upgrade
```
taysea, line 15: look at caching

taysea: check: --update-checksums

PR #973, opened 2020-07-09
Edit button v2

### LINKED ISSUE #968: Explore how to create link between docs page and place to edit in Github
Taking from the MDX discussion, we may want an easier way to go from locating an issue with the site copy to where to fix it in github. For example, MUI has an "Edit this page" button at the top of each page which directs you to that page's file in github. We might try something similar?
Screen Shot 2020-07-08 at 1 24 21 PM

taysea:
Placing the edit in the feedback page footer

Closes #968


ericsoderberghp:
This seems like a reasonable approach to start with. Can we move this along?

PR #984, opened 2020-07-16
add required with astrisk and new message

britt6612:
## Work Completed 

[Preview](https://deploy-preview-984--keen-mayer-a86c8b.netlify.app/templates/forms)
[figma](https://www.figma.com/file/3fkwBelW5UsCbfwdDCJkT8/HPE-Form-Templates?node-id=0%3A1)

- Added custom `formfield` that would add an asterisk to required fields next to label 
- used custom `formfield` with required 
- Changed message for required to match designs 

## To Do 

I will switch the regular `!` with new icon
once the icon is ready

### DIFF
--- a/aries-site/src/examples/templates/forms/PayExample.js
+++ b/aries-site/src/examples/templates/forms/PayExample.js
@@ -86,16 +86,7 @@ const RequiredFormField = props => {
   const { required, label, ...rest } = props;
   return (
     <FormField
-      label={
-        required ? (
-          <Box direction="row">
-            <Text size="xsmall">{label}</Text>
-            <Text size="xsmall">*</Text>
-          </Box>
-        ) : (
-          label
-        )
-      }
+      label={required ? `${label}*` : label}
       required={required}
       {...rest}
     />
--- a/aries-site/src/examples/templates/forms/PayExample.js
+++ b/aries-site/src/examples/templates/forms/PayExample.js
@@ -82,7 +82,7 @@ const creditCardMask = [
   },
 ];
 
-const FormFieldRequiredLabel = props => {
+const RequiredFormField = props => {
   const { required, label, ...rest } = props;
   return (
     <FormField
@@ -102,7 +102,7 @@ const FormFieldRequiredLabel = props => {
   );
 };
 
-FormFieldRequiredLabel.propTypes = {
+RequiredFormField.propTypes = {
   required: PropTypes.bool,
   label: PropTypes.string,
 };
@@ -164,7 +164,7 @@ export const PayExample = () => {
             <Heading level={4} margin={{ bottom: 'small', top: 'none' }}>
               Credit Card Information
             </Heading>
-            <FormFieldRequiredLabel
+            <RequiredFormField
               name="cardName"
               required
               htmlFor="cardName"
@@ -175,8 +175,8 @@ export const PayExample = () => {
                 name="cardName"
                 placeholder="Jane Smith"
               />
-            </FormFieldRequiredLabel>
-            <FormFieldRequiredLabel
+            </RequiredFormField>
+            <RequiredFormField
               htmlFor="cardNumber"
               required
               name="cardNumber"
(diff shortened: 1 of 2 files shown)


aries-site/src/examples/templates/forms/PayExample.js
```diff
@@ -118,6 +140,9 @@ export const PayExample = () => {
           pad={{ horizontal: 'xxsmall' }}
         >
           <Form
+            messages={{
+              required: '! This is required',
```
taysea, line 144: Would we consider updating grommet to allow a node as required message type? Designs call for the `!` to be bold but know this would require doing something with `<Box />` potentially.

taysea, line 144:
```suggestion
              required: '! This is a required field.',
```

aries-site/src/examples/templates/forms/PayExample.js
```diff
@@ -81,14 +82,35 @@ const creditCardMask = [
   },
 ];
 
+const FormFieldRequiredLabel = props => {
```
taysea, line 85: Would the name `RequiredFormField` or `FormFieldRequired` be more accurate? Wondering since this contains the entire FormField, not just the label.

aries-site/src/examples/templates/forms/ShippingExample.js
```diff
@@ -165,30 +187,33 @@ export const ShippingExample = () => {
             validate="blur"
             value={formValues}
             onChange={setFormValues}
+            messages={{
+              required: '! This is required',
```
taysea, line 191: Same comments above regarding accepting more than just a string

taysea, line 191:
```suggestion
              required: '! This is a required field.',
```

aries-site/src/examples/templates/forms/ShippingExample.js
```diff
@@ -126,15 +127,36 @@ const emailValidation = [
...
-        flex
-        pad={{ horizontal: 'medium', vertical: 'medium' }}
-        {...rest}
-       />
+      <Box flex pad={{ horizontal: 'medium', vertical: 'medium' }} {...rest} />
     </Box>
   );
 };
 
+const FormFieldRequiredLabel = props => {
```
taysea, line 135: Same as above regarding naming

aries-site/src/examples/templates/forms/ShippingExample.js
```diff
@@ -165,30 +187,33 @@ export const ShippingExample = () => {
...
+            }}
             onSubmit={({ value, touched }) => onSubmit({ value, touched })}
           >
             <Box>
               <Heading level={4} margin={{ bottom: 'small', top: 'none' }}>
                 Shipping Information
               </Heading>
               <Text margin={{ horizontal: 'small', vertical: 'xsmall' }}>
-                Shipping Address
+                Shipping Address *
```
taysea, line 200: This pattern seems a little suspicious from an accessibility perspective since we are not placing a label on the actual formfields. Not to be tackled in this PR, but would be a good grommet investigation about proper behavior for grouped formfields like this.

aries-site/src/examples/templates/forms/PayExample.js
britt6612: yes I left it for now just until that icon is made and then I can have an issue to refer to for grommet and then fix prop types in grommet

aries-site/src/examples/templates/forms/ShippingExample.js
britt6612: Yes I also was not a fan of this I put it there to match the designs but should be more discussion around how we group things in a `formfield` and still meet accessibility

aries-site/src/examples/templates/forms/PayExample.js
britt6612: I think `RequiredFormField` is a good naming

aries-site/src/examples/templates/forms/PayExample.js
```diff
@@ -81,14 +82,35 @@ const creditCardMask = [
...
+const RequiredFormField = props => {
+  const { required, label, ...rest } = props;
+  return (
+    <FormField
+      label={
+        required ? (
+          <Box direction="row">
+            <Text size="xsmall">{label}</Text>
+            <Text size="xsmall">*</Text>
+          </Box>
```
ShimiSun, line 94: Please see if this block could be simplified with only using `label${required ? ’ *' : ''}`

PR #998, opened 2020-07-28
add long form example

### LINKED ISSUE #1032: Long Form Focus Example
Currently in Figma Files for form validation there is a want for focus to go to the first error input on submit
### Is this issue related to the design system core library or design system website?
design system website

### Expected Behavior
When a long form is used the user should be taken to the first error field with scroll and focus

### Actual Behavior
There is no behavior for this currently 

### Steps to Reproduce

1. 2. 3.

### Your Environment

- Design System version:
- Grommet version:
- Browser Name and version:
- Operating System and version (desktop or mobile):

britt6612:
Preview: https://deploy-preview-998--keen-mayer-a86c8b.netlify.app/templates/forms#shipping

This PR demonstrates how `onValidate` prop of Form can be used in long forms to assign focus to a specific field with an error. In long forms, we want to place focus on the first field that has an error when user submits so that it is clear to them where the error occurred and what they need to fix without having to scroll. 

The logic here is that the user provides an array of the FormField `names`. The order corresponds to their order on the page. The `onValidate` function receives an object with `errors` and `infos` that correspond to the `name` of the FormField. using the ordered names array, we find the first one that matches the name of either an error or an info and place focus on that field.

Closes #1032


halocline:
> Just summarizing the workflow here and where the issue is, correct me if I'm missing something.
> 
> GOAL:
> 
> 1. Identify any required fields or fields that did not pass validation.
> 2. Focus on the first field with an error.
> 
> PROCESS:
> 
> 1. Click "Submit" button to validate the form and run `checkDom` function.
> 2. Within `checkDom`, scan the DOM and identify any fields with errors.
> 3. Update error state to contain the array of DOM nodes with error.
> 4. Anytime errors is updated, run `setFocus` which should focus on the first error input.
> 
> ERROR:
> When onSubmit is first clicked, the `spans` that `checkDom` is searching for (containing error messages) do not yet exist in the DOM, meaning that `errors` remains empty.

That is my understanding as well. Need to run the checkDom after re-render.

britt6612:
yes thank you @taysea for writing this up! 
I put the comments in the code

ShimiSun:
An idea regarding removing the 'sniffing' of errors from the DOM.
Grommet could allow the error object as part of a new `onValidate` prop,
So it would look something like:
```
onValidate = ({error, info}) => {}
```
While error and info will be the objects that will be returned on the callback.

This contribution will help eliminate the portion of looking for either the info or error elements on the DOM.

britt6612:
> An idea regarding removing the 'sniffing' of errors from the DOM.
> Grommet could allow the error object as part of a new `onValidate` prop,
> So it would look something like:
> 
> ```
> onValidate = ({error, info}) => {}
> ```
> 
> While error and info will be the objects that will be returned on the callback.
> 
> This contribution will help eliminate the portion of looking for either the info or error elements on the DOM.

I believe if we could have the errors and info back we could look at the one that would be furthest from the submit button then we could focus on that field. Of course this wouldnt work if we had a collapse field.

aries-site/src/examples/templates/forms/ShippingExample.js
```diff
@@ -196,8 +223,8 @@ export const ShippingExample = () => {
           pad={{ horizontal: 'xxsmall' }}
         >
           <Form
-            validate="blur"
             value={formValues}
+            // messages={{ required: requiredErrorMessage }}
```
britt6612, line 227: was there a reason for adding this if its below ?

aries-site/src/examples/templates/forms/ShippingExample.js
```diff
@@ -177,6 +177,33 @@ export const ShippingExample = () => {
...
+    'city',
+    'state',
+    'zipcode',
+    'isBusiness',
+    'fullName-shipping',
+    'phone-shipping',
+    'email-shipping',
+  ];
+
+  const onValidate = validationResults => {
```
britt6612, line 196: I know why and what your doing but maybe add in a quick comment

britt6612:
LGTM! Left a couple comments
So happy this worked! So much cleaner

aries-site/src/examples/templates/forms/ShippingExample.js
taysea: good idea.

aries-site/src/examples/templates/forms/ShippingExample.js
taysea: whoops comment should we taken out.

PR #999, opened 2020-07-28
add example for invalid credentials

britt6612:
## Work Completed 

Provided an example of invalid credentials and how the error should look on a form 

## Work to do 

Add icon when ready

### DIFF
--- a/aries-site/src/examples/templates/forms/PayExample.js
+++ b/aries-site/src/examples/templates/forms/PayExample.js
@@ -10,7 +10,7 @@ import {
   Text,
   TextInput,
 } from 'grommet';
-import { Apple, CreditCard } from 'grommet-icons';
+import { Apple, CreditCard, CircleAlert } from 'grommet-icons';
 
 const currentDate = new Date();
 const dateValidation = [
@@ -131,7 +131,17 @@ export const PayExample = () => {
         >
           <Form
             messages={{
-              required: '! This is a required field.',
+              required: (
+                <Box
+                  background="background-front"
+                  align="center"
+                  gap="xsmall"
+                  direction="row"
+                >
+                  <CircleAlert size="small" />
+                  <Text size="xsmall">This is a required field.</Text>
+                </Box>
+              ),
             }}
             value={formValues}
             onChange={setFormValues}
--- a/aries-site/src/examples/templates/forms/PayExample.js
+++ b/aries-site/src/examples/templates/forms/PayExample.js
@@ -105,6 +105,19 @@ const FormContainer = ({ ...rest }) => {
   );
 };
 
+const Error = ({ textError, ...rest }) => {
+  return (
+    <Box {...rest} align="center" gap="xsmall" direction="row">
+      <CircleAlert size="small" />
+      <Text size="xsmall">{textError}</Text>
+    </Box>
+  );
+};
+
+Error.propTypes = {
+  textError: PropTypes.string,
+};
+
 export const PayExample = () => {
   const [formValues, setFormValues] = React.useState({});
   // eslint-disable-next-line no-unused-vars
@@ -132,15 +145,10 @@ export const PayExample = () => {
           <Form
             messages={{
               required: (
-                <Box
+                <Error
                   background="background-front"
-                  align="center"
-                  gap="xsmall"
-                  direction="row"
-                >
-                  <CircleAlert size="small" />
-                  <Text size="xsmall">This is a required field.</Text>
-                </Box>
+                  textError="This is a required field."
+                />
               ),
             }}
             value={formValues}
--- a/aries-site/src/examples/templates/forms/SignInExample.js
+++ b/aries-site/src/examples/templates/forms/SignInExample.js
@@ -223,6 +223,7 @@ export const SignInExample = () => {
             {credentialError && (
               <Box
                 direction="row"
+                animation="fadeIn"
                 margin={{ top: 'medium', bottom: 'medium' }}
                 gap="xsmall"
                 round="4px"
--- a/aries-site/src/examples/templates/forms/SignInExample.js
+++ b/aries-site/src/examples/templates/forms/SignInExample.js
@@ -22,7 +22,7 @@ const emailErrorMessage = (
     direction="row"
   >
     <CircleAlert size="small" />
-    <Text size="xsmall">This is a required field.</Text>
+    <Text size="xsmall">Enter a valid email address.</Text>
   </Box>
 );
 
@@ -30,12 +30,12 @@ const emailValidation = [
   {
     regexp: new RegExp('[^@ \\t\\r\\n]+@'),
     message: emailErrorMessage,
-    status: 'info',
+    status: 'error',
   },
   {
     regexp: new RegExp('[^@ \\t\\r\\n]+@[^@ \\t\\r\\n]+\\.[^@ \\t\\r\\n]+'),
     message: emailErrorMessage,
-    status: 'info',
+    status: 'error',
   },
   {
     regexp: new RegExp('[^@ \\t\\r\\n]+@[^@ \\t\\r\\n]+\\.[^@ \\t\\r\\n]+'),
@@ -43,6 +43,19 @@ const emailValidation = [
     status: 'error',
   },
 ];
+
+const Error = ({ textError, ...rest }) => {
+  return (
+    <Box {...rest} align="center" gap="xsmall" direction="row">
+      <CircleAlert size="small" />
+      <Text size="xsmall">{textError}</Text>
+    </Box>
+  );
+};
+
+Error.propTypes = {
+  textError: PropTypes.string,
+};
 
 const FormContainer = ({ ...rest }) => {
   return (
@@ -166,15 +179,10 @@ export const SignInExample = () => {
             onChange={setFormValues}
             messages={{
               required: (
-                <Box
+                <Error
                   background="background-front"
-                  align="center"
-                  gap="xsmall"
-                  direction="row"
-                >
-                  <CircleAlert size="small" />
-                  <Text size="xsmall">This is a required field.</Text>
-                </Box>
+                  textError="This is a required field."
+                />
               ),
             }}
             onSubmit={({ value, touched }) => onSubmit({ value, touched })}
@@ -222,10 +230,7 @@ export const SignInExample = () => {
                 pad="small"
                 background={{ light: '#FC61613D', dark: '#C54E4B5C' }}
               >
-                <Box align="center" gap="xsmall" direction="row">
-                  <CircleAlert size="small" />
-                  <Text size="xsmall">Invalid credentials.</Text>
-                </Box>
+                <Error textError="Invalid credentials." />
               </Box>
             )}
             <Box align="start" margin={{ top: 'medium', bottom: 'small' }}>
--- a/aries-site/src/examples/templates/forms/SignInExample.js
+++ b/aries-site/src/examples/templates/forms/SignInExample.js
@@ -228,7 +228,7 @@ export const SignInExample = () => {
                 gap="xsmall"
                 round="4px"
                 pad="small"
-                background={{ light: '#FC61613D', dark: '#C54E4B5C' }}
+                background="validation-critical"
               >
                 <Error textError="Invalid credentials." />
               </Box>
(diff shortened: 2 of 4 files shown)


aries-site/src/examples/templates/forms/SignUpExample.js
```diff
@@ -221,6 +228,19 @@ export const SignUpExample = () => {
                 }
               />
             </FormField>
+            {credentialError && (
+              <Box
+                direction="row"
```
taysea, line 233: Can we do an `align="center"` here to make the text better aligned w the icon?

Screen Shot 2020-07-29 at 10 37 08 AM

Maybe instead of using `strong` we should use `Text` with `weight="bold"` to keep it more grommety?

aries-site/src/examples/templates/forms/SignUpExample.js
```diff
@@ -85,6 +85,8 @@ export const SignUpExample = () => {
     email: 'jane.smith@hpe.com',
   });
   const [passwordRules, setPasswordRules] = React.useState(passwordRulesStrong);
+  const [password] = React.useState('');
```
taysea, line 88: With useState does the format need to be `[password, setPassword]` ?

aries-site/src/examples/templates/forms/SignUpExample.js
taysea: Actually I just looked back at the final designs, and it looks like it's calling for a different icon now:
Screen Shot 2020-07-29 at 10 44 48 AM

britt6612:
> It feels like this invalid credentials example should be applied to the Sign In form as opposed to the Sign Up. It seems like a more common behavior that this would occur when your credentials are already in a database somewhere.

That sounds fine I can change it!

aries-site/src/examples/templates/forms/SignUpExample.js
britt6612: yeah I wasnt to worried about the alignment with icon since I knew it was going to be switched out

aries-site/src/examples/templates/forms/SignUpExample.js
britt6612: yeah I took out setPassword since I wasnt setting it I can put it back tho and just comment it out that its not being used

aries-site/src/examples/templates/forms/SignUpExample.js
```diff
@@ -100,6 +102,10 @@ export const SignUpExample = () => {
   // eslint-disable-next-line no-unused-vars
   const onSubmit = ({ value, touched }) => {
     // Your submission logic here
+    // Your call to api error will result in credential error
```
halocline, line 105:
```suggestion
    // For demonstration purposes, your call to API returns a credential error
```

halocline:
I think we need to discuss where these examples live. The purpose here is to demonstrate "Field-Level and Form-Level Validation." I think that deserves its own section in the page flow. 

@gregfuruiye This relates to our discussions over the past couple days and is a candidate for the new approach.

Section should be called "Field and Form Validation" with subsections: 1) "Field-Level Validation" with descriptive text/guidance (e.g. field-level validation should be performed for any client-side validation rules, surface field-level errors as soon as they are known so that the user can resolve them in "flow", perform field level validation onBlur ...), plus example demonstrating the behavior; 2) "Form-Level Validation" with descriptive text (e.g. for server-side validation, validation will often occur at the form level, use the server's response message to display messaging to help the user resolve the error, display a summary message above the submit button, in addition display field-level errors when possible, plus 'invalid credentials example' plus 'scroll-to error example'.

For the purposes of this PR, I think we can apply @taysea 's recommendations and proceed, however, can we create an issue for "Re-work Forms page content flow"?

britt6612:
> I think we need to discuss where these examples live. The purpose here is to demonstrate "Field-Level and Form-Level Validation." I think that deserves its own section in the page flow.
> 
> @gregfuruiye This relates to our discussions over the past couple days and is a candidate for the new approach.
> 
> Section should be called "Field and Form Validation" with subsections: 1) "Field-Level Validation" with descriptive text/guidance (e.g. field-level validation should be performed for any client-side validation rules, surface field-level errors as soon as they are known so that the user can resolve them in "flow", perform field level validation onBlur ...), plus example demonstrating the behavior; 2) "Form-Level Validation" with descriptive text (e.g. for server-side validation, validation will often occur at the form level, use the server's response message to display messaging to help the user resolve the error, display a summary message above the submit button, in addition display field-level errors when possible, plus 'invalid credentials example' plus 'scroll-to error example'.
> 
> For the purposes of this PR, I think we can apply @taysea 's recommendations and proceed, however, can we create an issue for "Re-work Forms page content flow"?

@halocline I agree my intentions were to get these examples in then rework the page to have a form layout at the top then the content based on that. Then after provide content based around validation and have those examples follow the content so essentially two sections on the form page

aries-site/src/examples/templates/forms/SignUpExample.js
taysea: I see, I think it would be helpful just to add a clarifying comment of why this is here. Something along the lines of this being used for demonstration purposes.

britt6612: done 👍

aries-site/src/examples/templates/forms/SignInExample.js
```diff
@@ -118,6 +123,10 @@ export const SignInExample = () => {
   // eslint-disable-next-line no-unused-vars
   const onSubmit = ({ value, touched }) => {
     // Your submission logic here
+    // For demonstration purposes, your call to API returns a credential error
```
taysea, line 126:
```suggestion
    // For demonstration purposes, we are mocking a validation where the password is incorrect.
    // This will cause the error state to appear.
```

aries-site/src/examples/templates/forms/PayExample.js
```diff
@@ -131,7 +131,17 @@ export const PayExample = () => {
...
+              required: (
+                <Box
+                  background="background-front"
+                  align="center"
+                  gap="xsmall"
+                  direction="row"
+                >
+                  <CircleAlert size="small" />
+                  <Text size="xsmall">This is a required field.</Text>
+                </Box>
```
halocline, line 143: Same comment, worth abstracting?

aries-site/src/examples/templates/forms/SignInExample.js
```diff
@@ -12,22 +12,34 @@ import {
...
+
+const emailErrorMessage = (
+  <Box
+    background="background-front"
+    align="center"
+    gap="xsmall"
+    direction="row"
+  >
+    <CircleAlert size="small" />
+    <Text size="xsmall">This is a required field.</Text>
```
halocline, line 25: Worth abstracting?

aries-site/src/examples/templates/forms/SignInExample.js
```diff
@@ -142,9 +164,23 @@ export const SignInExample = () => {
...
+              required: (
+                <Box
+                  background="background-front"
+                  align="center"
+                  gap="xsmall"
+                  direction="row"
+                >
+                  <CircleAlert size="small" />
+                  <Text size="xsmall">This is a required field.</Text>
+                </Box>
```
halocline, line 177: worth abstracting?

halocline:
Confused about why the background color is white on this error state:
Screen Shot 2020-07-30 at 2 13 22 PM

halocline:
Do we want to showcase this behavior in a Form Validation / Error handling section in addition to as a Sign In pattern?

It seems like they are two different topical areas -- 1) Here are some form patterns / use cases, 2) Here are some specific examples and guidance for how Field and Form validation should be handed.

It feels like we should have a Form component.

This comment block shouldn't hold up this PR, more related to how we approach presenting content as a whole and related to https://github.com/hpe-design/design-system/issues/986

britt6612:
> Confused about why the background color is white on this error state:
> Screen Shot 2020-07-30 at 2 13 22 PM

strange ill look at that!

aries-site/src/examples/templates/forms/SignInExample.js
```diff
@@ -222,10 +230,7 @@ export const SignInExample = () => {
                 pad="small"
                 background={{ light: '#FC61613D', dark: '#C54E4B5C' }}
```
taysea, line 231: Can we potentially access the theme color here instead of hard-coding a string?

aries-site/src/examples/templates/forms/PayExample.js
```diff
@@ -105,6 +105,19 @@ const FormContainer = ({ ...rest }) => {
...
 };
 
+const Error = ({ textError, ...rest }) => {
+  return (
+    <Box {...rest} align="center" gap="xsmall" direction="row">
+      <CircleAlert size="small" />
+      <Text size="xsmall">{textError}</Text>
+    </Box>
+  );
+};
```
halocline, line 115: Move ...rest to be last property so that it ensures whatever is supplied to rest gets applied and not overwritten.

```suggestion
const Error = ({ textError, ...rest }) => {
  return (
    <Box align="center" gap="xsmall" direction="row" {...rest} >
      <CircleAlert size="small" />
      <Text size="xsmall">{textError}</Text>
    </Box>
  );
};
```

aries-site/src/examples/templates/forms/PayExample.js
```diff
@@ -131,7 +144,12 @@ export const PayExample = () => {
         >
           <Form
             messages={{
-              required: '! This is a required field.',
+              required: (
+                <Error
+                  background="background-front"
+                  textError="This is a required field."
+                />
```
halocline, line 151: This might seem nit picky, but since these are being used as code samples, from a style and simplicity standpoint, I'd think we are trying to avoid unnecessarily creating props. We can make this more generic and less opinionated by passing through children.
```suggestion
                <Error>This is a required field.</Error>
```

aries-site/src/examples/templates/forms/PayExample.js
halocline, line 115: Related to: https://github.com/hpe-design/design-system/pull/999/files#r463792055

```suggestion
const Error = ({ children, ...rest }) => {
  return (
    <Box align="center" background="background-front" gap="xsmall" direction="row"  {...rest} >
      <CircleAlert size="small" />
      <Text size="xsmall">{children}</Text>
    </Box>
  );
};
```

aries-site/src/examples/templates/forms/SignInExample.js
```diff
@@ -176,6 +221,18 @@ export const SignInExample = () => {
...
               />
             </FormField>
+            {credentialError && (
+              <Box
+                direction="row"
+                margin={{ top: 'medium', bottom: 'medium' }}
+                gap="xsmall"
+                round="4px"
+                pad="small"
+                background={{ light: '#FC61613D', dark: '#C54E4B5C' }}
```
halocline, line 231: These look like they are specifying the error background colors, correct? 

Looks like these should be added to the theme so that we are not promoting use of custom HEX values/colors in the Design System.  Can you add an issue for "Add validation error background colors to theme"? Should include: 
- new color namespace - reference: https://design-system.hpe.design/foundation/color#input-palette
- apply new namespace here: https://github.com/grommet/grommet-theme-hpe/blob/1e352af3ec70f5ab3ebcdbaa55a74ddca0f536c1/src/js/index.js#L507
- apply new namespace in SignInExample.js

britt6612: yes! @taysea and I were just saying we did not want to hard code these they should be pulled from theme!

aries-site/src/examples/templates/forms/PayExample.js
britt6612: that sounds good! Also I think Im going to move the icon to be more of grommet theme option

aries-site/src/examples/templates/forms/SignInExample.js
```diff
@@ -176,6 +220,18 @@ export const SignInExample = () => {
                 label="Remember me"
               />
             </FormField>
+            {credentialError && (
+              <Box
+                direction="row"
+                margin={{ top: 'medium', bottom: 'medium' }}
```
halocline, line 226: Might make the appearance a little less abrupt. Thoughts?
```suggestion
                animation="fadeIn"
                margin={{ top: 'medium', bottom: 'medium' }}
```

britt6612: I just tried this! I think it is less abrupt more settle for an error

PR #1001, opened 2020-07-29
First pass at Wizard template

### LINKED ISSUE #997: First pass at Wizard template
### Deliverables

- [ ] Wizard page in design system
- [ ] links to Figma
- [ ] initial guidance, more guidance can follow in a separate backlog item later

Build off of https://github.com/hpe-design/design-system/issues/849

I suggest starting with a dev/design review of the figma content. We can create new backlog items for additional design work as needed.

taysea:
#### What does this PR do?

Preview: https://deploy-preview-1001--keen-mayer-a86c8b.netlify.app/templates/wizard
Designs: https://www.figma.com/file/GEhgYzx3b8MejBjg9gAxyA/Wizarding?node-id=71%3A406

First pass at Wizard page.
1. Implements both single column and two column example.
2. Provides starter for guidance based on what was provided in designs.

From a technical standpoint, these examples demonstrate how to create a multi-step form using state. However, in the future it may be beneficial to show a multi-step form example with routing (https://stackoverflow.com/questions/36841762/react-router-navigation-in-multi-step-form/36854195#36854195)

aries-site/src/pages/templates/wizard.mdx

All e2e tests pass. Examples have been viewed and interacted with in Chrome, Safari, and FF.
 
Interact with examples in Chrome, Safari, and FF.

Closes #997 

Yes.

Yes, backwards compatible.

### DIFF
--- a/aries-site/src/examples/templates/wizard/SimpleWizardExample.js
+++ b/aries-site/src/examples/templates/wizard/SimpleWizardExample.js
@@ -205,7 +205,11 @@ export const SimpleWizardExample = () => {
             />
           </Box>
         </Header>
-        <Box align="center" pad={size !== 'small' ? 'large' : 'medium'}>
+        <Box
+          align="center"
+          pad={size !== 'small' ? 'large' : 'medium'}
+          flex={false}
+        >
           <Box width="medium" gap="medium">
             <Box gap="medium" flex={false}>
               <Box>
--- a/aries-site/src/examples/templates/wizard/SimpleWizardExample.js
+++ b/aries-site/src/examples/templates/wizard/SimpleWizardExample.js
@@ -176,35 +176,7 @@ export const SimpleWizardExample = () => {
   return (
     <>
       <Box width={{ max: 'xxlarge' }} margin="auto" fill>
-        <Header
-          border={{ side: 'bottom', color: 'border-weak' }}
-          pad="small"
-          fill="horizontal"
-          justify="center"
-        >
-          <Box direction="row" flex>
-            {active > 1 && (
-              <Button
-                label={size !== 'small' ? `Step ${active - 1}` : undefined}
-                icon={<FormPreviousLink color="text-strong" />}
-                onClick={() => setActive(active - 1)}
-              />
-            )}
-          </Box>
-          <Box>
-            <Text color="text-strong" weight="bold">
-              Action Title
-            </Text>
-          </Box>
-          <Box direction="row" flex justify="end">
-            <Button
-              label={size !== 'small' ? 'Cancel' : undefined}
-              icon={<FormClose color="text-strong" />}
-              reverse
-              onClick={() => setOpen(true)}
-            />
-          </Box>
-        </Header>
+        <WizardHeader active={active} setActive={setActive} setOpen={setOpen} />
         <Box
           align="center"
           pad={size !== 'small' ? 'large' : 'medium'}
@@ -240,6 +212,47 @@ export const SimpleWizardExample = () => {
       {open && <CancellationLayer onSetOpen={setOpen} />}
     </>
   );
+};
+
+const WizardHeader = ({ active, setActive, setOpen }) => {
+  const size = useContext(ResponsiveContext);
+  return (
+    <Header
+      border={{ side: 'bottom', color: 'border-weak' }}
+      pad="small"
+      fill="horizontal"
+      justify="center"
+    >
+      <Box direction="row" flex>
+        {active > 1 && (
+          <Button
+            label={size !== 'small' ? `Step ${active - 1}` : undefined}
+            icon={<FormPreviousLink color="text-strong" />}
+            onClick={() => setActive(active - 1)}
+          />
+        )}
+      </Box>
+      <Box>
+        <Text color="text-strong" weight="bold">
+          Action Title
+        </Text>
+      </Box>
+      <Box direction="row" flex justify="end">
+        <Button
+          label={size !== 'small' ? 'Cancel' : undefined}
+          icon={<FormClose color="text-strong" />}
+          reverse
+          onClick={() => setOpen(true)}
+        />
+      </Box>
+    </Header>
+  );
+};
+
+WizardHeader.propTypes = {
+  active: PropTypes.number.isRequired,
+  setActive: PropTypes.func.isRequired,
+  setOpen: PropTypes.func.isRequired,
 };
 
 const CancellationLayer = ({ onSetOpen }) => {
--- a/aries-site/src/examples/templates/wizard/SimpleWizardExample.js
+++ b/aries-site/src/examples/templates/wizard/SimpleWizardExample.js
@@ -1,0 +1,285 @@
+import React, { useState, useContext } from 'react';
+import PropTypes from 'prop-types';
+import {
+  Box,
+  Button,
+  CheckBoxGroup,
+  Form,
+  FormField,
+  Header,
+  Heading,
+  Layer,
+  List,
+  RadioButtonGroup,
+  ResponsiveContext,
+  Select,
+  Text,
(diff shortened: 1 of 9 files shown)


britt6612:
The scroll seems a little off on desktop view

britt6612:
When you click next on it takes you below at a weird spot in the page 
Screen Shot 2020-07-30 at 10 54 10 PM

aries-site/src/examples/templates/wizard/SimpleWizardExample.js
```diff
@@ -0,0 +1,289 @@
...
+    >
+      <Box pad="large" gap="small" width="large">
+        <Box>
+          <Heading color="text-strong" margin="none">
+            Cancel
+          </Heading>
+          <Text color="text-strong">Action Title</Text>
+        </Box>
+        <Text color="text-strong">
+          Cancelling setup will lose all of your progress. Are you sure you want
```
britt6612, line 260: was this actually suppose to happen or just example ?

aries-site/src/pages/templates/wizard.mdx
```diff
@@ -0,0 +1,66 @@
...
+  figma="https://www.figma.com/file/GEhgYzx3b8MejBjg9gAxyA/Wizarding?node-id=1%3A2"
+  screenContainer
+  width="100%"
+>
+  <SimpleWizardExample />
+</Example>
+
+## Guidance
+
+Use a wizard when your form involves multiple steps. This makes the experience more manageable for the user.
```
britt6612, line 15: while helping to limit user error ?

britt6612: Adding in that a good way to deal with long forms is to break them down into multiple steps

taysea: Good idea. I will add something here. I think we also need to add some guidance around validation and how validation occurs in a wizard.

taysea:
> When you click next on it takes you below at a weird spot in the page
> Screen Shot 2020-07-30 at 10 54 10 PM

Good note -- it seems that scroll position persists since we are just swapping out state, and it feels like we want to scrollToTop when the next / previous step is reached

aries-site/src/pages/templates/wizard.mdx
britt6612: Yeah I was going to ask about validation as well

aries-site/src/examples/templates/wizard/SimpleWizardExample.js
britt6612: okay yeah I wasnt sure if we actually wanted to mock this behavior with state since if you cancel now the values are still there or they will get the idea and have their own state they will update

ericsoderberghp:
> Good note -- it seems that scroll position persists since we are just swapping out the active step using state, and it feels like we want to scrollToTop when the next / previous step is reached. Also, it looks like the `Next` button is receiving focus automatically here as well which may be pulling the scroll position down.

Is there a way to either increase the Example height or reduce the form height such that we don't have to scroll?

ericsoderberghp:
I think we should add a Related link to Form.

Looking good overall. I think this is a good first pass.

aries-site/src/examples/templates/wizard/SimpleWizardExample.js
taysea: Gotcha, good point. Yeah, wondering how far we want to mock the behavior. My thought is that if you truly "cancelled", the wizard would close and you'd go back to the application wherever you were prior to opening. Will get clarification on this behavior.

taysea:
> > Good note -- it seems that scroll position persists since we are just swapping out the active step using state, and it feels like we want to scrollToTop when the next / previous step is reached. Also, it looks like the `Next` button is receiving focus automatically here as well which may be pulling the scroll position down.
> 
> Is there a way to either increase the Example height or reduce the form height such that we don't have to scroll?

We definitely can do that. However, this behavior exists even if I enter full screen and then shrink my window height. While a wizard is definitely best if you allow your window to grow, I would still expect it to bring me to the top of each step as I advancee.

ericsoderberghp:
agreed

aries-site/src/pages/templates/wizard.mdx
taysea: I've added some initial guidance on validation but still need to add it into the examples themselves.

aries-site/src/examples/templates/wizard/StickyHeaderFooterExample.js
```diff
@@ -0,0 +1,286 @@
...
+              value={formValues}
+              onChange={setFormValues}
+              onSubmit={({ value }) => console.log(value)}
+            >
+              {/* Index an array starts at 0 */}
+              {steps[active - 1].inputs(active, setActive)}
+            </Form>
+          </Box>
+        </Box>
+        <Footer
```
halocline, line 209: This is feeling like the button is too far removed from form steps. Is this trying to avoid having the button "jump" around from step to step?

 
Screen Shot 2020-07-31 at 2 06 00 PM

aries-site/src/examples/templates/wizard/StickyHeaderFooterExample.js
```diff
@@ -0,0 +1,286 @@
...
+          </Box>
+          <Box direction="row" flex justify="end">
+            <Button
+              label={size !== 'small' ? 'Cancel' : undefined}
+              icon={<FormClose color="text-strong" />}
+              reverse
+              onClick={() => setOpen(true)}
+            />
+          </Box>
+        </Header>
```
halocline, line 175: Abstract to <WizardHeader />to aid readability?

aries-site/src/examples/templates/wizard/StickyHeaderFooterExample.js
halocline, line 209: Abstract to `<WizardFooter />` to aid readability?

aries-site/src/pages/templates/wizard.mdx
```diff
@@ -0,0 +1,87 @@
...
+## Variants
+
+Common wizard variants.
+
+### Sticky Header and Footer Wizard
+
+Using a sticky header and footer allows the user to easily access buttons that allow to advance or move back within the wizard.
+
+<Example
+  code="https://raw.githubusercontent.com/hpe-design/design-system/master/aries-site/src/examples/templates/wizard/TwoColumnWizardExample.js"
```
halocline, line 68:
```suggestion
  code="https://raw.githubusercontent.com/hpe-design/design-system/master/aries-site/src/examples/templates/wizard/StickyHeaderFooterExample.js"
```

aries-site/src/pages/templates/wizard.mdx
```diff
@@ -0,0 +1,87 @@
...
+  figma="https://www.figma.com/file/GEhgYzx3b8MejBjg9gAxyA/Wizarding?node-id=71%3A406"
```
halocline, line 69:
```suggestion
  figma="https://www.figma.com/file/GEhgYzx3b8MejBjg9gAxyA/Wizarding?node-id=71%3A22"
```

aries-site/src/examples/templates/wizard/StickyHeaderFooterExample.js
taysea: This to demonstrate where the next/finish button persists in the across all steps in the same position. I agree it does feel far away from the steps though

taysea:
> Couple suggestions.
> 
> Sticky footer experience seems a bit weird to me (especially on large screens), but want to think on it more. Would not let that get in the way of publishing. Maybe a background-contrast? Window height based padding/footer height?

Agreed on your thought that is looks strange on large screens -- we can add that as a behavior to discuss start of next week?

PR #1002, opened 2020-07-30
fix background color for validation password

britt6612:
[Preview](https://deploy-preview-1002--keen-mayer-a86c8b.netlify.app/templates/forms)
In the first iteration of password validation the background was just white regardless of whether the checks were true for password. I wanted to match up better with designs so the background color is the same error red color if all the checks are not meet.

### DIFF
--- a/aries-site/src/examples/templates/forms/SignUpExample.js
+++ b/aries-site/src/examples/templates/forms/SignUpExample.js
@@ -1,4 +1,5 @@
 import React from 'react';
+import PropTypes from 'prop-types';
 import {
   Anchor,
   Box,
@@ -12,22 +13,55 @@ import {
   Text,
   TextInput,
 } from 'grommet';
-import { FormCheckmark, FormClose } from 'grommet-icons';
+import { FormCheckmark, CircleAlert } from 'grommet-icons';
+
+const Error = ({ children, ...rest }) => {
+  return (
+    <Box direction="row" gap="xsmall" {...rest}>
+      <Box flex={false} margin={{ top: 'hair' }} pad={{ top: 'xxsmall' }}>
+        <CircleAlert size="small" />
+      </Box>
+      <Text size="xsmall">{children}</Text>
+    </Box>
+  );
+};
+
+Error.propTypes = {
+  children: PropTypes.object,
+};
+
+const passwordRequirements = [
+  {
+    regexp: new RegExp('(?=.*?[A-Z])(?=.*?[a-z])(?=.*?[#?!@$ %^&*-]).{8,}'),
+    message: (
+      <Error background="background-front">
+        Password requirements not meet.
+      </Error>
+    ),
+    status: 'error',
+  },
+];
 
 const emailValidation = [
   {
     regexp: new RegExp('[^@ \\t\\r\\n]+@'),
-    message: 'Missing an @?',
-    status: 'info',
+    message: (
+      <Error background="background-front">Enter a valid email address.</Error>
+    ),
+    status: 'error',
   },
   {
     regexp: new RegExp('[^@ \\t\\r\\n]+@[^@ \\t\\r\\n]+\\.[^@ \\t\\r\\n]+'),
-    message: 'Missing an .?',
-    status: 'info',
+    message: (
+      <Error background="background-front">Enter a valid email address.</Error>
+    ),
+    status: 'error',
   },
   {
     regexp: new RegExp('[^@ \\t\\r\\n]+@[^@ \\t\\r\\n]+\\.[^@ \\t\\r\\n]+'),
-    message: "Email address doesn't look quite right",
+    message: (
+      <Error background="background-front">Enter a valid email address.</Error>
+    ),
     status: 'error',
   },
 ];
@@ -153,6 +187,7 @@ export const SignUpExample = () => {
             </FormField>
             <FormField
               label="Password"
+              validate={passwordRequirements}
               htmlFor="password-sign-up"
               name="password"
               required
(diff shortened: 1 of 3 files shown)


aries-site/src/examples/templates/forms/SignUpExample.js
```diff
@@ -92,6 +93,11 @@ export const SignUpExample = () => {
       const adjustedRule = { ...rule };
       const valid = adjustedRule.regexp.test(values.password);
       adjustedRule.valid = valid;
+      if (valid !== true) {
+        setBackgroundError('#FC61613D');
```
taysea, line 97: Could we reference the theme value here?

aries-site/src/examples/templates/forms/SignUpExample.js
```diff
@@ -92,6 +93,11 @@ export const SignUpExample = () => {
...
+      } else {
+        setBackgroundError('white');
```
taysea, line 99: Should this and the default backgroundError be `undefined`? I assume when I switch to dark mode, we wouldn't want white to be there.

aries-site/src/examples/templates/forms/SignUpExample.js
taysea: Also, can you also summarize why the grommet validation isn't applying properly for the background color again? I know we discussed, but I forgot the details and it would be good to have documented in case it could be a future grommet unhancement.

aries-site/src/examples/templates/forms/SignUpExample.js
britt6612: yes thanks for catching!

halocline:
Can we talk through the intent of this PR next week? I could use some background as to why it is needed.

britt6612:
> Can we talk through the intent of this PR next week? I could use some background as to why it is needed.

yes of course! It wasn’t necessary needed but I wanted to match the designs with showing the red error color!
We. can chat Monday 😄

aries-site/src/examples/templates/forms/SignUpExample.js
```diff
@@ -81,17 +82,28 @@ const FormContainer = ({ ...rest }) => {
...
+  //  then we can change background color to indicate there is no error.
 
   const onChange = values => {
     setFormValues(values);
     const adjustedPasswordRules = passwordRules.map(rule => {
       const adjustedRule = { ...rule };
       const valid = adjustedRule.regexp.test(values.password);
       adjustedRule.valid = valid;
+      if (valid !== true) {
+        setBackgroundError(themeMode === 'dark' ? '#C54E4B5C' : '#FC61613D');
```
taysea, line 103: Do we even need to call for themeMode anymore? Could we just set it to `validation-critical` and grommet would take care of the dark or light mode?

aries-site/src/examples/templates/forms/SimpleSignUpExample.js
```diff
@@ -191,6 +204,7 @@ export const SimpleSignUpExample = () => {
               }
             >
               <TextInput
+                style={{ backgroundColor: backgroundError }}
```
taysea, line 207: Soft opinion, but would we want to wrap this textinput in a box and set the background on there as a prop? using the `style` prop feels a little anti grommety

aries-site/src/examples/templates/forms/SignUpExample.js
britt6612: no this will change

aries-site/src/examples/templates/forms/SimpleSignUpExample.js
britt6612: yeah this will also change after talking with Greg we dont want validation on blur so the whole background will be taken out

aries-site/src/examples/templates/forms/SignUpExample.js
```diff
@@ -180,9 +215,7 @@ export const SignUpExample = () => {
                             <FormCheckmark size="small" />
                           </Box>
                         ) : (
-                          <Box alignSelf="center">
-                            <FormClose size="small" />
-                          </Box>
+                          <></>
```
taysea, line 218: instead of doing an empty tag, can we change line 213 to be:
```
formValues.password && rule.valid &&
```
then we don't have to do a `?` and `:`

aries-site/src/examples/templates/forms/SimpleSignUpExample.js
```diff
@@ -179,9 +214,7 @@ export const SimpleSignUpExample = () => {
                             <FormCheckmark size="small" />
                           </Box>
                         ) : (
-                          <Box alignSelf="center">
-                            <FormClose size="small" />
-                          </Box>
+                          <></>
```
taysea, line 217: Same as other comment regarding swapping `?` `:` to `&&` because we don't need empty tag

aries-site/src/examples/templates/forms/SignUpExample.js
```diff
@@ -12,22 +13,55 @@ import {
...
+Error.propTypes = {
+  children: PropTypes.object,
+};
+
+const passwordRequirements = [
+  {
+    regexp: new RegExp('(?=.*?[A-Z])(?=.*?[a-z])(?=.*?[#?!@$ %^&*-]).{8,}'),
+    message: (
+      <Error background="background-front">
+        Password requirements not meet.
```
halocline, line 38:
```suggestion
        Password requirements not met.
```

aries-site/src/examples/templates/forms/SimpleSignUpExample.js
```diff
@@ -12,7 +13,34 @@ import {
...
+Error.propTypes = {
+  children: PropTypes.object,
+};
+
+const passwordRequirements = [
+  {
+    regexp: new RegExp('(?=.*?[A-Z])(?=.*?[a-z])(?=.*?[#?!@$ %^&*-]).{8,}'),
+    message: (
+      <Error background="background-front">
+        Password requirements not meet.
```
halocline, line 38: met

PR #1003, opened 2020-07-30
Update List Copy

### LINKED ISSUE #964: Add List guidance
Pulling from comments on https://github.com/hpe-design/design-system/pull/956

- [ ] In each variant, elaborate on each unique element (image, icon, status, etc.). For example, we should explain how image should be used to assist scan-ability -- specific to this example, the image should be tied to the data present.
- [ ] What happens with longer labels? (at various widths), does text truncate, wrap?
- [ ] What happens with subtitle/descriptions? (at various widths), does text truncate, wrap?
- [ ] Max-width of lists used to ensure readability/scan-ability. `xxlarge` is max-width for page content, so should list follow this?
- [ ] Accessibility: List `action` button and `onClick` should not be used together. Either the entire list item has an `onClick` or there is an action button on the far right of the list item.
- [ ] What is the anatomy of a list item?
- [ ] When do I use List vs Table?
- [ ] Hover state: If the entire list item is clickable, the background of a list item receives the active background color and the cursor changes to a pointer hand. Otherwise, there is no hover state when a list item is not interactive. If a list item contains an action button, then that button respects its natural hover behavior.

Also, we could likely improve the mock data to feel more realistic. If @cherih or others have thoughts on what the data should be, feel free to comment that here.

taysea:
#### What does this PR do?

Updates List copy based on conclusions here: https://github.com/hpe-design/design-system/issues/964#issuecomment-666677416

aries-site/src/pages/templates/lists.mdx

e2e

Closes #964 

No.

Yes.

### DIFF
--- a/aries-site/src/pages/templates/lists.mdx
+++ b/aries-site/src/pages/templates/lists.mdx
@@ -65,7 +65,7 @@ Use hover states to indicate which kind of selection is being used. If the entir
 ### Action Menu in List Item
 
 - If space is limited, then don’t use an Action Menu in a List item (repetition of the Action Menu icon uses a lot of space)
-- If items can be selected using only the List contents, then use an Action Menu in a List item. Otherwise, go to a detail view to see additional information for the List item and use an Action Menu there.
+- If more context is needed to perform an action, the entire list row should be clickable and should route the user to a detail page for the List item. On this detail page, an Action Menu can be provided to allow the user to engage with the data.
 - If an action is simple and the results can be seen in the List contents (e.g. enable/disable), then use an Action Menu in a List item
 - If multi-select is required, then don’t use the Action Menu in the List item.
 


aries-site/src/pages/templates/lists.mdx
```diff
@@ -65,7 +65,7 @@ Use hover states to indicate which kind of selection is being used. If the entir
 ### Action Menu in List Item
 
 - If space is limited, then don’t use an Action Menu in a List item (repetition of the Action Menu icon uses a lot of space)
-- If items can be selected using only the List contents, then use an Action Menu in a List item. Otherwise, go to a detail view to see additional information for the List item and use an Action Menu there.
+- If more context is needed to perform an action, the entire list row should be clickable and should route the user to a detail page for the List item. On this detail page, an Action Menu can be provided to allow the user to engage with the data.
```
britt6612, line 68: Im not best at grammar but could it be 
route the user to a detailed page ? or is detail page the correct way ?

taysea: Good comment. I think it could go either way, I think the way I was using it was almost more "naming" this type of page.

britt6612: oh okay yeah then! this is good by me

Issue #1007, opened 2020-07-31
Enhance Wizard implementation


taysea:
Starting a list of behaviors/enhancements for wizard page:
- [x] Demonstrate validation per step
- [x] Scroll to top of step when someone advances
- [x] Discuss sticky wizard footer display on large screens (button feels very far from content)

gregfuruiye:
I think we only need to show what happens when the user clicks 'cancel'. What happens when they confirm cancel is up to the service team. It could vary depending on how they entered the wizard to begin with. However, if we wanted to demonstrate something, we could create a kind of dummy launch page that would allow them to start the wizard again.

click cancel --> cancel layer --> click to confirm cancel --> close the layer and the wizard and show some simple homepage with sidebar and header and one big card/button in main --> click the card/button 'Start Wizard'

taysea:
Gotcha, it feels like for simplicity of the code/purpose of keeping it focused on wizard then we should stop at the `cancel` layer. otherwise, lots of additional code will be living in the "show code" section of the site which may clutter the readability of the example.

PR #1023, opened 2020-08-04
Demonstrate validation within a wizard step

taysea:
#### What does this PR do?

Demonstrates how validation can be employed in a wizard. A wizard is built out as a single form where inputs are divided into various "steps". The active step is controlled via state. Before advancing to the next step, we want to validate the currently visible inputs. Grommet's formfield/form is not able to handle this kind of validation out of the box, so some custom logic and functions were needed to achieve this approach.

aries-site/src/examples/templates/wizard/WizardValidationExample.js

Try to advance through the "With validation" example. The first step will demonstrate validation errors when not completed properly.

As noted above.

Related to #1007 

Yes.

Yes, backwards compatible.


aries-site/src/examples/templates/wizard/WizardValidationExample.js
```diff
@@ -0,0 +1,346 @@
...
+        >
+          <TextInput
+            placeholder="jane.smith@hpe.com"
+            id="text-input-validation"
+            name="text-input-validation"
+            type="email"
+          />
+        </FormField>
+        <FormField
+          help="Click on the ‘eye’ icon for the description text field to hide."
```
halocline, line 71: I am not sure I understand this message.

aries-site/src/examples/templates/wizard/WizardValidationExample.js
```diff
@@ -0,0 +1,346 @@
...
+          htmlFor="radio-button-group-validation"
+          label="RadioButtonGroup"
+          name="radio-button-group-validation"
+        >
+          <RadioButtonGroup
```
halocline, line 76: For some reason, in full-screen mode, I am not able to make a selection for the radiobuttongroup input. May be unrelated to this PR, but noting for investigation.

aries-site/src/examples/templates/wizard/WizardValidationExample.js
```diff
@@ -0,0 +1,346 @@
...
+            id="radio-button-group-validation"
+            name="radio-button-group-validation"
+            options={['Radio button 1', 'Radio button 2']}
+          />
+        </FormField>
+      </Box>
+      <Button
```
halocline, line 83: Feeling like we need some treatment on or near the next button letting the user know there is an error (Like the error message @britt6612 has been working on). Or maybe not, if we assuming the error will be always visible in a wizard context? Can we make that assumption? Probably not on mobile?

Thinking aloud here.

aries-site/src/examples/templates/wizard/WizardValidationExample.js
```diff
@@ -0,0 +1,346 @@
...
+        label="Next"
+        icon={<FormNextLink />}
+        primary
+        reverse
+        onClick={() => {
+          // check for errors
+          const validation = stepOneValidate(formValues);
+          // advance to next step if successful
+          if (validation.isValid) setActive(active + 1);
+          // otherwise, display error
```
halocline, line 94:
```suggestion
          // otherwise, display error and wizard will not advance to next step
```

aries-site/src/examples/templates/wizard/WizardValidationExample.js
```diff
@@ -0,0 +1,346 @@
...
+    description: 'Provide a summary of what was accomplished or configured. ',
+    inputs: (active, setActive) => (
+      <StepThree active={active} setActive={setActive} />
+    ),
+    title: 'Finish title',
+  },
+];
+
+export const WizardValidationExample = () => {
+  const [active, setActive] = useState(1);
```
halocline, line 210: Soft opinion here.

Would `activeStep` or `activeIndex` help for clarity?

aries-site/src/examples/templates/wizard/WizardValidationExample.js
```diff
@@ -0,0 +1,346 @@
...
+                {steps[active - 1].description}
+              </Text>
+            </Box>
+            <Form
+              value={formValues}
+              onChange={nextValue => setFormValues(nextValue)}
+              onSubmit={({ value }) => console.log(value)}
+            >
+              {/* Index an array starts at 0 */}
+              {steps[active - 1].inputs(active, setActive, formValues)}
```
halocline, line 251: If we used the term `activeIndex` we could initialize state at 0 (as opposed to 1). Then we wouldn't need to worry about all of the `active - 1`. Could eliminate the need for code comments, unnecessary operations, and probably less prone to bugs.

halocline: Would require changes throughout each reference to active, but it feels cleaner.

aries-site/src/examples/templates/wizard/WizardValidationExample.js
taysea: I like `activeStep` let me make that update for this file here. I will update the other wizard files in a separate PR.

aries-site/src/examples/templates/wizard/WizardValidationExample.js
taysea: There would still need to me comments and calculations in the areas where we are using the active state to display the step number. in these places, we would do `active + 1` instead.

taysea: I am open to either direction, but feel like some calculation/comments will be necessary in either direction.

aries-site/src/examples/templates/wizard/WizardValidationExample.js
taysea: Good note, I think this is separate from this PR but we should file as an issue.

aries-site/src/examples/templates/wizard/WizardValidationExample.js
taysea: This is coming from the designs, but I'm reading it now and agree it's unclear what is supposed to happen here. I am not sure any help text is necessary for this field.

aries-site/src/examples/templates/wizard/WizardValidationExample.js
taysea: That is a good point. I think adding something similar to what Brittany has implemented would be effective.

taysea: Implemented.

aries-site/src/examples/templates/wizard/WizardValidationExample.js
taysea: For readability, I have changed `active` to `activeIndex` with default state of 0. This will be more aligned with typical array conventions/expectations. However, I've also introduced `activeStep` which is used for displaying the numeric value of the step in the UI. This is automatically updated by a useEffect whenever activeIndex updates.

aries-site/src/examples/templates/wizard/WizardValidationExample.js
taysea: Noticing this is happening even in just the radiobuttongroup page when you open to full screen. Tracking this issue here: #1031

aries-site/src/pages/templates/wizard.mdx
```diff
@@ -28,7 +29,7 @@ It is important to give the user an indication of where they are within the form
 
 ### Validation
 
-Validation should occur within each step of the wizard. This can happen `onBlur` of each input or when the user clicks the button that would advance them to the next step. If there are any errors with the inputs in the current step, the user should not be able to advance.
+Validation should occur within each step of the wizard. If there are any errors with the inputs in the current step, the user should not be able to advance and errors should be displayed. Error display should follow conventions established by [Form](/templates/forms).
```
halocline, line 32: Wonder if we should reference and provide a link to the example below?

aries-site/src/pages/templates/wizard.mdx
```diff
@@ -85,3 +86,16 @@ When additional guidance is necessary, use a two-column wizard. Click "See Fulls
 >
   <TwoColumnWizardExample />
 </Example>
+
+### With validation
+
+Validation should occur on each step. If a user has not properly completed a given step, they should not be able to advance to subsequent steps.
+
+<Example
+  code="https://raw.githubusercontent.com/hpe-design/design-system/master/aries-site/src/examples/templates/wizard/TwoColumnWizardExample.js"
```
halocline, line 95:
```suggestion
  code="https://raw.githubusercontent.com/hpe-design/design-system/master/aries-site/src/examples/templates/wizard/WizardValidationExample.js"
```

aries-site/src/pages/templates/wizard.mdx
```diff
@@ -85,3 +86,16 @@ When additional guidance is necessary, use a two-column wizard. Click "See Fulls
...
+  figma="https://www.figma.com/file/GEhgYzx3b8MejBjg9gAxyA/Wizarding?node-id=71%3A406"
```
halocline, line 96: Should we exclude the Figma link here since it is not pointing to any specific validation example?

aries-site/src/examples/templates/wizard/WizardValidationExample.js
```diff
@@ -0,0 +1,391 @@
...
+const StepOne = ({ activeIndex, setActiveIndex, formValues }) => {
+  const [error, setError] = useState({
+    email: '',
+    radiobuttongroup: '',
+    isValid: true,
+  });
+
+  return (
+    <>
+      <Box margin={{ bottom: 'large' }}>
```
halocline, line 57: Does the gap between the field and error seems too large?
Screen Shot 2020-08-06 at 11 02 40 AM

If so, a few potential solutions:
- Move the error to appear within the Box
- Make this `bottom: 'medium'` and give the Button a `margin={{top: 'medium'}}`
- probably some others

aries-site/src/examples/templates/wizard/WizardValidationExample.js
```diff
@@ -0,0 +1,391 @@
...
+};
+
+const Error = ({ children, ...rest }) => {
+  return (
+    <Box
+      background="validation-critical"
+      margin={{ bottom: 'medium' }}
+      pad="small"
+      round="4px"
+    >
```
halocline, line 378: I think @britt6612 ended up adding a little animation to the error reveal, so that the error doesn't feel like a slap in the face. Check it out and see if you like it or not.

```suggestion
    <Box
      animation="fadeIn"
      background="validation-critical"
      margin={{ bottom: 'medium' }}
      pad="small"
      round="4px"
    >
```

aries-site/src/pages/templates/wizard.mdx
taysea: Yes, I think you're right. I will remove it.

aries-site/src/pages/templates/wizard.mdx
taysea: Yes I think that would be helpful

taysea: Implemented.

aries-site/src/examples/templates/wizard/WizardValidationExample.js
taysea: I would agree. I've tightened the spacing to better match the example here: https://www.figma.com/file/3fkwBelW5UsCbfwdDCJkT8/HPE-Form-Templates?node-id=2668%3A3547

aries-site/src/examples/templates/wizard/WizardValidationExample.js
taysea: Implemented this so the behaviors match.

PR #1026, opened 2020-08-05
Fix scrolling for wizard demo

taysea:
#### What does this PR do?

When a user advances to the next step of a wizard, the step is scrolled to the top to create a more enjoyable user experience. Prior, the scroll position would persist, so you'd be advanced to the next step but sometimes land in the middle.

Due to the fact that we are demonstrating examples inline (but in reality, the wizard would take the whole screen height), I've had to add an implementation that works for the site (needed a little extra logic) and added a comment with the proper code to use for application implementation. The comments aim to direct the user what to keep vs discard if they were to copy and paste.

aries-site/src/examples/templates/wizard/SimpleWizardExample.js

Click through to next step at various scrolling positions.

Same as above.

Related to #1007 

No.

Yes, backwards compatible.


aries-site/src/examples/templates/wizard/StickyHeaderFooterExample.js
```diff
@@ -140,14 +140,28 @@ export const StickyHeaderFooterExample = () => {
   // controls state of cancel layer
   const [open, setOpen] = useState(false);
   const size = useContext(ResponsiveContext);
+
+  const wizardRef = React.useRef();
+
```
britt6612, line 145: can we add some comments here like above ?

aries-site/src/examples/templates/wizard/TwoColumnWizardExample.js
```diff
@@ -144,9 +144,27 @@ export const TwoColumnWizardExample = () => {
   const [open, setOpen] = useState(false);
   const size = useContext(ResponsiveContext);
 
+  const wizardRef = React.useRef();
```
britt6612, line 147: This is great! I think the comments make sense as well so the users know they will not need this

aries-site/src/examples/templates/wizard/StickyHeaderFooterExample.js
taysea: Yes, I can add some comments.

aries-site/src/examples/templates/wizard/SimpleWizardExample.js
```diff
@@ -173,9 +173,29 @@ export const SimpleWizardExample = () => {
...
+    const layerContainer = document.querySelector('#layer-wrapper');
+    if (layerContainer) {
+      wizardRef.current.scrollIntoView();
+    } else {
+      const container = wizardRef.current.parentNode;
+      container.scrollTop = -container.getBoundingClientRect().top;
+    }
+
+    // USE THIS CODE FOR APPLICATION:
+    // wizardRef.current.scrollIntoView();
```
halocline, line 193:
```suggestion
    // FOR SCROLL USE IN APPLICATION: Uncomment line below.
    // wizardRef.current.scrollIntoView();
    
    // MODIFIED SCROLL FOR USE IN DEMO:
    // This block is purely to ensure proper scrolling for the inline
    // site demo. Use line above and remove this block for your app.
    const layerContainer = document.querySelector('#layer-wrapper');
    if (layerContainer) {
      wizardRef.current.scrollIntoView();
    } else {
      const container = wizardRef.current.parentNode;
      container.scrollTop = -container.getBoundingClientRect().top;
    }
```

aries-site/src/layouts/content/Example/Example.js
```diff
@@ -170,7 +170,12 @@ export const Example = ({
                 {...rest}
               >
                 {screenContainer || showResponsiveControls ? (
-                  <Box width={screen === screens.mobile ? 'medium' : '100%'}>
+                  <Box
+                    // for purpose of inline scroll behavior on templates
+                    // this id is needed to reference the scroll parent
```
halocline, line 175: Great use of a comment

aries-site/src/examples/templates/wizard/SimpleWizardExample.js
halocline: Apply to each example.

halocline:
One other thing to play with is potentially adding some slight animation for when step screens are swapped out. The fragment for each step could be replaced with a `<Box animation="fadeIn" >`.

This could make the changes a bit less abrupt.

PR #1034, opened 2020-08-07
Removed local icons to use grommet-icons instead

ShimiSun:
Removed the usage of the local Sun/Moon icons and use grommet-icons instead.


aries-site/src/components/content/ThemeModeToggle.js
```diff
@@ -1,7 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import { Button } from 'grommet';
-import { IconMoon, IconSun } from '../icons';
+import { Moon as IconMoon, Sun as IconSun } from 'grommet-icons';
```
taysea, line 4: Do we want to leave with no alias and just change the lines below to reference `Moon` and `Sun`?

ShimiSun: Sure, done.

PR #1155, opened 2020-09-30
Utilize page object model for search

taysea:
#### What does this PR do?
Utilizes testcafe's page object model structure to simplify search tests. Removes lots of repeated code so that functions can be more reusable/scalable.

aries-site/src/tests/navigation/search.js

Passes all tests still.


halocline (approved):
I like the approach and the enhancements!

Issue #1147, opened 2020-09-24
Table - Update row selection examples to use new onSelect prop


halocline:
onSelect was recently added (9/17/2020). Examples on the Table page should use this functionality.

PR #1158, opened 2020-09-30
replaced multi-select example with onSelect functionality

halocline:
[Preview](https://deploy-preview-1158--keen-mayer-a86c8b.netlify.app/components/table#selecting-multiple-records-and-batch-actions)

#### What does this PR do?

Simplifies DataTable multi-select example by applying Grommet enhancements.

DataTable multi-select example

Manual on design-system site

Manual on design-system site

Grommet added select and onSelect props to DataTable.

https://github.com/grommet/hpe-design-system/issues/1147

Perhaps.

Backwards.

### DIFF
--- a/aries-site/src/examples/components/table/TableMultiSelectExample.js
+++ b/aries-site/src/examples/components/table/TableMultiSelectExample.js
@@ -4,7 +4,6 @@ import styled from 'styled-components';
 import {
   Box,
   Button,
-  CheckBox,
   DataTable,
   Heading,
   Menu,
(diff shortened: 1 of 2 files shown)


taysea (commented):
All looks good from usage of props, but noticing the upper Checkbox isn't aligning with the rest. I think this is due to how DataTable header button is filling available space, but the Checkbox itself is aligning start.

Let's handle the theme stuff separately. Implementation of this looks good to me.

taysea (approved):
Since we will handle theme stuff separately, this LGTM

PR #1160, opened 2020-09-30
update reference to hpe theme from core

halocline:
#### What does this PR do?

Closes https://github.com/grommet/hpe-design-system/issues/1161

Points aries-core to grommet-theme-hpe stable instead of next-stable. This will fix a conflict in our monorepo node_modules.


taysea (approved):
LGTM. Good catch

PR #1164, opened 2020-10-01
DataTable theme properties and theme file clean up

halocline:
[Preview](https://deploy-preview-1164--keen-mayer-a86c8b.netlify.app/components/table)

#### What does this PR do?

DataTable theme properties and theme file clean up

aries.js

Manual, Table page (dark and light modes)

Manual, Table page (dark and light modes)

https://github.com/grommet/hpe-design-system/issues/1159

No.

Backwards

### DIFF
--- a/aries-site/src/themes/aries.js
+++ b/aries-site/src/themes/aries.js
@@ -1,5 +1,6 @@
 import { hpe } from 'grommet-theme-hpe';
 import { deepMerge } from 'grommet/utils';
+import { Unsorted } from 'grommet-icons';
 
 export const aries = deepMerge(hpe, {
   defaultMode: 'dark',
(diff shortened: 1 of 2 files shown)


aries-site/src/themes/aries.js
```diff
@@ -41,16 +29,25 @@ export const aries = deepMerge(hpe, {
...
     header: {
-      pad: { horizontal: 'small', vertical: 'xsmall' },
+      // background: undefined,
+      border: { side: 'bottom' },
       color: 'text-strong',
       font: {
         weight: 'bold',
+        // size: undefined,
       },
+      // gap: 'small',
```
taysea, line 39: Looking back at designs, I think we'll actually want this to be `gap="none"` so that once we add in any resizing, there won't be any space between the button and the resizer

taysea (commented):
Everything looks great here. The only other thing I notice is once a header cell label wraps, then the other header cell buttons don't fill the full height one hover. I'm wondering if there's a way to theme it so that those buttons always take 100% height, but i think we can take that in a separate PR if it seems like it needs more investigation.

aries-site/src/themes/aries.js
halocline: Added.

halocline:
> Everything looks great here. The only other thing I notice is once a header cell label wraps, then the other header cell buttons don't fill the full height one hover. I'm wondering if there's a way to theme it so that those buttons always take 100% height, but i think we can take that in a separate PR if it seems like it needs more investigation.

I took a stab at filling the height which gets us close. Let me know what you think. If we aren't satisfied, I think we might need to open an issue for investigation (there are other areas where the theme capabilities could be enhanced as well).

taysea:
> > Everything looks great here. The only other thing I notice is once a header cell label wraps, then the other header cell buttons don't fill the full height one hover. I'm wondering if there's a way to theme it so that those buttons always take 100% height, but i think we can take that in a separate PR if it seems like it needs more investigation.
> 
> I took a stab at filling the height which gets us close. Let me know what you think. If we aren't satisfied, I think we might need to open an issue for investigation (there are other areas where the theme capabilities could be enhanced as well).

I think this is great for now. It solves the issue, and we can subsequently look into if it'd be something we'd want to be able to control on grommet or not.

taysea (approved):
LGTM. Great work.

PR #1165, opened 2020-10-01
Percy testing

### LINKED ISSUE #1162: Enhance Example tests on site to avoid visual regressions

taysea:
#### What does this PR do?

**1. Adds in new snapshot tests for the various Example layouts**

Currently, our various examples can be displayed in many ways. In some cases, like a single component, the component should be aligned in the center of the example. In other cases, we want to wrap the example in a mock browser. In template cases, we want to set the Example aspect ratio to something that better resembles the viewport of a laptop or mobile screen. 

Each time we update the Example component, there is risk that the desired layout variations could be affected and until now we have been manually checking the pages to make sure nothing was affected. Adding in snapshot tests would give us much more confidence that we are not seeing any visual regressions as we continue to iterate on the site.

The tests check both the inline display, full screen display, and various viewport (desktop, laptop, mobile) displays if applicable.
**2. Adds in CI configuration with Percy**

I have configured this repo with Percy so that on pull requests, the snapshots will be compared to those stored on `master` and any visual regressions will be flagged as part of the pull request checks.

**3. Adjustments to test scripts**

I adjusted the test scripts so in the CI setting, all the e2e tests will run, including the snapshot tests. However, locally, the snapshot tests won't run but all other e2e tests will. The decision for this set-up was to preserve the number of snapshots being taken since Percy's free tier only allows for 5,000 snapshots per month. Additionally, I've configured Percy to just take snapshots at a 1280px screen width. If we want to explore taking screen shots at more widths, we can but for the purposes of these tests and preserving available snapshots, it felt that a single screen width would be sufficient.

**Why Percy?**

I chose to use Percy because it had a simple integration with TestCafe, which we are already using for the rest of our e2e tests. It is also a reliable and popular visual regression testing service, so I felt it would be a nice option to explore for the site.

**To-do:**
- [x] Add Shimi and Matt as members on Percy project so they can review snapshots

**Test set-up**
I created a utils function called `takeResponsiveSnapshots` that handles resizing the window for mobile, taking the snapshots at both desktop and mobile width, and naming the files properly. Relying on Percy's built-in snapshot widths was not properly rendering the mobile site, and I believe this is due to how it takes a snapshot of the DOM and then renders that across various browsers. I experimented with enabling the percy flag of `enable javascript`, but this then broke the testcafe ability to interact with that page. A similar thing was noted here for Cypress + Percy users, and indicated the resolution I went with which was handling the responsive snapshots by resizing the window within the test. The only negative to this is that in Percy's UI, we can't toggle directly between the two widths, instead we have to scroll to the two separate snapshots. However, it is successful in our intention to capture the screen shots on both a desktop and mobile width.

aries-site/src/tests/examples/screen-container.js

Workflow completed successfully, all e2e tests still run and Percy snapshots appear in the percy project.

Closes #1162 

No.

Backwards compatible.


.github/workflows/testcafe.yml
```diff
@@ -21,3 +21,5 @@ jobs:
       - name: Build and run e2e tests on Chrome
         run: yarn test:ci
         working-directory: aries-site
+        env:
+          PERCY_TOKEN: ${{ secrets.PERCY_TOKEN }}
```
halocline, line 25: Should this have a comment mentioning token is defined in Percy's project settings?

halocline: Also, I am noticing that the account is still under your name. Do we want to get that transferred to an HPE held account so that it doesn't get stuck with you?

aries-site/src/examples/templates/persistent-navigation/MinimalSidebarExample.js
```diff
@@ -153,30 +154,28 @@ const NavButton = ({ active, icon, name, ...rest }) => {
         onBlur={() => setHover(false)}
         {...rest}
       />
-      {
-        /* Show tooltip on hover and focus states as a supplemental
+      {/* Show tooltip on hover and focus states as a supplemental
```
halocline, line 157: These changes are formatting only, correct? Can't spot any other diffs.

aries-site/src/tests/examples/screen-container.js
```diff
@@ -0,0 +1,27 @@
...
+});
+
+test(`should wrap content, content should fill partial width on 
+desktop`, async t => {
+  const desktopButton = Selector('button').withText('Desktop');
+  await t.click(desktopButton);
+  await takeResponsiveSnapshots(t, 'Screen Container on Desktop');
+});
+
+test('should not appear on mobile', async t => {
```
halocline, line 23: The name of this test confuses me. What should not appear on mobile? The mock browser wrapper? How about the content?

aries-site/src/tests/utils.js
```diff
@@ -60,3 +61,14 @@ export const tabToSearch = ClientFunction(() => {
   // 4 times to get there.
   return tabCount + 1;
 });
+
+export const takeResponsiveSnapshots = async (t, title) => {
```
halocline, line 65: Nice utility function.

halocline (commented):
No issues. Just a couple clarifying questions. Looking good.

aries-site/src/examples/templates/persistent-navigation/MinimalSidebarExample.js
taysea: Yes, just a linter fix.

.github/workflows/testcafe.yml
taysea: We can definitely do that. Which email do you think would be appropriate?

aries-site/src/tests/examples/screen-container.js
taysea: Good question, since the fixture is called `Screen Container`, then it will print out in the test runner as:
```
Screen Container
   should not appear on mobile
   name of other test
   name of another thest
```
but I can see how just reading line 23 is ambiguous. How about we change it to "should not display mock browser for mobile display"?

.github/workflows/testcafe.yml
taysea: Added Mike as an admin. I would add the hpe design system email, but it does not have an associated github account.

halocline: > Should this have a comment mentioning token is defined in Percy's project settings?

Do you think a comment specifying where the token is defined is useful? (e.g. "Token can be found in Percy's project settings") --- Or, is that already well understood?

halocline (commented):
One small question.

halocline (approved):

taysea:
Closing in favor of using Applitools: https://github.com/grommet/hpe-design-system/pull/1228

PR #1175, opened 2020-10-06
Update reset filters behavior

taysea:
[Preview](https://deploy-preview-1175--keen-mayer-a86c8b.netlify.app/templates/filtering)

#### What does this PR do?
When a user clicks "Reset Filters" on a Filtering layer/drop, the layer/drop should remain open. Clicking "Apply Filters" or the Close icon should be the ways a layer can close. Greg pointed out that the behavior on the site did not match his design expectations.

Any file.

Clicking the "Reset Filters" button leaves the layer/drops open.

Same as above.

N/A


halocline (approved):
Looks good and agree that this is a nice improvement.

PR #1183, opened 2020-10-12
Update alignment of checkbox in table header

### LINKED ISSUE #1181: Fix padding of checkbox on Table header to align with checkboxes below it

taysea:
#### What does this PR do?
When `onSelect` is added as a prop for DataTable, a checkbox appears in the header cell. However, the alignment was off with the checkboxes in the body cells because we have removed padding from header cells. This adjusts the alignment by identifying the `label` which surrounds the checkbox and justifying it to the center.

aries-site/src/themes/aries.js

Locally looking at DataTable w select.

Same as above.

Closes #1181 

No.

Yes.


aries-site/src/themes/aries.js
```diff
@@ -89,6 +89,10 @@ export const aries = deepMerge(hpe, {
           height: 100%;
           justify-content: center;
         }
+        // align onSelect checkbox to center of header cell
```
ericsoderberghp, line 92: Why isn't this in the HPE theme?

taysea: we will be moving all of this over to the HPE theme soon as the dataTable theming is fluctuating less now. As designs were getting finalized we wanted to iron items out first before placing on theme branch.

halocline (commented):
Unrelated to this work, but I am noticing that with the update of the yarn.lock we have visual regressions related to Menu (padding removed, disabled styling no longer applying, etc.):
Screen Shot 2020-10-12 at 12 13 32 PM

taysea:
> Unrelated to this work, but I am noticing that with the update of the yarn.lock we have visual regressions related to Menu (padding removed, disabled styling no longer applying, etc.):
> Screen Shot 2020-10-12 at 12 13 32 PM

uh oh. good catch there. that is definitely something to look at ..

PR #1190, opened 2020-10-14
Update header cells to match designs

### LINKED ISSUE #1184: Aligning table header abbreviations to Table Figma file

taysea:
Preview
#### What does this PR do?
Designs called for units to be `text` and size `xsmall` in table header cells. This updates this and the layout of the text so that when header cells wrap there is better alignment.

aries-site/src/examples/components/table/TableExample.js

Checked local deploy.

Look at preview.

Closes #1184 

Backwards compatible.


ericsoderberghp:
I'm wondering if we should enhance grommet DataTable `columns` to build this in more. I'm thinking including units in a header cell is typical enough to warrant it. This would allow the styling to be built into the theme rather than having callers do this for each table they have.

taysea:
> I'm wondering if we should enhance grommet DataTable `columns` to build this in more. I'm thinking including units in a header cell is typical enough to warrant it. This would allow the styling to be built into the theme rather than having callers do this for each table they have.

I like the thought of that. I'll file an issue on grommet for this.

halocline:
> > I'm wondering if we should enhance grommet DataTable `columns` to build this in more. I'm thinking including units in a header cell is typical enough to warrant it. This would allow the styling to be built into the theme rather than having callers do this for each table they have.
> 
> I like the thought of that. I'll file an issue on grommet for this.

https://github.com/grommet/grommet/issues/4610

halocline (commented):
No concerns with the work here. Like the Grommet issue filed. Should we have a subsequent issue created for "Refactor Design System Table Header Units" once https://github.com/grommet/grommet/issues/4610 is closed so that we can circle back here? Or is the plan to put this on hold and do the Grommet work prior?

taysea:
> No concerns with the work here. Like the Grommet issue filed. Should we have a subsequent issue created for "Refactor Design System Table Header Units" once [grommet/grommet#4610](https://github.com/grommet/grommet/issues/4610) is closed so that we can circle back here? Or is the plan to put this on hold and do the Grommet work prior?

I'd prefer to create a subsequent issue and update once the grommet fix is merged. Don't think it warrants putting this on hold.

PR #1198, opened 2020-10-15
Add Applitools integration

taysea:
#### What does this PR do?
**1. Adds in new snapshot tests for the various Example layouts**

Currently, our various examples can be displayed in many ways. In some cases, like a single component, the component should be aligned in the center of the example. In other cases, we want to wrap the example in a mock browser. In template cases, we want to set the Example aspect ratio to something that better resembles the viewport of a laptop or mobile screen. 

Each time we update the Example component, there is risk that the desired layout variations could be affected and until now we have been manually checking the pages to make sure nothing was affected. Adding in snapshot tests would give us much more confidence that we are not seeing any visual regressions as we continue to iterate on the site.

The tests check both the inline display, full screen display, and various viewport (desktop, laptop, mobile) displays if applicable.

**2. Adds in CI configuration with Applitools**

I have configured this repo with Applitools so that on pull requests, the snapshots will be compared to those stored on `master` and any visual regressions will be flagged as part of the pull request checks.

**3. Adjustments to test scripts**

I adjusted the test scripts so in the CI setting, all the e2e tests will run, including the snapshot tests. However, locally, the snapshot tests won't run but all other e2e tests will. The decision for this set-up was to preserve the number of snapshots being taken.

**Why Applitools?**

I chose to use Applitools because it had a simple integration with TestCafe, which we are already using for the rest of our e2e tests. It is also a reliable and popular visual regression testing service, so I felt it would be a nice option to explore for the site. It offers the ability to do cross browser screenshots, including IE11 which we typically are unable to test locally on our machines.

**To-do:**
- [ ] Add Shimi and Matt as members on Percy project so they can review snapshots

**Test set-up**
I created a utils function called `startResponsiveSnapshots` that handles resizing the window for mobile or desktop and naming the files properly. Relying on Applitool's built-in snapshot widths was not properly rendering the mobile site, and I believe this is due to how it takes a snapshot of the DOM and then renders that across various browsers. A similar thing was noted here for Testcafe + Percy users, and indicated the resolution I went with which was handling the responsive snapshots by resizing the window within the test.

Applitools groups tests based on their notion of `Eyes` -- each test requires that eyes `open`, take snapshots, `close`, and `waitForResults`.


taysea:
Closing bc https://github.com/grommet/hpe-design-system/pull/1228 provides working implementation

PR #1210, opened 2020-10-22
Update slack invite link

taysea:
#### What does this PR do?
The previous url is no longer valid/being used. This updates to the latest.


halocline (approved):

ShimiSun (approved):
Good catch!

PR #1216, opened 2020-10-29
Remove redundant Error messages

taysea:
#### What does this PR do?

Removing redundant uses of custom `Error` component that are now being handled by the theme. I have also removed the local theme changes now that they have merged into grommet-theme-hpe.

Any file.

Tested locally.

Check deploy on changed files. Only one `CircleAlert` icon should appear on the error state.

No.

Yes, backwards compatible.

### DIFF
--- a/aries-site/src/themes/aries.js
+++ b/aries-site/src/themes/aries.js
@@ -1,23 +1,12 @@
 import { hpe } from 'grommet-theme-hpe';
 import { deepMerge } from 'grommet/utils';
-import { CircleAlert, Unsorted } from 'grommet-icons';
+import { Unsorted } from 'grommet-icons';
 
 export const aries = deepMerge(hpe, {
   defaultMode: 'dark',
   // To be stripped out once theme changes are made in grommet-theme-hpe
-
   // keeping file for use as playground for future theme adjusments that need
   // to be quickly tested
-
-  // to be removed once merged into grommet-theme-hpe
-  formField: {
-    error: {
-      icon: <CircleAlert size="small" style={{ marginTop: '4px' }} />,
-      container: {
-        gap: 'xsmall',
-      },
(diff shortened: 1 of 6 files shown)


aries-site/src/themes/aries.js
```diff
@@ -1,23 +1,12 @@
 import { hpe } from 'grommet-theme-hpe';
 import { deepMerge } from 'grommet/utils';
-import { CircleAlert, Unsorted } from 'grommet-icons';
```
halocline, line 3: 👍

halocline (commented):
One more: https://github.com/grommet/hpe-design-system/blob/afe14ef6ca23662d7bf7d975221bb49396561463/aries-site/src/examples/templates/wizard/WizardValidationExample.js#L377

Can we remover this one as well?

taysea:
> One more:
> 
> https://github.com/grommet/hpe-design-system/blob/afe14ef6ca23662d7bf7d975221bb49396561463/aries-site/src/examples/templates/wizard/WizardValidationExample.js#L377
> 
> Can we remover this one as well?

I believe this one is a custom step level error message, in which case we still need it because it's not supported out of the box by grommet. It appears at the bottom of a step if there is an issue with an input.

halocline:
> > One more:
> > https://github.com/grommet/hpe-design-system/blob/afe14ef6ca23662d7bf7d975221bb49396561463/aries-site/src/examples/templates/wizard/WizardValidationExample.js#L377
> > 
> > Can we remover this one as well?
> 
> I believe this one is a custom step level error message, in which case we still need it because it's not supported out of the box by grommet. It appears at the bottom of a step if there is an issue with an input.

Ahh, gotcha. That makes sense.

halocline (approved):

PR #1255, opened 2020-11-12
Temp show applitools logs

taysea:
#### What does this PR do?
No review needed. Investigating how to increase rate that tests run so that testing suite takes less time on PR checks.


taysea:
No longer needed.

PR #1283, opened 2020-11-19
Contain Layers to Example region

### LINKED ISSUE #1115: Change examples to open Layers within the Examples
Currently, when opening Layers in the Examples, the Layer is positioned relative to the design system site. Instead, it would be better if it could be positioned within the Example. This would be especially helpful for the mobile example view.

A suggestion for how to do this would be to use https://v2.grommet.io/layer#target and set it to the appropriate Example viewport. The trick will be to not include this in the Example code. Or, perhaps we can include a helpful comment explaining what's going on. Perhaps something like `const { layerTarget } = useContext(ExampleContext);`?

taysea:
[Preview](https://deploy-preview-1283--keen-mayer-a86c8b.netlify.app/templates/filtering#filters-in-a-center-layer)

#### What does this PR do?
Allow Layers that are part of Examples to be contained to the dimensions of the example. This is done by creating refs within the internal Example component that can be consumed be the example as a prop. The ref is used as the target for the example Layer. This does require adding in a target to the example (which may not be desired in production code).  To mitigate confusion, I have placed a comment on every line where the ref or target is mentioned saying "[x] is for demo purposes only, remove in production". I have also cleaned up some console warnings re: prop types.

aries-site/src/layouts/content/Example/Example.js

Tested with Filter examples and updated all Filter examples.

Check Filter page on deploy. The Layers should open in the bounds of the example region.

Closes #1115 

No.

Yes, backwards compatible.


halocline (commented):
Are you seeing the Layer floating like in this video?
layer

taysea:
> Are you seeing the Layer floating like in this video?
> layer

Thanks for noting, yes I should've put that in the description. Carrying over from Slack, "One note, is when the layer is open and page scrolls, there’s a bit of a lag for the layer to move with the target. I chatted with eric and am going to address this as separate improvements on grommet."

halocline (commented):
From a code perspective, this all looks good. I think the comments speaking to the demo aspects work well.

I am just wondering about if we want to fix the Layer bounds portion prior to merging this? I think this is probably a product owner judgement. In my opinion, including this prior to having the Layer position fixed in place, reduces my confidence as an implementor -- it makes me feel like something is broken. Whereas, what we have currently isn't entirely precise in demonstrating behavior, but it does get the point across and is a solid experience re-enforcing confidence.

taysea:
> From a code perspective, this all looks good. I think the comments speaking to the demo aspects work well.
> 
> I am just wondering about if we want to fix the Layer bounds portion prior to merging this? I think this is probably a product owner judgement. In my opinion, including this prior to having the Layer position fixed in place, reduces my confidence as an implementor -- it makes me feel like something is broken. Whereas, what we have currently isn't entirely precise in demonstrating behavior, but it does get the point across and is a solid experience re-enforcing confidence.

I would agree with the "is something broken?" impression that you experienced is not something we want to extend to consumer of the site. I am fine waiting to merge this until the Layer behavior improves.

halocline (approved):

PR #1325, opened 2020-12-14
Update Dashboard example

### LINKED ISSUE #1285: Change heading color and icon color on cards on dashboard to text-strong
Derek called out a discrepancy on the DS site where the icons and text are set to default text color but should be text-strong instead.

Here is the slack convo: https://hpe.slack.com/archives/C04LMJWQT/p1605892510125700

DS site link of where the issue is found: https://design-system.hpe.design/templates/dashboards

Screenshot:

Screen Shot 2020-11-20 at 10 43 42 AM

taysea:
[Preview](https://deploy-preview-1325--keen-mayer-a86c8b.netlify.app/templates/dashboards)

#### What does this PR do?

Our dashboard example had not been refreshed since our initial pass of the site. One item we needed to update was text color usage, but I went and updated the cards/file references to leverage the global header work.

Using CCS as a reference: https://www.figma.com/file/qxdE6A2qswCVbW3R6Ti7KG/CCS-v0.09-New-Navigation?node-id=2629%3A18640

aries-site/src/examples/templates/dashboards/DashboardExample.js

Local.

Check deploy.

Closes #1285 

Backwards compatible via theme but provides new guidance around dashboard visuals.

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,6 +10,7 @@
   },
   "devDependencies": {
     "@hpe/project-scripts": "^1.0.8",
+    "eslint-plugin-prettier": "^3.3.0",
     "jest": "^25.1.0"
   },
   "workspaces": [
--- a/aries-site/src/examples/templates/dashboards/DashboardExample.js
+++ b/aries-site/src/examples/templates/dashboards/DashboardExample.js
@@ -1,291 +1,49 @@
-import React, { useContext } from 'react';
-import PropTypes from 'prop-types';
-import {
-  Avatar,
-  Header,
-  Heading,
-  Button,
-  Box,
-  Card,
-  CardBody,
-  CardFooter,
-  Grid,
-  Menu,
-  Nav,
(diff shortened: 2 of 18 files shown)


ericsoderberghp (approved):

package.json
```diff
@@ -10,6 +10,7 @@
   },
   "devDependencies": {
     "@hpe/project-scripts": "^1.0.8",
+    "eslint-plugin-prettier": "^3.3.0",
```
halocline, line 13: Background on why this was added?

taysea: This might be able to be removed -- when I was running the pre-commit linting, it wasn't finding eslint-plugin-prettier but it could be an issue with how I had things installed locally. if you're able to remove it and still pass all the pre-commit checks, then we should remove it and I will explore what's happening locally for me.

halocline: Okay. We can do some investigating.

Issue #1333, opened 2020-12-17
Update "required" text within Forms to be sentence case. 


MikeWalrath:
![image](https://user-images.githubusercontent.com/17712467/102552943-49e99080-407f-11eb-9e52-8be4d781d34f.png)

PR #1338, opened 2021-01-04
Make required form field messaging sentence cased

halocline:
#### What does this PR do?

Closes https://github.com/grommet/hpe-design-system/issues/1333
Primary - Consistent use of sentence casing on required error messaging.
Secondary - Opportunistic cleanup of examples.

Any changed example.

Manual site review on local dev + testing suite.

Submit form examples and verify displayed error messaging.

https://github.com/grommet/hpe-design-system/issues/1333

Backwards.

### DIFF
--- a/aries-site/src/examples/templates/forms/SimpleSignUpExample.js
+++ b/aries-site/src/examples/templates/forms/SimpleSignUpExample.js
@@ -126,6 +126,9 @@ export const SimpleSignUpExample = () => {
         <Form
           validate="blur"
           value={formValues}
+          messages={{
+            required: 'This is a required field.',
+          }}
           onChange={nextValue => onChange(nextValue)}
           onSubmit={({ value, touched }) => onSubmit({ value, touched })}
         >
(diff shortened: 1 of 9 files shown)


aries-site/src/examples/templates/forms/SimpleSignUpExample.js
```diff
@@ -126,6 +126,9 @@ export const SimpleSignUpExample = () => {
         <Form
           validate="blur"
           value={formValues}
+          messages={{
+            required: 'This is a required field.',
```
taysea, line 130: I am fine with this for now (as this is how grommet works), but a nice enhancement would be to be able to enforce this out of the box with grommet-theme-hpe or some other way. Thoughts? Seems like this opens up room for inconsistency if teams need to implement a custom required string on every form they implement.

halocline: Added https://github.com/grommet/hpe-design-system/issues/1339 to address for discussion and prioritization.

taysea (approved):
LGTM once circleci workflow fixes.

PR #1340, opened 2021-01-05
Provide example where anchor has weight=bold

### LINKED ISSUE #1329: Add Bold Anchor to site
CCS required bold Anchors which we added to Figma. Just need to add to site please :)

https://www.figma.com/file/I7PsiUmvr7OEJ6311rBUfg/HPE-Anchor-Component?node-id=0%3A1

taysea:
#### What does this PR do?
Adds an example where a weight="bold" is applied to anchor.

AnchorWeightExample.js

Checked locally.

Deploy.

Closes #1329 

No.

Yes, backwards compatible.

### DIFF
--- a/aries-site/src/pages/components/anchor.mdx
+++ b/aries-site/src/pages/components/anchor.mdx
@@ -5,6 +5,7 @@ import {
   AnchorDisabledExample,
   AnchorExternalExample,
   AnchorInlineExample,
+  AnchorWeightExample,
   SignUpExample,
 } from '../../examples';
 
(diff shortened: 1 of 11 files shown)


aries-site/src/pages/components/anchor.mdx
```diff
@@ -115,3 +116,14 @@ Used to indicate that an anchor cannot be interacted with.
 >
   <AnchorDisabledExample />
 </Example>
+
+### Anchor with weight
+
+A weight can be applied to Anchor in cases where extra emphasis is needed. This example demonstrates an Anchor with weight="bold".
```
halocline, line 122: Should we style this as code? I think we have an issue in the backlog to style inline code references, but we can do opportunistically when we come across these instances. Thoughts?

```suggestion
A weight can be applied to Anchor in cases where extra emphasis is needed. This example demonstrates an Anchor with `weight="bold"`.
```

halocline (approved):
Looks good. One question.

taysea:
> So, was this a breaking change on their end?

It does seem like something changed on their end that requires having to add `fully: true` to the tests. I didn't invest too much time stepping back through the versions to see which caused it but I'll send a note to their team bc they might have quicker insights.

PR #1392, opened 2021-01-28
Add Pagination Page

taysea:
[Preview](https://deploy-preview-1392--keen-mayer-a86c8b.netlify.app/templates/pagination)

#### What does this PR do?

Starting content on pagination so that it will be ready once [Pagination merges in grommet](https://github.com/grommet/grommet/pull/4646).

Still have more content to fill in, but have started the general scaffold and filled in some content.

Figma: https://www.figma.com/file/slRqNvhJub0QlgN46zSXoY/HPE-Pagination-Component?node-id=1%3A6


aries-site/src/data/structures/templates.js
```diff
@@ -275,4 +275,24 @@ export const templates = [
...
+          dark: '/templateImages/template-preview-persistent-invert.svg',
+        },
+        alt: 'HPE Design System Global Header',
+        fit: 'contain',
+      },
+    },
+    seoDescription: `Pagination divides content into separate pages in order to 
+    enhance navigation to specific items.`,
+    sections: [],
+    relatedContent: ['Table', 'Lists', 'Card'],
```
ShimiSun, line 296: Regardless of this PR, should we use a singular List instead of Lists to be consistent?

ShimiSun:
Any feedback about the Pagination API as you were consuming it? Any redundancy in the prop values? easy to use and control? In other words... Is the component good to go? :)

aries-site/src/data/structures/templates.js
taysea: That's a good point. I can file a clean up PR to use the component name instead of plural. Will we need to do a redirect for anyone who has potentially bookmarked the existing link?

taysea:
> Any feedback about the Pagination API as you were consuming it? Any redundancy in the prop values? easy to use and control? In other words... Is the component good to go? :)

The only thing I noticed was on Table (when you use `paginate` prop) and the table has horizontal scroll, the Pagination component doesn't stay fixed. I think we should pin the pagination component and just allow the table to scroll. No issues with the API though.

https://user-images.githubusercontent.com/12522275/111184785-8c334880-856e-11eb-9b63-52097b949c91.mov

aries-site/src/data/structures/templates.js
ShimiSun: Good question, I think that if the page is not routing to 404 we can change it as is (e.g the home page), but if the Lists page would be routed to a 404 we might want to have a fallback if it's not too complicated.

ShimiSun:
> > Any feedback about the Pagination API as you were consuming it? Any redundancy in the prop values? easy to use and control? In other words... Is the component good to go? :)
> 
> The only thing I noticed was on Table (when you use `paginate` prop) and the table has horizontal scroll, the Pagination component doesn't stay fixed. I think we should pin the pagination component and just allow the table to scroll. No issues with the API though.
> 

I like this idea, we can definitely improve this experience.

aries-site/src/examples/templates/pagination/PaginationListExample.js
```diff
@@ -0,0 +1,30 @@
...
+export const PaginationListExample = () => (
+  <Box overflow="auto" pad="xsmall">
+    <List
+      data={data}
+      action={(item, index) => (
+        <Menu
+          key={index}
+          icon={<More />}
+          hoverIndicator
+          items={[{ label: 'one' }]}
```
britt6612, line 22: Not sure if it matters that each label is 'one' could also use a for loop like above

aries-site/src/pages/templates/pagination.mdx
```diff
@@ -0,0 +1,76 @@
...
+  figma="https://www.figma.com/file/slRqNvhJub0QlgN46zSXoY/HPE-Pagination-Component?node-id=0%3A1"
+>
+  <PaginationTableExample />
+</Example>
+
+## When to use
+
+Use pagination when a user’s goal is to find a specific item rather than to peruse the full set of data. By breaking the data up into smaller, consistently sized groups, the user is more easily able to jump to specific sections of the data.
+
+If you expect the user will be scrolling and perusing large amounts of the data set, consider using [Infinite Scroll]().
```
britt6612, line 20: think you forgot the link

aries-site/src/examples/templates/pagination/PaginationListExample.js
taysea: Maybe we should change this to some action , cause this is the drop that appears if you click the "..." on the list item. So maybe something like "See Details" or something

aries-site/src/pages/templates/pagination.mdx
taysea: Good catch. We don't have a page yet for this so I'll add link to Grommet docs

britt6612:
Overall I think this is good to start. I think we should make a card for design if this is not already done. Also as I mentioned it feels like the pagination should be sticky and not move when the data is shorter but something to discuss.

aries-site/src/data/structures/templates.js
```diff
@@ -275,4 +275,24 @@ export const templates = [
...
   },
+  {
+    name: 'Pagination',
+    available: false,
+    description: `Pagination divides content into separate pages in order to 
+    enhance navigation to specific items.`,
+    preview: {
+      image: {
+        src: {
+          light: '/templateImages/template-preview-persistent.svg',
```
britt6612, line 286: we will need to add its own preview and alt text

taysea:
Closing in favor of https://github.com/grommet/hpe-design-system/pull/1487

PR #1410, opened 2021-02-05
Remove custom formfield labels, use theme

taysea:
#### What does this PR do?
Removing instances where we implemented custom required formfield. This styling is now provided with the theme.

Any file.

Tested locally.

Deploy.

Yes. https://github.com/grommet/grommet-theme-hpe/pull/161

Backwards compatible.

### DIFF
--- a/aries-site/src/examples/templates/forms/RequiredFieldsExample.js
+++ b/aries-site/src/examples/templates/forms/RequiredFieldsExample.js
@@ -1,5 +1,4 @@
 import React, { useContext } from 'react';
-import PropTypes from 'prop-types';
 import {
   Box,
   Button,
@@ -17,22 +16,6 @@ import { CircleAlert } from 'grommet-icons';
 
 const superPower = ['Flying', 'Sky Runner', 'Invisibility'];
 const weakness = ['Fire', 'PB & J', 'Kryptonite'];
-
-const RequiredFormField = props => {
-  const { required, label, ...rest } = props;
-  return (
-    <FormField
-      label={required ? `${label}*` : label}
-      required={required}
-      {...rest}
-    />
-  );
-};
-
-RequiredFormField.propTypes = {
-  required: PropTypes.bool,
-  label: PropTypes.string,
-};
 
 export const RequiredFieldsExample = () => {
   const [formValues, setFormValues] = React.useState({
@@ -80,7 +63,7 @@ export const RequiredFieldsExample = () => {
           value={formValues}
           onChange={onFormChange}
         >
-          <RequiredFormField
+          <FormField
             required
             error="Provide a unique name."
             htmlFor="name__input"
@@ -88,32 +71,27 @@ export const RequiredFieldsExample = () => {
             label="Name"
           >
             <TextInput id="name" name="name" />
-          </RequiredFormField>
-          <RequiredFormField
+          </FormField>
+          <FormField
             required
             htmlFor="superPower__input"
             name="superPower"
             label="Superpower"
           >
             <Select options={superPower} id="superPower" name="superPower" />
-          </RequiredFormField>
-          <RequiredFormField
+          </FormField>
+          <FormField
             required
             htmlFor="weakness__input"
             name="weakness"
             label="Weakness"
           >
             <Select options={weakness} id="weakness" name="weakness" />
-          </RequiredFormField>
-          <RequiredFormField
-            required
-            htmlFor="email__input"
-            name="email"
-            label="Email"
-          >
+          </FormField>
+          <FormField required htmlFor="email__input" name="email" label="Email">
             <TextInput id="email" name="email" />
-          </RequiredFormField>
-          <RequiredFormField
+          </FormField>
+          <FormField
             required
             help="Would you like to apply nemesis character"
             htmlFor="nemesis__input"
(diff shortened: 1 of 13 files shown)


aries-site/src/examples/templates/forms/RequiredFieldsExample.js
```diff
@@ -80,48 +63,43 @@ export const RequiredFieldsExample = () => {
...
-          </RequiredFormField>
-          <RequiredFormField
+          </FormField>
+          <FormField
             required
             help="Would you like to apply nemesis character"
             htmlFor="nemesis__input"
             name="nemesis"
             label="Nemesis"
           >
```
halocline, line 100: Just noticing this, but in this instance where we are asking "Would you like to apply nemesis character?", it feels awkward to have the field required, especially since the input is a toggle. I think we can remove the required without hurting the example.
```suggestion
          <FormField
            help="Would you like to apply nemesis character?"
            htmlFor="nemesis__input"
            name="nemesis"
            label="Nemesis"
          >
```

aries-site/src/examples/templates/forms/ShippingExample.js
```diff
@@ -302,7 +285,7 @@ export const ShippingExample = () => {
                 mask={phoneMask}
               />
             </FormField>
-            <RequiredFormField
+            <FormField
```
halocline, line 288: I am having a hard time tagging line 212 which reads `Shipping Address *`.

I am noticing two things:
- One, the visual treatment of the label and * is different than the new execution
- Two, how should we approach this in general? We should probably fix it while we are here.

```
            <FormField required htmlFor="firstName" name="firstName" label="Shipping Address">
              <TextInput
                id="firstName"
                name="firstName"
                placeholder="First Name"
              />
            </FormField>
```

and remove
```
<Text
              margin={{ horizontal: 'none', vertical: 'xsmall' }}
              size="xsmall"
            >
              Shipping Address *
            </Text>
```

halocline (commented):
A couple comments in line. Plus:

Wondering about instances where our validation examples are behaving like required... Should we apply if the example implies required (not all validation examples imply required)?
- https://deploy-preview-1410--keen-mayer-a86c8b.netlify.app/components/textinput#validation
- https://deploy-preview-1410--keen-mayer-a86c8b.netlify.app/components/textarea#validation
- https://deploy-preview-1410--keen-mayer-a86c8b.netlify.app/components/radiobuttongroup#validation
- https://deploy-preview-1410--keen-mayer-a86c8b.netlify.app/components/dateinput#validation
- https://deploy-preview-1410--keen-mayer-a86c8b.netlify.app/components/checkboxgroup#validation
- https://deploy-preview-1410--keen-mayer-a86c8b.netlify.app/components/checkbox#validation

taysea:
> A couple comments in line. Plus:
> 
> Wondering about instances where our validation examples are behaving like required... Should we apply if the example implies required (not all validation examples imply required)?
> 
> * https://deploy-preview-1410--keen-mayer-a86c8b.netlify.app/components/textinput#validation
> * https://deploy-preview-1410--keen-mayer-a86c8b.netlify.app/components/textarea#validation
> * https://deploy-preview-1410--keen-mayer-a86c8b.netlify.app/components/radiobuttongroup#validation
> * https://deploy-preview-1410--keen-mayer-a86c8b.netlify.app/components/dateinput#validation
> * https://deploy-preview-1410--keen-mayer-a86c8b.netlify.app/components/checkboxgroup#validation
> * https://deploy-preview-1410--keen-mayer-a86c8b.netlify.app/components/checkbox#validation

Yes I think given the error that these examples are showing "requires a value", then it would make sense to adjust these to required fields.

aries-site/src/examples/templates/forms/ShippingExample.js
taysea: This example is very wrong for accessibility. Glad you called it out. Ideally here, we'd want to be using a `fieldset` and group the inputs together where their individual labels are hidden (but still accessible by screen reader). Grommet doesn't currently support `fieldset` or visually-hidden labels. This example requires a bit more work, and I wonder if it would be better to remove it from the page for now until we can work on an improvement to the example?

halocline: Yes. We seem to keep kicking this can down the road. We should raise it as an issue / create an issue for it so that we make sure it gets addressed.

I'd be okay removing temporarily, but would want us to notify the team for the decision + the rationale + ask for any objections. And, also create an issue to talk through / propose a fieldset solution in Grommet. --> which could be something like:
```
const FormFieldSet = ( {children} ) => {
  return <fieldset ...>
    <legend ... />
    {children}
  </fieldset>;
};
```
I don't know all the fieldset options off the top of my head, plus whatever styling would be necessary.

halocline: Thinking more about the fieldset implementation above, I think we'd still need the hidden labels functionality though, right?

taysea: Yep we'd still need hidden labels as well. I'll file an issue if we don't have one already for fieldset work and raise it as an issue in planning next week. I'll leave the example in the page for now.

halocline (approved):

Issue #1426, opened 2021-02-16
Align center layer example to Figma 


MikeWalrath:
Please update center layer example to match the Figma file. The three areas that need alignment are: 
1. Icon needs to be updated to "email icon" as per the Figma. DS site uses "mail-option"
2. Add "x" close button in the upper right have corner 
3. Left aligned text 
https://www.figma.com/file/Ts53TAipMolmsv9DxWG3p0/HPE-Layer-Component?node-id=100%3A70

PR #1434, opened 2021-02-18
fix example to match figma

britt6612:
#### What does this PR do?
This PR aligns the center layer example with the figma file
There were 3 key things missing. 
I will need to get the updated icon this is not yet in Grommet Icons

- [X] Add "x" close button in the upper right have corner
- [X] Left aligned text

LayerCenterExample
https://www.figma.com/file/Ts53TAipMolmsv9DxWG3p0/HPE-Layer-Component?node-id=100%3A70
The example should match the figma fle
Issue #1426
no
yes

### DIFF
--- a/aries-site/src/examples/components/layer/LayerCenterExample.js
+++ b/aries-site/src/examples/components/layer/LayerCenterExample.js
@@ -52,7 +52,9 @@ export const LayerCenterExample = () => {
               background="background-contrast"
             >
               <Alert color="status-critical" />
-              <Text weight="bold">Footer if you need it</Text>
+              <Text color="text-strong" weight="bold">
+                Footer if you need it
+              </Text>
             </Box>
           </Box>
         </Layer>
--- a/aries-site/src/examples/components/layer/LayerCenterExample.js
+++ b/aries-site/src/examples/components/layer/LayerCenterExample.js
@@ -1,6 +1,6 @@
 import React, { useContext, useState } from 'react';
-import { Button, Box, Heading, Layer, ResponsiveContext, Text } from 'grommet';
-import { Alert, MailOption } from 'grommet-icons';
+import { Button, Box, Layer, ResponsiveContext, Text } from 'grommet';
+import { Alert, MailOption, Close } from 'grommet-icons';
 
 export const LayerCenterExample = () => {
   const [open, setOpen] = useState(false);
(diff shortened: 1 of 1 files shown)


aries-site/src/examples/components/layer/LayerCenterExample.js
```diff
@@ -18,27 +18,36 @@ export const LayerCenterExample = () => {
...
-                <MailOption />
-              </Box>
-              <Box gap="medium">
-                <Heading margin="none" size="small">
+            <Box justify="between" direction="row">
+              <Box flex={false} gap="small" direction="row">
+                <Box justify="center">
+                  <MailOption />
+                </Box>
+                <Text margin="none" size="xlarge">
```
taysea, line 28: Any reason to not use Heading?

aries-site/src/examples/components/layer/LayerCenterExample.js
```diff
@@ -18,27 +18,36 @@ export const LayerCenterExample = () => {
...
+                For modal dialogs, the use case will determine the design and
+                size of the box for your content
+              </Text>
             </Box>
+          </Box>
+          <Box
+            fill="vertical"
+            overflow="auto"
+            width={size !== 'small' ? 'medium' : undefined}
+          >
```
taysea, line 50: Can't comment on line 54, but I think the pad should be `{{ vertical: 'small', horizontal: 'medium' }}`

aries-site/src/examples/components/layer/LayerCenterExample.js
```diff
@@ -18,27 +18,36 @@ export const LayerCenterExample = () => {
...
                   Modal Dialog
-                </Heading>
-                <Text>
-                  For modal dialogs, the use case will determine the design and
-                  size of the box for your content
                 </Text>
               </Box>
+              <Box justify="center">
+                <Button
+                  icon={<Close size="small" align="center" />}
```
taysea, line 34: The icon doesn't look aligned when you hover over it. Do you notice that? Seems like `FormClose` icon works better and aligns with what we're doing in Filtering.

britt6612: okay let me change that and see I did notice the alignment was off without the align which didn't feel correct

aries-site/src/examples/components/layer/LayerCenterExample.js
britt6612: You're correct I changed it  back but used level 3 to match figma

taysea (approved):
LGTM.

PR #1443, opened 2021-02-22
POC delete footer

britt6612:
Start a POC for deleting footer

[preview](https://deploy-preview-1443--keen-mayer-a86c8b.netlify.app/)


aries-site/src/components/home/Community.js
```diff
@@ -72,6 +72,23 @@ const Backlog = () => (
...
+  <Box gap="medium">
+    <Info size="large" color="text-strong" />
+    <Box>
+      <Heading level={2} margin="none" size="xsmall">
+        Feedback
+      </Heading>
+      <Text>Let us know your feedback!</Text>
+    </Box>
+    <LinkButton
+      label="View Backlog"
```
taysea, line 85:
```suggestion
      label="Provide feedback"
```

aries-site/src/components/home/Community.js
```diff
@@ -72,6 +72,23 @@ const Backlog = () => (
...
+      href="https://github.com/orgs/grommet/projects/10"
```
taysea, line 86:
```suggestion
      href="/feedback"
```

aries-site/src/layouts/main/Header.js
```diff
@@ -47,6 +48,7 @@ const StyledHeader = ({ ...rest }) => {
             icon={<SearchIcon />}
             onClick={() => setSearchFocused(true)}
           />
+          <ThemeModeToggle />
```
taysea, line 51: When I open the search, the theme toggle disappears and the search shifts down. it feels like it'd be fine to let the theme toggle stay where it is, and the search input opens to the left

aries-site/src/components/content/ThemeModeToggle.js
```diff
@@ -9,12 +9,14 @@ export const ThemeModeToggle = ({ active }) => {
...
   return (
-    <Button
-      active={active}
-      onClick={() => darkMode.toggle()}
-      icon={icon}
-      label={label}
-    />
+    <Tip content={label}>
+      <Button
+        id="theme-button"
```
taysea, line 14: Can we use tip prop?

britt6612: your right good catch

britt6612: so since the label is an object I cant use the tip prop in `button` since passing children to react for as an object

PR #1446, opened 2021-02-24
example 3 for taking footer out

britt6612:
POC for taking out the footer and removing header to be in the middle of the page


britt6612:
not going with this approach

PR #1492, opened 2021-03-17
Restore copy changes

taysea:
#### What does this PR do?

Mike made copy changes from Brand Central feedback that somehow got removed from master. Restoring here.


britt6612:
Looks good!  not sure why they got taken out from before!

PR #1518, opened 2021-03-30
Update dashboards.mdx

abbyperez:


abbyperez:

ericsoderberghp (commented):
Just a small wording suggestion, as "/" could be interpreted as a division symbol with numbers on each side.

PR #1519, opened 2021-03-30
Update table.mdx

abbyperez:

### DIFF
--- a/aries-site/src/examples/components/table/TableSingleSelectExample.js
+++ b/aries-site/src/examples/components/table/TableSingleSelectExample.js
@@ -277,7 +277,7 @@ export const TableSingleSelectExample = () => {
         </Box>
         <Box gap="small" direction="column">
           {pageDetailsList.map((item, index) => (
-            <Text key={index} size="small">
+            <Text key={index} color="text-strong">
               {item}{' '}
             </Text>
           ))}
--- a/aries-site/src/examples/components/table/TableSingleSelectExample.js
+++ b/aries-site/src/examples/components/table/TableSingleSelectExample.js
@@ -9,7 +9,7 @@ import {
 } from 'grommet';
 import { FormPrevious } from 'grommet-icons';
 
-const orderDeatils = [
+const orderDetails = [
   'Order Name:',
   'Purchase Order:',
   'State:',
--- a/aries-site/src/examples/components/table/TableSingleSelectExample.js
+++ b/aries-site/src/examples/components/table/TableSingleSelectExample.js
@@ -1,4 +1,5 @@
 import React from 'react';
+import PropTypes from 'prop-types';
 import {
   Box,
   DataTable,
@@ -9,15 +10,19 @@ import {
 } from 'grommet';
 import { FormPrevious } from 'grommet-icons';
 
-const orderDetails = [
-  'Order Name',
-  'Purchase Order',
-  'State',
-  'Service',
-  'Tenant',
-  'Order Date',
-  'Contact',
-];
+const DetalsPage = ({ orderDetails, orderPageDetails, ...rest }) => (
+  <Box direction="row" align="start" {...rest}>
+    <Box width="small">
+      <Text size="small">{orderDetails}</Text>
+    </Box>
+    <Text color="text-strong">{orderPageDetails}</Text>
+  </Box>
+);
+
+DetalsPage.propTypes = {
+  orderDetails: PropTypes.string,
+  orderPageDetails: PropTypes.string,
+};
 
 const data = [
   {
@@ -215,15 +220,6 @@ export const TableSingleSelectExample = () => {
   const size = React.useContext(ResponsiveContext);
   const [pageDetails, setPageDetals] = React.useState({});
 
-  const pageDetailsList = [
-    pageDetails.id,
-    pageDetails.purchaseOrder,
-    pageDetails.state,
-    pageDetails.service,
-    pageDetails.tenant,
-    pageDetails.orderDate,
-  ];
-
   return !pageDetails.id ? (
     <>
       <Heading level={3} margin={{ bottom: 'small', top: 'none' }}>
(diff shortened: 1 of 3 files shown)


aries-site/src/pages/components/table.mdx
```diff
@@ -199,7 +199,9 @@ is setting out to accomplish.
 
 See [Filtering](/templates/filtering) for greater detail.
 
-### Selecting single records
+### Navigation via table
+
+Clicking on a table row navigates to a page whose name matches the primary column in the row.
```
ericsoderberghp, line 204: I think the key is that clicking takes the user to a context where the most prominent heading matches the primary column text. This could either be a page or a modal. I think the example should be enhanced to show that as well, since the current example doesn't do that.

aries-site/src/examples/components/table/TableSingleSelectExample.js
```diff
@@ -193,27 +201,55 @@ const columns = [
...
+          setPageDetals('');
+        }}
+        alignSelf="start"
+        icon={<FormPrevious />}
+        label="Orders"
+       />
+      <Box margin={{ horizontal: 'large' }} border="bottom">
+        <Heading level={3}>Order Number: {pageDetails.id}</Heading>
+      </Box>
+      <Text
```
taysea, line 221: Thinking this should be an `h2`

aries-site/src/examples/components/table/TableSingleSelectExample.js
```diff
@@ -193,27 +201,55 @@ const columns = [
...
+      >
+        Details
+      </Text>
+      <Box
+        margin={{ horizontal: 'large' }}
+        gap="medium"
+        direction="row-responsive"
+      >
+        <Box gap="small" direction="column">
+          <Text>Order Name:</Text>
```
taysea, line 233: All of these should be `size="small"`

aries-site/src/examples/components/table/TableSingleSelectExample.js
```diff
@@ -193,27 +201,55 @@ const columns = [
...
+          <Text>Purchase Order:</Text>
+          <Text>State:</Text>
+          <Text>Service:</Text>
+          <Text>Tenant:</Text>
+          <Text>Contact:</Text>
+          <Text>Order Date:</Text>
+        </Box>
+        <Box gap="small" direction="column">
+          {pageDetails.orderName}
```
taysea, line 242: Wrap these in `<Text>` ?

aries-site/src/examples/components/table/TableSingleSelectExample.js
```diff
@@ -193,27 +201,55 @@ const columns = [
...
-    other actions as you see fit.
-  `);
-};
-
 export const TableSingleSelectExample = () => {
   const size = React.useContext(ResponsiveContext);
+  const [pageDetails, setPageDetals] = React.useState('');
 
-  return (
+  return pageDetails.id !== undefined ? (
```
taysea, line 208: For clarity for those coming into this, can we swap the condition so the Table return comes first?

`!showPageDetails ? <Table /> : <PageDetails />`

aries-site/src/examples/components/table/TableSingleSelectExample.js
```diff
@@ -193,27 +201,55 @@ const columns = [
...
-    
-    You can use onClick() to navigate to a record's detail
-    page, open a panel or modal to edit the record, or perform 
-    other actions as you see fit.
-  `);
-};
-
 export const TableSingleSelectExample = () => {
   const size = React.useContext(ResponsiveContext);
+  const [pageDetails, setPageDetals] = React.useState('');
```
taysea, line 206: Should this be empty object? What is the type of `datum`

aries-site/src/examples/components/table/TableSingleSelectExample.js
```diff
@@ -225,10 +261,35 @@ export const TableSingleSelectExample = () => {
...
-          onClickRow={({ datum }) => onClickHandler(datum)}
+          onClickRow={({ datum }) => setPageDetals(datum)}
           pin={size === 'small'}
         />
       </Box>
     </>
   );
 };
+
+/* <>
```
taysea, line 272: Remove unused code

aries-site/src/examples/components/table/TableSingleSelectExample.js
britt6612: ok ill add that for the ones below as well

aries-site/src/examples/components/table/TableSingleSelectExample.js
britt6612: oops will do

aries-site/src/examples/components/table/TableSingleSelectExample.js
taysea: Whoops I meant "Order Number: OID789345" would be h1 (since it's page title) -- we may adjust size to something small.

Then "Details" should be an h2 (also maybe adjust size to small). double check w designs.

aries-site/src/examples/components/table/TableSingleSelectExample.js
```diff
@@ -230,66 +247,24 @@ export const TableSingleSelectExample = () => {
...
         </Box>
         <Box gap="small" direction="column">
-          {pageDetails.orderName}
-          {pageDetails.purchaseOrder}
-          {pageDetails.state}
-          {pageDetails.service}
-          {pageDetails.tenant}
-          {pageDetails.contact.email}
-          {pageDetails.orderDate}
+          <Text size="small">{pageDetails.orderName} </Text>
```
taysea, line 259: Any way to store these data items in an array and map through? instead of having to repeat `<Text size="small">` each time?

Same with the above section

aries-site/src/examples/components/table/TableSingleSelectExample.js
```diff
@@ -1,5 +1,23 @@
...
+  Box,
+  DataTable,
+  Heading,
+  Text,
+  ResponsiveContext,
+  Button,
+} from 'grommet';
+import { FormPrevious } from 'grommet-icons';
+
+const orderDeatils = [
```
taysea, line 12:
```suggestion
const orderDetails = [
```
probably have to change below as well where you're mapping

aries-site/src/examples/components/table/TableSingleSelectExample.js
```diff
@@ -1,5 +1,23 @@
...
+  DataTable,
+  Heading,
+  Text,
+  ResponsiveContext,
+  Button,
+} from 'grommet';
+import { FormPrevious } from 'grommet-icons';
+
+const orderDetails = [
+  'Order Name:',
```
taysea, line 13: also it seems like the designs don't use a ":" so I think we can remove that in all of these and just have the text.

britt6612: your right I think I added that by habbit

aries-site/src/examples/components/table/TableSingleSelectExample.js
```diff
@@ -225,10 +236,54 @@ export const TableSingleSelectExample = () => {
...
+        <Box gap="small" direction="column">
+          {orderDetails.map((item, index) => (
+            <Text key={index} size="small">
+              {item}
+            </Text>
+          ))}
+        </Box>
+        <Box gap="small" direction="column">
+          {pageDetailsList.map((item, index) => (
+            <Text key={index} size="small">
```
taysea, line 280:
```suggestion
            <Text key={index}>
```

aries-site/src/examples/components/table/TableSingleSelectExample.js
```diff
@@ -225,10 +236,54 @@ export const TableSingleSelectExample = () => {
...
+              {item}{' '}
+            </Text>
+          ))}
+          <Text size="small">{pageDetails.contact.email}</Text>
```
taysea, line 284:
```suggestion
          <Text>{pageDetails.contact.email}</Text>
```

aries-site/src/examples/components/table/TableSingleSelectExample.js
taysea: Designs call for the data to be text default

taysea: and calls for text-strong

aries-site/src/examples/components/table/TableSingleSelectExample.js
```diff
@@ -225,10 +236,54 @@ export const TableSingleSelectExample = () => {
...
+      >
+        Details
+      </Heading>
+      <Box
+        margin={{ horizontal: 'large' }}
+        gap="medium"
+        direction="row-responsive"
+      >
+        <Box gap="small" direction="column">
+          {orderDetails.map((item, index) => (
```
taysea, line 272: I think this section should be mapping through and render each item of the `orderDetails` and `pageDetailsList` in a row with `align="start"` so that the line heights match. Right now, it's getting misaligned because the rows have different heights given differing text sizes:
Screen Shot 2021-04-05 at 10 14 36 AM

taysea (approved):
LGTM.

PR #1522, opened 2021-03-31
Adjust Github actions to accomodate forked repo

taysea:
#### What does this PR do?

For public repositories, if a forked repo files a PR, the fork will not have access to any github secrets. In our checks, we rely on an API Key held in github secrets to run the snapshot tests. For security reasons, it's not recommended that these keys are passed to forked repos.

However, for teams that are aiming to file copy edits, we still would like the rest of the e2e tests to run. To do so, we create a separate script `test:fork` that will run the e2e tests (but not snapshot tests) on headless chrome. An update is made to the `testcafe.yml` file to check if a repo is forked or not.

.github/workflows/testcafe.yml

Backwards compatible.

### DIFF
--- a/.github/workflows/testcafe.yml
+++ b/.github/workflows/testcafe.yml
@@ -25,7 +25,14 @@ jobs:
         working-directory: aries-site
       # Tests are only run on headless Chrome
       - name: Build and run e2e tests on Chrome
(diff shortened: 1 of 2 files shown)


britt6612:
LGTM!

PR #1536, opened 2021-04-05
Update how wizard width is determined

### LINKED ISSUE #1399: Update Wizard input "Grid"
Wizard inputs should define fixed width, such as "medium" and have a gap of "large". The total width of the wizard will not be a t-shirt size, but instead the sum of (# columns * "medium") + gap. This conclusion was reached between CCS and Eric.

Currently, wizard width is t-shirt sized and inputs will that space.

taysea:
#### What does this PR do?
Wizard content/footer width should be determined by the total width of the inputs, gap between columns, and a small amount of pad to accommodate focus outline on inputs.

In a wizard, the width should remain constant across steps, matching the width needed for the step with the max number of columns. The inputs themselves should be `width="medium"` and so in the case of a two-column wizard the width of the content and footer should be:
```
medium * numberColumns + gap * (numberColumns - 1) + tiny amount of pad for focus ring
```

This PR adjusts how this styling occurs, makes sure the inputs wrap appropriately and maintain their designated widths, and ensures there's always some small margin between step content/footer content and the edges of the screen (which previously was not applying properly).

aries-site/src/examples/templates/wizard/WizardValidationExample.js

Check wizard deploy page.

Same as above.

Discussions with CCS determined that the inputs should remain a width of `medium` and wrap when there's not enough space.

Closes #1399 

Yes. Updated Wizard layout so that inputs always remain width="medium" and wrap when there is not sufficient horizontal space.

New behavior to align with a conclusion between Design System and CCS.

### DIFF
--- a/aries-site/src/examples/templates/wizard/TwoColumnWizardExample.js
+++ b/aries-site/src/examples/templates/wizard/TwoColumnWizardExample.js
@@ -64,7 +64,6 @@ const StepOne = () => {
       direction={size !== 'small' ? 'row' : 'column-reverse'}
       margin={{ bottom: 'medium' }}
       gap={size === 'small' ? 'small' : undefined}
-      width={{ max: 'large' }}
       justify="between"
       wrap
     >
@@ -74,8 +73,8 @@ const StepOne = () => {
         gap="medium"
         flex={false}
       >
-        <>
-          <Box>
+        <Box gap="medium">
+          <>
             <FormField
               label="Email"
               htmlFor="twocolumn-textinput"
@@ -103,13 +102,13 @@ const StepOne = () => {
                 options={['Radio button 1', 'Radio button 2']}
               />
             </FormField>
-          </Box>
+          </>
           {!error.isValid && (
             <Error>There is an error with one or more inputs.</Error>
           )}
-        </>
+        </Box>
       </Box>
-      <Box flex width={{ max: 'xsmall' }} />
+      <Box flex width={{ max: 'xxsmall' }} />
       <Guidance />
     </Box>
   );
@@ -241,8 +240,8 @@ export const TwoColumnWizardExample = () => {
     container.scrollTop = -header.getBoundingClientRect().bottom;
   }, [activeIndex, open]);
 
-  const countColumns = 2;
-  const width = getWidth(countColumns, theme, size);
+  const numberColumns = 2;
+  const width = getWidth(numberColumns, theme, size);
   return (
     <WizardContext.Provider
       value={{
@@ -285,7 +284,7 @@ const Guidance = () => {
       pad="medium"
       round="small"
       flex
-      width={size !== 'small' ? { max: 'medium' } : '100%'}
+      width={size !== 'small' ? { min: 'small', max: 'medium' } : '100%'}
     >
       <Text color="text-strong" size="large">
         When guidance is required for the form or content of the wizard, you
--- a/aries-site/src/examples/templates/wizard/TwoColumnWizardExample.js
+++ b/aries-site/src/examples/templates/wizard/TwoColumnWizardExample.js
@@ -240,8 +240,7 @@ export const TwoColumnWizardExample = () => {
     container.scrollTop = -header.getBoundingClientRect().bottom;
   }, [activeIndex, open]);
 
-  const numberColumns = 2;
-  const width = getWidth(numberColumns, theme, size);
+  const width = getWidth(numberColumns = 2, theme, size);
   return (
     <WizardContext.Provider
       value={{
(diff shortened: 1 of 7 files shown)


britt6612 (approved):

aries-site/src/examples/templates/wizard/WizardValidationExample.js
```diff
@@ -131,12 +133,12 @@ const StepOne = () => {
     WizardContext,
   );
   const size = useContext(ResponsiveContext);
-  console.log(error.isValid);
+
   return (
     <>
       <Box margin={{ bottom: 'medium' }}>
```
halocline, line 139: Can this margin be applied to the Grid itself? I'm trying to see if we can minimize the number of Boxes

halocline, line 139: Can we eliminate either the Fragment or Box? I don't think we need both.

aries-site/src/examples/templates/wizard/WizardValidationExample.js
```diff
@@ -57,7 +59,7 @@ const validation = [
 
 const stepOneInputs = [
   (attemptedAdvance, error, formValues, setError) => (
-    <Box>
+    <Box width={{ max: 'medium' }}>
```
halocline, line 62: These Box wrappers (L62 & L112) seem to be doing the same thing. Can we move inside the map at L144?

aries-site/src/examples/templates/wizard/WizardValidationExample.js
halocline: Actually, it looks like this margin-bottom is desired across steps. It looks like the desire is to have a medium space between the StepContent and StepFooter. If so, can we eliminate each individual instance and handle that by placing a Box with gap around
```
<Box gap="medium">
  <StepContent />
  <StepFooter />
</Box>
```

halocline (commented):
Question double checking the getWidth logic, plus some potential opportunities to clean up layout -- trying to see if we can reduce the number of Boxes.

aries-site/src/examples/templates/wizard/utils.js
```diff
@@ -0,0 +1,18 @@
+// Wizard width is not a t-shirt size but instead dependent
+// on the max number of columns it will contain * the width
+// of those columns (medium) + gap between columns + small amount
```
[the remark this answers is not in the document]
taysea: Good call, you're right!

aries-site/src/examples/templates/wizard/TwoColumnWizardExample.js
```diff
@@ -237,6 +241,8 @@ export const TwoColumnWizardExample = () => {
     container.scrollTop = -header.getBoundingClientRect().bottom;
   }, [activeIndex, open]);
 
+  const countColumns = 2;
```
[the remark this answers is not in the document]
taysea: Yes, maybe `numberColumns` would be better?

aries-site/src/examples/templates/wizard/WizardValidationExample.js
taysea: Thanks let me have a look into it.

taysea: I think we want to keep with bottom margin, otherwise the scroll of the content doesn't reach all the way to the footer. There is a medium gap of white which makes the content look like it's broken up strangely.

taysea: Yes, can eliminate the Fragment, thanks.

aries-site/src/examples/templates/wizard/WizardValidationExample.js
taysea: Yes, good cleanup

aries-site/src/examples/templates/wizard/TwoColumnWizardExample.js
```diff
@@ -237,6 +240,8 @@ export const TwoColumnWizardExample = () => {
     container.scrollTop = -header.getBoundingClientRect().bottom;
   }, [activeIndex, open]);
 
+  const numberColumns = 2;
+  const width = getWidth(numberColumns, theme, size);
```
halocline, line 244:
```suggestion
  const width = getWidth(numberColumns = 2, theme, size);
```

aries-site/src/examples/templates/wizard/WizardValidationExample.js
```diff
@@ -267,6 +268,8 @@ export const WizardValidationExample = () => {
     container.scrollTop = -header.getBoundingClientRect().bottom;
   }, [activeIndex, open]);
 
+  const numberColumns = 2;
+  const width = getWidth(numberColumns, theme, size);
```
halocline, line 272:
```suggestion
  const width = getWidth(numberColumns = 2, theme, size);
```

halocline (approved):
Applitools snapshots look reasonable to me.

2 inconsequential nits, but good to go as is.

PR #1537, opened 2021-04-05
Update checkbox.mdx

abbyperez:

### DIFF
--- a/aries-site/src/pages/components/checkbox.mdx
+++ b/aries-site/src/pages/components/checkbox.mdx
@@ -22,7 +22,7 @@ CheckBox can be used to present an individual option or group of options to the 
 
 ### Usage
 
-CheckBox can be used in the form of a [toggle](#toggle), a single option which can indicate something such as on or off. Toggle is prefered to quickly switch between two possible states. Toggles usually provide immediate results in which gives users the ability to control their preference.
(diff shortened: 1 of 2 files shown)


aries-site/src/pages/components/checkbox.mdx
```diff
@@ -22,7 +22,7 @@ CheckBox can be used to present an individual option or group of options to the
 
 ### Usage
 
-CheckBox can be used in the form of a [toggle](#toggle), a single option which can indicate something such as on or off. Toggle is prefered to quickly switch between two possible states. Toggles usually provide immediate results in which gives users the ability to control their preference.
+CheckBox can be used in the form of a [](#toggle), a single option which can indicate something such as on or off. Toggle is prefered to quickly switch between two possible states. Toggles usually provide immediate results (i.e. no form submission is required) in which gives users the ability to control their preference.
```
taysea, line 25:
```suggestion
CheckBox can be used in the form of a [toggle](#toggle), a single option which can indicate something such as on or off. Toggle is prefered to quickly switch between two possible states. Toggles usually provide immediate results (i.e. no form submission is required) in which gives users the ability to control their preference.
```
We should leave `toggle` here because it is linking to a section later in the page. If the intent is to remove the link "toggle" entirely, then the sentence may need some rewording

taysea (approved):
LGTM.

Issue #1535, opened 2021-04-02
Search & Filtering examples - Ensure search value is cleared when "Clear filters" anchor is clicked.


halocline:
**Expected behavior**: 
When "Clear filters" is activated, all filters and results are reset to an unfiltered result set.

**Current behavior**:
When "Clear filters" is activated, filter facets are reset, results are reset, however the search field still retains a value. This can be confusing as it implies the search is still being applied to the result set, when in fact the result set is shown in its entirety.

**Background**
Related to this thread: https://hpe.slack.com/archives/C04LMJWQT/p1617378911062300

PR #1543, opened 2021-04-07
add search to also be cleared

britt6612:
#### What does this PR do?
This PR adds setSearch to be passed through to the `reset filters` This will ensure the search is clear when someone clicks `clearAll`
lists
local
check to see if the search is cleared out 
issue #1535 
no
backwards compatible

### DIFF
--- a/aries-site/src/examples/templates/filtering/FilteringCards.js
+++ b/aries-site/src/examples/templates/filtering/FilteringCards.js
@@ -136,9 +136,9 @@ export const FilteringCards = ({ containerRef }) => {
 
     let filterResults;
     const filterKeys = Object.keys(criteria);
-    filterResults = array.filter(item => 
+    filterResults = array.filter(item =>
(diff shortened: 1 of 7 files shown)


aries-site/src/examples/templates/filtering/FilteringCards.js
```diff
@@ -254,6 +256,7 @@ const Filters = ({
     setName([]);
     setFilters(defaultFilters);
     setFiltering(false);
+    setSearch();
```
taysea, line 259: Since we default to `''` should this `setSearch('')` ?

halocline: Yes, good catch!

britt6612: Yes fixing now!

taysea (approved):
LGTM. Nice job.

PR #1544, opened 2021-04-07
Use new plain: true for DataTable column

taysea:
[Preview](https://deploy-preview-1544--keen-mayer-a86c8b.netlify.app/components/table#selecting-multiple-records-and-batch-actions)

#### What does this PR do?
Leveraging new DataTable behavior of `plain: true` on DataTable column to allow an entire cell to be clickable. Updating our example.

See "Multi Select" Table example.

Backwards compatible.


britt6612 (approved):

halocline (approved):
Looks MUCH better!

PR #1545, opened 2021-04-09
Fix search focus

### LINKED ISSUE #1532: Clicking within an active search closes search
When you click on the search icon then click in the search input it closes the search. 
image

britt6612:
#### What does this PR do?
This PR adds a `onBlur` this `onBlur` sets `focus` to false as long as the user is not interacting with the drop.
aries-site/src/layouts/navigation/Search.js
Ran the testcafe 
click on the search and you will see that as long as you click within the `textinput` it will stay focused.
Also tabbing on keyboard and and clicking a suggestion. 
https://github.com/grommet/hpe-design-system/issues/1532
closes #1532 
no
backwards compatible

### DIFF
--- a/aries-site/src/layouts/navigation/Search.js
+++ b/aries-site/src/layouts/navigation/Search.js
@@ -65,9 +65,9 @@ export const Search = ({ focused, setFocused }) => {
 
   const onSelect = event => {
     const href = nameToPath(event.suggestion);
-
     if (internalLink.test(href)) {
       router.push(href);
+      setFocused(false);
     } else {
       // external links should open in a new tab
       window.open(href);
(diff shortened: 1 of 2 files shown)


aries-site/src/layouts/navigation/Search.js
```diff
@@ -88,12 +88,18 @@ export const Search = ({ focused, setFocused }) => {
...
-              onSelect={onSelect}
+              onSuggestionSelect={onSelect}
               suggestions={suggestions}
-              onSuggestionsClose={() => setFocused(false)}
               value={value}
               plain
               reverse
+              onBlur={event => {
+                // Only treat it as a blur if the element receiving focus
+                // isn't in our drop. The relatedTarget will be our drop
```
taysea, line 98: Good comment, thanks!

aries-site/src/layouts/navigation/Search.js
```diff
@@ -65,9 +65,9 @@ export const Search = ({ focused, setFocused }) => {
 
   const onSelect = event => {
     const href = nameToPath(event.suggestion);
-
     if (internalLink.test(href)) {
       router.push(href);
+      setFocused(false);
```
taysea, line 70: Do we need to add this here? If the site navigates will setFocused already change to false?

britt6612: yes I added it because with keyboard usage without setting `focus(false)` the input was staying open after the user selected a  suggestion and clicked enter.

taysea: Great, thanks for clarifying.

taysea (approved):
LGTM. Nice job.

halocline (approved):

Issue #1530, opened 2021-04-01
Clean up Figma Color page to align with the site 


MikeWalrath:
zfigma.com/file/eZYR3dtWdb9U90QvJ7p3T9/HPE-Color-Styles?node-id=0%3A1
There is currently some confusing language within the color page. Specifically "Secondary." It would be nice to clean it up and align with the site. 
![image](https://user-images.githubusercontent.com/17712467/113327565-5b764180-92d8-11eb-9075-0d6c5b15b198.png)

vavalos5:
I've made the changes above to better align with the site.

Figma file: https://www.figma.com/file/eZYR3dtWdb9U90QvJ7p3T9/HPE-Color-Styles?node-id=0%3A1

One thing I noticed is that on the DS site, we have 2 colors named "green"
<img width="1088" alt="Screen Shot 2021-04-13 at 11 21 03 AM" src="https://user-images.githubusercontent.com/62117454/114601509-6cb13d80-9c4a-11eb-836a-028aca3fe332.png">

Our brand color is named "green" and our light palette green color is called "green". However, in Figma, we have it as "light Green. Though on the DS site, the "light green color" is under "Light Palette", I think we should call all of the colors "light (color)" do differentiate that from our brand color.

<img width="344" alt="Screen Shot 2021-04-13 at 11 23 00 AM" src="https://user-images.githubusercontent.com/62117454/114601801-c7e33000-9c4a-11eb-8a57-6d8eb0c2942e.png">

ericsoderberghp:
The color names on the design system site refer to the names in the HPE theme. There are no colors named "light ..." in the theme. So, the way the design system site labels the colors is accurate in that sense. Would it make sense to re-label the Figma colors to something like "green - light"?

vavalos5:
Done :)

vavalos5:
Changed brand color from "green" to "brand". Aligned this with Mike and Matt. Matt has also changed this in the back end. All colors are now aligned to the Design System site.

PR #1550, opened 2021-04-14
Relabel brand color to align namespaces across theme, site, Figma

halocline:
#### What does this PR do?

Completes https://github.com/grommet/hpe-design-system/issues/1530

Also, yarn build was failing to build Next. Error was complaining about how `<hr />` were being used on the whats-new page. I removed the `----------------` throughout the mdx.  Latest entries on the whats-new page have not been making use of the `hr` as a delineator, so figured we could live without.

color.js

### DIFF
--- a/aries-site/src/data/color.js
+++ b/aries-site/src/data/color.js
@@ -11,8 +11,8 @@ const { layer } = theme;
 /* Color Palette */
 const primaryColors = [
   {
-    name: 'green',
-    value: 'green!',
+    name: 'brand',
+    value: 'brand',
     hex: colors['green!'],
   },
 ];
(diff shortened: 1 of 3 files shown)


vavalos5 (approved):
LGTM! Thanks for making these changes.

aries-site/src/data/color.js
```diff
@@ -11,8 +11,8 @@ const { layer } = theme;
 /* Color Palette */
 const primaryColors = [
   {
-    name: 'green',
-    value: 'green!',
+    name: 'brand',
+    value: 'brand',
     hex: colors['green!'],
```
taysea, line 16:
```suggestion
    hex: colors['brand'],
```
Wondering if we want to change this too just for consistency? Doesn't affect the HEX value, so I'm fine either way.

halocline: Confirmed that we need to leave as `green!`. `brand` will return and display the string `green!`

PR #1556, opened 2021-04-15
Simplified wizard form validation and set focus on invalid field on next/submit

### LINKED ISSUE #1540: Ensure wizard validation aligns with forms
It would be good to review the wizard examples to ensure the validation scroll aligns with the form implementation. Currently to validation happens below the scroll but the user isnt notified that the validation has occurred. 
image
image

With a sticky footer would it be a good idea to add the error messaging within the footer?

MikeKingdom:
[Preview](https://deploy-preview-1556--keen-mayer-a86c8b.netlify.app/templates/wizard)

#### What does this PR do?
Simplified how field validation is done in Wizards.
Made Next/Submit move focus to first invalid field in the form.

WizardValidationExample.js

yarn start

yarn start and check the wizard pages.

Previously the wizards would avoid having the page Next buttons do a form submit. As a result they had to re-implement ways of doing field validations. This lets the Next button be a submit button so it allows all the usual Form and FormField validation capability work as usual. It just doesn't call the passed in onSubmit until the Finish button on the last page is pressed.

Fixes #1540
 
No

Backwards compatible


aries-site/src/examples/templates/wizard/TwoColumnWizardExample.js
```diff
@@ -79,10 +54,13 @@ const StepOne = () => {
               label="Email"
               htmlFor="twocolumn-textinput"
               name="twocolumn-textinput"
-              error={error.email}
-              onChange={() =>
-                attemptedAdvance && setError(stepOneValidate(formValues))
-              }
+              validate={{
```
taysea, line 57: I think we only want this to validate onBlur/onChange if the user has already tried to advance to the next step. My understanding was we don't want to run validation as they're first trying to go through inputs because popping errors up prematurely can feel alarming. 

What I noticed here was if I typed an invalid email and pressed "tab" to move on to the next input, the validation ran and the email input got put in focus right away. Expected would be if this is my first attempt at entering info, no validation until I press "Next".

taysea: ^^ Was testing on `WizardValidationExample` but same behavior should apply here.

aries-site/src/examples/templates/wizard/components/StepContent.js
```diff
@@ -36,7 +65,9 @@ export const StepContent = () => {
             id={`${id}-form`}
             value={formValues}
             onChange={nextValue => setFormValues(nextValue)}
-            onSubmit={({ value }) => console.log(value)}
+            onSubmit={handleSubmit}
+            onValidate={onValidate}
+            validate="blur"
```
taysea, line 70: My understanding is we want validation to happen when a user attempts to advance to the next step (Clicking "Next") to avoid throwing validation errors too often to the user. More inviting if we let them fill out the inputs and just display errors when they click "next" meaning they are trying to move on but we should tell them if something went wrong.

taysea (commented):
Left a few comments re: changes in validation before vs what we have existing. Also, I noticed this console warning. I had a value in "First name" and then clicked "Next":
Screen Shot 2021-04-16 at 8 27 31 AM

Screen Shot 2021-04-16 at 8 35 59 AM

In another time that I was interacting I saw a similar warning but swap out where it says "Button" and insert "RadioButtonGroup".

aries-site/src/examples/templates/wizard/TwoColumnWizardExample.js
MikeKingdom: Thanks Taylor. I was wondering about this. The main reason I did onBlur is it was easiest to make the error message at the bottom come and go at the right time. It's confusing if it sticks around until Next is hit again. I think the other way around this would to put back in the 'attemptedAdvance' flag and something in the onChange handlers although it seemed very awkward to implement. Thoughts on a better way to provide that message at the bottom?

MikeKingdom:
> Left a few comments re: changes in validation before vs what we have existing. Also, I noticed this console warning. I had a value in "First name" and then clicked "Next":

I'll take a closer look at those. I'm not quite sure how it's trying to set state during the render(s)

MikeKingdom:
> > Left a few comments re: changes in validation before vs what we have existing. Also, I noticed this console warning. I had a value in "First name" and then clicked "Next":
> 
> I'll take a closer look at those. I'm not quite sure how it's trying to set state during the render(s)

Any set state related things and or .focus() during the Form onValidation callback seems to cause this. To avoid the issue I put those in timeouts with 0 delay which basically makes them happen immediately after the onValidation chain of calls. Any other ideas welcome!

aries-site/src/examples/templates/wizard/components/StepContent.js
MikeKingdom: Done. I switched it back to the default which is validation="submit". I instead make the onChange of the validated fields clear the error message at the very end of the form until the user hits the Next or Finish button again which seems less confusing. There's isn't another good way to figure out the valid state of the form from what I've seen. I thought grommet sets the aria-invalid property on elements but that doesn't seem to be true.

aries-site/src/examples/templates/wizard/TwoColumnWizardExample.js
MikeKingdom: Fixed as noted below.

taysea:
> > > Left a few comments re: changes in validation before vs what we have existing. Also, I noticed this console warning. I had a value in "First name" and then clicked "Next":
> > 
> > 
> > I'll take a closer look at those. I'm not quite sure how it's trying to set state during the render(s)
> 
> Any set state related things and or .focus() during the Form onValidation callback seems to cause this. To avoid the issue I put those in timeouts with 0 delay which basically makes them happen immediately after the onValidation chain of calls. Any other ideas welcome!

Do you think it has to do with the timeout set on this useEffect by chance? This timeout was introduced to solve a different bug but I wonder if it conflicts with what we're trying to do here? https://github.com/grommet/grommet/blob/master/src/js/components/Form/Form.js#L83

aries-site/src/examples/templates/wizard/components/StepContent.js
```diff
@@ -1,19 +1,48 @@
...
     formValues,
     id,
     ref,
     setFormValues,
+    setValid,
     steps,
     width,
   } = useContext(WizardContext);
 
+  const handleSubmit = (evt) => {
```
taysea, line 20:
```suggestion
  const handleSubmit = (event) => {
```
Small suggestion but just for consistency across our other examples.

aries-site/src/examples/templates/wizard/components/StepContent.js
```diff
@@ -1,19 +1,48 @@
...
+    setValid(true);
+    if (activeIndex < steps.length - 1) {
+      setActiveIndex(activeIndex + 1);
+    }
+    else if (onSubmit) {
+      onSubmit(evt);
```
taysea, line 26:
```suggestion
      onSubmit(event);
```

MikeKingdom:
> > > > Left a few comments re: changes in validation before vs what we have existing. Also, I noticed this console warning. I had a value in "First name" and then clicked "Next":
> > > 
> > > 
> > > I'll take a closer look at those. I'm not quite sure how it's trying to set state during the render(s)
> > 
> > 
> > Any set state related things and or .focus() during the Form onValidation callback seems to cause this. To avoid the issue I put those in timeouts with 0 delay which basically makes them happen immediately after the onValidation chain of calls. Any other ideas welcome!
> 
> Do you think it has to do with the timeout set on this useEffect by chance? This timeout was introduced to solve a different bug but I wonder if it conflicts with what we're trying to do here? https://github.com/grommet/grommet/blob/master/src/js/components/Form/Form.js#L83
> 
> BTW validation behavior looks great now.

Thanks. Great question. I would think that would actually help since the it moves the onValidate out of the render path unless I'm not understanding the reason for the React warning.

aries-site/src/examples/templates/wizard/components/StepContent.js
MikeKingdom: Done

aries-site/src/examples/templates/wizard/components/StepContent.js
MikeKingdom: Done

aries-site/src/examples/templates/wizard/WizardValidationExample.js
```diff
@@ -25,64 +25,34 @@ import {
...
     <>
       <FormField
         label="First Name"
         htmlFor="firstname-validation"
         name="firstname-validation"
-        error={error.firstname}
-        onChange={() =>
-          attemptedAdvance && setError(stepOneValidate(formValues))
-        }
+        required
```
taysea, line 43: Wherever the required Form message gets set, can we set it to "This is a required field." to match Figma/form examples? As opposed to just "requried"

MikeKingdom: Seems overly wordy. Is that the general UX guidance we want to promote?

MikeKingdom: Added the required message like Form does. Thanks for the pointer.

taysea (approved):
LGTM. Great job, nice that we can leverage Grommet's form validation now.

Issue #1118, opened 2020-09-10
Filters - Control Button direction


taysea:
In Filter designs, we are introducing a button pattern of:
<img width="124" alt="Screen Shot 2020-09-10 at 10 11 31 AM" src="https://user-images.githubusercontent.com/12522275/92798780-199ba300-f368-11ea-8c20-4ec81b60b2d8.png">

Currently, this treatment only applies to controls like Select. However, in the case of filters, we want the option for this styling to apply to both Button and DropButton (which relies on button). In the current Filter implementation, this is being done via a styled component, but it feels like we should be able to handle this via a prop and theme styling if possible.

Suggested approach is use `button.extend` to control the theming of this kind of button and introduce a prop that just lives within an application (not on grommet) called control. We could achieve the desired styling as such:
```
button: {
    extend: ({ control, theme }) =>
      css`
        ${control &&
        `border: 1px solid
    ${theme.global.colors.border[theme.dark ? 'dark' : 'light']};
  &:hover {
    background: transparent;
  }`}
      `,
  },
```
and implement via <Button label="Label" control />. I've tried this locally and it works as expected.

TO-DO:
- [ ] Discuss implementation recommendation
- [ ] Discuss choice of `control` as naming
- [ ] Implement on grommet-theme-hpe and update Filter examples

PR #1589, opened 2021-04-27
Replace styled button with toolbar button in theme

britt6612:
#### What does this PR do?
This PR cleans up `filtering` by removing the `styledbutton` and moving that to theme.

For now, `toolbar` has been selected because it conveys information about where this button is typically used. If the usage of this button expands over time, we can always alias it in the theme with a new name but this seems appropriate for the information we have about this type so far.

aries.js
site 
look at deployment
issue #1118 
no
backwards compatible


aries-site/src/examples/templates/filtering/FilteringWithRangeSelector.js
```diff
@@ -235,7 +220,8 @@ const Filters = ({
     return (
       <>
         <Box align="center" direction="row" gap="xsmall">
-          <StyledButton
+          <Button
+            king="tertiary"
```
taysea, line 224:
```suggestion
            kind="tertiary"
```

taysea (commented):
Looks good, will wait on discussion for naming before approving.

aries-site/src/examples/templates/filtering/FilteringWithRangeSelector.js
britt6612: thanks for catching :)

aries-site/src/examples/templates/filtering/FilteringWithSelect.js
```diff
@@ -224,7 +217,8 @@ const Filters = ({
         {size !== 'small' ? (
           content
         ) : (
-          <StyledButton
+          <Button
+            king="toolbar"
```
taysea, line 221:
```suggestion
            kind="toolbar"
```

taysea:
```suggestion
            kind="toolbar"
```

Issue #1590, opened 2021-04-28
Design System site times out and causes blank screen


taysea:
Received feedback that when the DS site is left open for a long time (unsure specifically how long it takes, hours v days) the screen goes blanks and prints this console error

![Screen Shot 2021-04-22 at 6 23 28 PM](https://user-images.githubusercontent.com/12522275/116451756-c2bcde00-a811-11eb-8248-e62549fc97f2.png)

ericsoderberghp:
Hmm... `jquery`, I wonder if has something to do with the HPE global header.

MikeKingdom:
That's definitely the global header. I wonder if it's repeatable or it was a one time thing. Seems like they update it or change it every so often.

halocline:
@MikeKingdom Here is the documentation for the new Global Header Service:


**Documentation**
- [Implementation Guide](https://github.hpe.com/rugby/positano/files/42223/common-hfws-quick-js-guide-1.0.0.pdf)
- [Code generator and test page](https://h50007.www5.hpe.com/hfws-test-pages/common-hfws-test-page.php?version=latest.r&cc_ll=us%2Fen&console_mode=true&hide_country_selector=true&hide_main_nav=true&hide_search=true&hide_cart=true&hide_sign_in=true)
- [Video description](https://hpe.sharepoint.com/teams/Capability%20%20Projects/Shared%20Documents/Forms/AllItems.aspx?id=%2Fteams%2FCapability%20%20Projects%2FShared%20Documents%2FDemo%2DCommonHeaderFooterService%2Emp4&parent=%2Fteams%2FCapability%20%20Projects%2FShared%20Documents&p=true&ct=1618943947549&or=OWA-NT&cid=1b9984ec-b3bc-c941-7997-85fb720d27b1&originalPath=aHR0cHM6Ly9ocGUuc2hhcmVwb2ludC5jb20vOnY6L3QvQ2FwYWJpbGl0eSUyMCUyMFByb2plY3RzL0Vla2xmYUhvQ1haRmp5dUFxc3A5eWdBQlh6VU5vczk1RV8wQUQxdk5RaDNhbGc_cnRpbWU9cHFaQWxTc0UyVWc)

MikeKingdom:
Closing this since I haven't been able to reproduce this since updating the global header in #1598

samuelgoff:
[...]

PR #1598, opened 2021-05-04
Updated to latest HPE global header

MikeKingdom:
#### What does this PR do?
Update the HPE global header to the latest version.

Generated with [HPE Common HFWS test page](https://h50007.www5.hpe.com/hfws-test-pages/common-hfws-test-page.php?version=latest.r&cc_ll=us%2Fen&header_logo=hpe&slim_footer=true&hide_search=true&hide_cart=true&hide_country_selector=true&hide_sign_in=true
)

_document.js

Local yarn start

local yarn start or the CI/CD preview

May be related to #1590 

image

image

No

Backwards compatible


aries-site/src/pages/_document.js
```diff
@@ -36,16 +36,11 @@ export default class MyDocument extends Document {
...
-          <script src="https://h50007.www5.hpe.com/hfws-static/js/framework/jquery/v-3-5-1/jquery.js" />
-          <script src="https://h50007.www5.hpe.com/hfws/us/en/hpe/slim/root?contentType=js&color_scheme=dark" />
-          <style>
-            {`#hpeslh_site-header .hpeslh_container.hpeslh_wide {
-              border: none;
-            }`}
-          </style>
+          <div id="footer" className="footer" />
+          <script src="https://h50007.www5.hpe.com/hfws-static/js/framework/jquery/v-3-6-0/jquery.js" />
+          <script src="https://h50007.www5.hpe.com/hfws/us/en/hpe/latest.r/root?contentType=js&switchToCHF=true&hide_country_selector=true&hide_search=true&hide_cart=true&hide_sign_in=true&header_logo=hpe&slim_footer=true" />
```
halocline, line 43: Do you see any issues with placing these scripts at the end of the body, rather than in the head as the instructions specify?

I've always tried to push 3rd party scripts as far down as I can, like you are doing here...

aries-site/src/components/seo/Meta.js
```diff
@@ -92,10 +92,10 @@ export const Meta = ({ title, description, canonicalUrl, socialImageUrl }) => {
       <meta
         httpEquiv="Content-Security-Policy"
         content="default-src 'self' 'unsafe-eval'; 
-        style-src 'self' *.hpe.com/hfws-static/slim/css/ 'unsafe-inline';
-        connect-src 'self' *.githubusercontent.com/grommet/hpe-design-system/ https://www.google-analytics.com https://www.github.com/grommet/ https://eyes.applitools.com;
+        style-src 'self' *.hpe.com/hfws-static/5/css/ 'unsafe-inline';
+        connect-src 'self' *.githubusercontent.com/grommet/hpe-design-system/ https://www.google-analytics.com https://www.github.com/grommet/ https://eyes.applitools.com *.hpe.com/hpe/api/;
```
halocline, line 96: assuming the `*.hpe.com/hpe/api` is for any of the auth / signin, correct?

halocline (commented):
Looks good to go aside from the build failure.

I think Netlify is throwing up because the yarn.lock file is missing.

aries-site/src/pages/_document.js
MikeKingdom: Same. I did this with the previous one so that the elements are there when the script gets loaded. Not sure it matters here.

aries-site/src/components/seo/Meta.js
MikeKingdom: Yes although we have the sign in set to off but it was still trying to hit the user api.

MikeKingdom:
I thought I added the new yarn.lock but apparently I just staged the deletion :/

halocline:
@MikeKingdom This is good to go by me as soon as we get the e2e test check to pass. I'm not sure if we will need to re-write some of the tests -- seems like the header/footer is being caught for accessibility violations -- and then some of the functional tests aren't passing -- haven't looked into those, yet.

MikeKingdom:
> @MikeKingdom This is good to go by me as soon as we get the e2e test check to pass. I'm not sure if we will need to re-write some of the tests -- seems like the header/footer is being caught for accessibility violations -- and then some of the functional tests aren't passing -- haven't looked into those, yet.

@halocline Thanks Matt. I started looking at the test issues. Yeah, the Axe checker has some issues with the new global header. It did with the previous one also. I'll poke around with the tests to see what's failing. It's a little awkward on windows since they tend to run differently on windows not to mention you have to change the quoting in package.json for them to run at all

MikeKingdom:
@halocline I disabled the keyboard tests where it tries to Tab to a specific link. The way it tries to figure out how many clicks it will take to get a link is highly dependent on what's in the global header. The new global header is way harder to form a selector that weeds out the non-tabbable elements than the previous one, not to mention it's an external set of DOM that could change at anytime. I think a better solution would be to at least just click tab until you reach the element you want to test or maybe start with keyboard focus somewhere below the header.

ericsoderberghp:
> @halocline I disabled the keyboard tests where it tries to Tab to a specific link. The way it tries to figure out how many clicks it will take to get a link is highly dependent on what's in the global header. The new global header is way harder to form a selector that weeds out the non-tabbable elements than the previous one, not to mention it's an external set of DOM that could change at anytime. I think a better solution would be to at least just click tab until you reach the element you want to test or maybe start with keyboard focus somewhere below the header.

Do we have SkipLinks set up to overcome this?

MikeKingdom:
> > @halocline I disabled the keyboard tests where it tries to Tab to a specific link. The way it tries to figure out how many clicks it will take to get a link is highly dependent on what's in the global header. The new global header is way harder to form a selector that weeds out the non-tabbable elements than the previous one, not to mention it's an external set of DOM that could change at anytime. I think a better solution would be to at least just click tab until you reach the element you want to test or maybe start with keyboard focus somewhere below the header.
> 
> Do we have SkipLinks set up to overcome this?

Good question. I don't think so.

ericsoderberghp:
> > > @halocline I disabled the keyboard tests where it tries to Tab to a specific link. The way it tries to figure out how many clicks it will take to get a link is highly dependent on what's in the global header. The new global header is way harder to form a selector that weeds out the non-tabbable elements than the previous one, not to mention it's an external set of DOM that could change at anytime. I think a better solution would be to at least just click tab until you reach the element you want to test or maybe start with keyboard focus somewhere below the header.
> > 
> > 
> > Do we have SkipLinks set up to overcome this?
> 
> Good question. I don't think so.

Would be good add it if we don't

aries-site/src/tests/utils.js
```diff
@@ -19,29 +22,19 @@ export const repeatKeyPress = ClientFunction((key, number) =>
...
-  // If item we are aiming to tab to is at index = 3, that means we have to tab
-  // 4 times to get there.
-  return hrefIndex + 1;
-});
+  do {
+    await t.pressKey('tab');
+    activeElement = await getActiveElement();
+    count += 1;
+    href = activeElement.getAttribute('href');
+    // console.log('ACTIVE', href);
```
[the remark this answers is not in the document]
taysea: I like this approach though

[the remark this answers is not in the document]
MikeKingdom: Done and thanks

PR #1604, opened 2021-05-06
Fix % complete

### LINKED ISSUE #1584: Accessibility Issue: “% Complete” column is graphical, and not readable for a screen reader.
“% Complete” column is graphical, and not readable for a screen reader. 
This % Complete column is using color alone to convey information. If the color had a visual representation as well as color it would pass. Also, think about adding a numeric indication that could also be picked up by the screen reader.

Screen Shot 2021-04-26 at 11 49 23 AM

britt6612:
#### What does this PR do?
Add percentages next to meter for screen readers 
table.js 
The  column was graphical, and not readable for a screen reader
closes #1584

### DIFF
--- a/aries-site/src/examples/components/pagination/PaginationTableExample.js
+++ b/aries-site/src/examples/components/pagination/PaginationTableExample.js
@@ -487,13 +487,15 @@ const columns = [
       </Box>
     ),
     render: datum => (
-      <Box pad={{ vertical: 'xsmall' }}>
+      <Box direction="row" gap="xxsmall" pad={{ vertical: 'xsmall' }}>
         <Meter
+          alignSelf="center"
           values={[{ value: datum.pinned / datum.pinnable, color: 'graph-2' }]}
           max={1}
           thickness="small"
           size="small"
         />
+        <Text>{((datum.pinned / datum.pinnable) * 10).toFixed(0)}%</Text>
       </Box>
     ),
     sortable: false,
--- a/aries-site/src/examples/components/pagination/PaginationTableExample.js
+++ b/aries-site/src/examples/components/pagination/PaginationTableExample.js
@@ -487,14 +487,18 @@ const columns = [
       </Box>
     ),
     render: datum => (
-      <Box direction="row" gap="xxsmall" pad={{ vertical: 'xsmall' }}>
-        <Meter
-          alignSelf="center"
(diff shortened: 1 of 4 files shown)


aries-site/src/examples/components/pagination/PaginationTableExample.js
```diff
@@ -487,13 +487,15 @@ const columns = [
       </Box>
     ),
     render: datum => (
-      <Box pad={{ vertical: 'xsmall' }}>
+      <Box direction="row" gap="xxsmall" pad={{ vertical: 'xsmall' }}>
```
taysea, line 490: For some reason it looks like the rows are taller now than before but I'm not sure why on first glance. Don't think the row height should increase:

BEFORE:
Screen Shot 2021-05-06 at 4 12 37 PM

AFTER:
Screen Shot 2021-05-06 at 4 12 28 PM

aries-site/src/examples/components/pagination/PaginationTableExample.js
```diff
@@ -487,13 +487,15 @@ const columns = [
...
         <Meter
+          alignSelf="center"
           values={[{ value: datum.pinned / datum.pinnable, color: 'graph-2' }]}
           max={1}
           thickness="small"
           size="small"
         />
+        <Text>{((datum.pinned / datum.pinnable) * 10).toFixed(0)}%</Text>
```
taysea, line 498:
```suggestion
        <Text>{((datum.pinned / datum.pinnable) * 100).toFixed(0)}%</Text>
```
Think it should be * 100, this looks like "80%"
Screen Shot 2021-05-06 at 4 13 32 PM

taysea: If so we should adjust in the other examples also

aries-site/src/examples/components/pagination/PaginationTableExample.js
britt6612: oh good catch I didnt even see that let me go back and inspect

britt6612: so it was the `box` around `meter`  had a vertical `pad` so that was also going to the `text` so I closed that `box` before `text`

taysea (approved):
LGTM.

PR #1606, opened 2021-05-07
Add SkipLinks to bypass Header

### LINKED ISSUE #1585: Accessibility Issue: Navigation header buttons cannot be bypassed.
Need to add SkipLink

taysea:
#### What does this PR do?
This PR adds SkipLinks to the site in order to allow keyboard navigation to bypass the header and skip directly to the main content. While the addition of SkipLinks grommet component was very simple, the main adjustment in this PR is the useEffect added in _app which listens for when the route has changed and places focus back on the SkipLinks layer. This is necessary otherwise, NextJS will reset the focus placement to the main section of the application, meaning a keyboard user wouldn't get the SkipLink on their first tab.

Additionally, I uncovered a bug where Grommet is placing the `id` on multiple elements within Layer (which SkipLinks leverages), so I needed to disabled one of the accessibility check rules in order for this to pass. I think this is okay because having the SkipLinks in the site adds more usability benefits.

Tested locally. Open deploy, press "tab" and SkipLinks should appear. Press enter on a link that will route you to a new page. SkipLinks should appear again.

Same as above.

For accessibility with keyboard navigation, it provides a better experience if users can bypass headers/navigation sections and skip to main content as they navigate through a site.

Closes #1585 

No.

Backwards compatible.

### DIFF
--- a/aries-site/src/layouts/main/Layout.js
+++ b/aries-site/src/layouts/main/Layout.js
@@ -75,7 +75,7 @@ export const Layout = ({
                 canonicalUrl={`https://design-system.hpe.design${router.route}`}
               />
               <>
-                <SkipLinks>
+                <SkipLinks id="skip-links">
                   <SkipLink id="main" label="Main Content" />
                 </SkipLinks>
                 <Header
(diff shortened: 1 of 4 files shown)


britt6612:
Nice comments! thanks for adding those. Did you create a grommet issue for this? If so maybe link above

britt6612 (approved):

PR #1616, opened 2021-05-13
Fix csp in dev mode

taysea:
#### What does this PR do?

Turns out our Content-Security-Policy stops 'unsafe-eval' within the react hot reloader when you're running in dev mode, but only on non-windows. I think we avoided that issue before because safari and FF and maybe non-windows chrome ignored script-src-elem The workaround is to change our policy to be more lenient in NODE_ENV !== 'production'. 

aries-site/src/components/seo/Meta.js

Tested in dev mode and prod mode locally for MacOS Chrome, Firefox, and Safari.

Same as above.

No.

Backwards compatible.

### DIFF
--- a/aries-site/src/components/seo/Meta.js
+++ b/aries-site/src/components/seo/Meta.js
@@ -15,12 +15,11 @@ export const Meta = ({ title, description, canonicalUrl, socialImageUrl }) => {
   style-src 'self' *.hpe.com/hfws-static/5/css/ 'unsafe-inline';
   connect-src 'self' *.githubusercontent.com/grommet/hpe-design-system/ https://www.google-analytics.com https://www.github.com/grommet/ https://eyes.applitools.com *.hpe.com/hpe/api/;
   media-src 'self' https://d3hq6blov2iije.cloudfront.net/media/HPE+Design+System-v3.mp4;
-  img-src 'self' https://www.google-analytics.com https://images.unsplash.com/ http://s.gravatar.com/avatar/ *.hpe.com/hfws-static/5/;
-  ${
-    // in dev mode, use script-src-elem, otherwise we run into issues with
-    // react hot reloader
-    process.env.NODE_ENV !== 'production' ? 'script-src-elem' : 'script-src'
-  } 'self' *.hpe.com https://www.google-analytics.com/analytics.js;
+  img-src 'self' data: https://www.google-analytics.com https://images.unsplash.com/ http://s.gravatar.com/avatar/ *.hpe.com/hfws-static/5/;
+  script-src 'self' *.hpe.com https://www.google-analytics.com/analytics.js ${
+    // in dev mode, we allow unsafe-eval to work with react hot reloader
+    process.env.NODE_ENV !== 'production' ? "'unsafe-eval'" : ''
+  };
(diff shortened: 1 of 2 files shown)


aries-site/src/components/seo/Meta.js
```diff
@@ -11,6 +11,18 @@ export const Meta = ({ title, description, canonicalUrl, socialImageUrl }) => {
...
   const pageContent = 'products';
+  const csp = `default-src 'self' 'unsafe-eval'; 
+  style-src 'self' *.hpe.com/hfws-static/5/css/ 'unsafe-inline';
+  connect-src 'self' *.githubusercontent.com/grommet/hpe-design-system/ https://www.google-analytics.com https://www.github.com/grommet/ https://eyes.applitools.com *.hpe.com/hpe/api/;
+  media-src 'self' https://d3hq6blov2iije.cloudfront.net/media/HPE+Design+System-v3.mp4;
+  img-src 'self' https://www.google-analytics.com https://images.unsplash.com/ http://s.gravatar.com/avatar/ *.hpe.com/hfws-static/5/;
+  ${
+    // in dev mode, use script-src-elem, otherwise we run into issues with
+    // react hot reloader
+    process.env.NODE_ENV !== 'production' ? 'script-src-elem' : 'script-src'
```
MikeKingdom, line 22: Doesn't this mean in dev mode the global header doesn't show up in safari and FF? Seems like we might instead remove the 'unsafe-eval' in dev mode?

taysea: Yes I think that's the better approach. matt was still seeing some console errors. going to make an update.

aries-site/src/components/seo/Meta.js
```diff
@@ -11,6 +11,17 @@ export const Meta = ({ title, description, canonicalUrl, socialImageUrl }) => {
...
   const lifecycle = 'support';
   const pageContent = 'products';
+  const csp = `default-src 'self' 'unsafe-eval'; 
+  style-src 'self' *.hpe.com/hfws-static/5/css/ 'unsafe-inline';
+  connect-src 'self' *.githubusercontent.com/grommet/hpe-design-system/ https://www.google-analytics.com https://www.github.com/grommet/ https://eyes.applitools.com *.hpe.com/hpe/api/;
+  media-src 'self' https://d3hq6blov2iije.cloudfront.net/media/HPE+Design+System-v3.mp4;
+  img-src 'self' data: https://www.google-analytics.com https://images.unsplash.com/ http://s.gravatar.com/avatar/ *.hpe.com/hfws-static/5/;
+  script-src 'self' *.hpe.com https://www.google-analytics.com/analytics.js ${
+    // in dev mode, we allow unsafe-eval to work with react hot reloader
+    process.env.NODE_ENV !== 'production' ? "'unsafe-eval'" : ''
```
MikeKingdom, line 21: That seems to be better. Thanks!

MikeKingdom (commented):
LGTM. I'll let someone else double check since it always worked for me.

halocline (approved):
Tested MacOS 10.15.7 across latest FF, Chrome, Safari, and Edge.

Tested both development and production modes.

PR #1617, opened 2021-05-13
Reduced use of 'fill' on home page grids

ericsoderberghp:
#### What does this PR do?

Reduced use of 'fill' on home page grids. Also, changed the Featured Grid columns to go down to `small` instead of `medium` as that seemed to respond better at laptop resolutions, where it shows three cards instead of two over one with the cards being rather wide.

In looking into how we can make some basic page layouts easier to construct. I noticed that there seemed to be a bit too much use of `fill` in the home page. Also Grid didn't need `rows` defined.

anywhere

manually at different resolutions with Safari, Chrome, and Firefox.
TestCafe

home page at different sizes in your browser

no

backwards compatible

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,3 +22,4 @@ yarn-error.log*
 .env
 .tmp
 report*
+.DS_Store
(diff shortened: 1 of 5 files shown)


MikeKingdom (approved):
Looks good. Seems to behave well! You always come up with a simpler solution than I do :)

taysea (approved):
LGTM.

Issue #1348, opened 2021-01-06
Update search Figma file to match filter


vavalos5:
Saw some misalignments between search and filter Figma files. 

<img width="245" alt="Screen Shot 2021-01-06 at 2 36 29 PM" src="https://user-images.githubusercontent.com/62117454/103826899-9f84dd80-502c-11eb-816a-6d69ac6e8122.png">
<img width="509" alt="Screen Shot 2021-01-06 at 2 36 41 PM" src="https://user-images.githubusercontent.com/62117454/103826901-a01d7400-502c-11eb-9646-a0cc084859c7.png">

Spoke with Greg as to which direction we're going with. He confirmed that we want to have search be aligned with filter Figma file so when an item has been searched, the label underneath search bar should read as " # results of x number of items".

Also, another thing Greg mentioned: " one caveat, it can occur where there are 1000's of items and the ability of the backend to bring that forward may be an issue. That’s where we could provide an alternative that  displays, “showing 3 items” only. I think this is a ‘description’ rather than a ‘prescription’  but with a preference"

^^ think we may want to discuss that

**Search Figma file:** https://www.figma.com/file/ChTUhUi0PiDCJ5V6iAYw5p/HPE-Search-Templates?node-id=1%3A9

**Filter Figma file:** https://www.figma.com/file/EA0OZFdocKVcQ55c7Jainb/HPE-Filter-Templates?node-id=1%3A3738

Issue #1347, opened 2021-01-06
Total number of applied filters not shown on filtering page


vavalos5:
The total number of filters on the filtering page aren't shown when more than 1 filter is applied. In Figma we show a circle with the total number of filters applied Please include on site.

Note on this image that there is a circle on filter with total number of filters applied:
<img width="729" alt="Screen Shot 2021-01-06 at 1 57 37 PM" src="https://user-images.githubusercontent.com/62117454/103823658-5a11e180-5027-11eb-89e4-2be439e96a51.png">

Figma page: https://www.figma.com/file/EA0OZFdocKVcQ55c7Jainb/HPE-Filter-Templates?node-id=1%3A3738

HPE Filtering Site: https://design-system.hpe.design/templates/filtering#filtering-with-rangeselector

Issue #1122, opened 2020-09-11
Refactor Filter Examples


taysea:
Clean-up Filter Examples to aid readability since these examples are fairly dense. Functionality shouldn't be changing, just trying to simplify the code where possible.

ericsoderberghp:
No recent activity, closing for now

halocline:
I have just recently completed a decent chunk of work for Positano regarding searching/filtering which can/should be ported over and applied to our site examples.

When trying to consume our site examples and apply them to Positano, I uncovered a number of items we can smooth out and improve the developer experience. I've been sharing with @taysea along the way. This, combined with our new(er) file-based code examples (implemented on Wizard) would complete the work represented by this issue.

@ericsoderberghp I'd recommend we re-open this issue as a good chunk of the work has been done --- I also see this as being a highly "trafficked" template / pattern as more teams look to implement.

PR #1619, opened 2021-05-18
Refactor Filtering examples to make easier 

halocline:
#### What does this PR do?

Closes https://github.com/grommet/hpe-design-system/issues/1122
Closes https://github.com/grommet/hpe-design-system/issues/1347
Closes https://github.com/grommet/hpe-design-system/issues/1348

**DRAFT Status**

Refactors `FilteringCards.js` to make use of a suite of reusable Filtering components. Purpose is to make it easier for other projects to consume the filtering examples more easily and minimize re-factoring needed to implement filtering in their project.

FilteringCards2.js

**ToDos**
- [ ] Make filter layer target example container
- [ ] Add TestCafe fixture and suite of tests to ensure filtering works as expected
- [ ] Apply FilterControls to remaining examples


aries-site/src/examples/templates/FilterControls/FilterContext.js
```diff
@@ -0,0 +1,124 @@
...
+  const [data, setData] = useState([]);
+  // fields or columns in the data set
+  const [filterAttributes, setFilterAttributes] = useState([]);
+  // filterAttributes and associated values for which the data is to be filtered
+  const [filters, setFilters] = useState({});
+  // show/hide the layer presenting available filters
+  const [filtersLayer, setFiltersLayer] = useState(false);
+  // filteredResults are the subset of data which meets the filters criteria
+  const [filteredResults, setFilteredResults] = useState([]);
+  const [isFiltered, setIsFiltered] = useState(false);
```
taysea, line 16: would it be safe to say if `filterResults.length` then `isFiltered` would be true? Wondering if there's opportunity to eliminate a state variable

taysea: maybe taking back the statement actually because filtering could result in 0 matches

aries-site/src/examples/templates/FilterControls/Filters.js
```diff
@@ -0,0 +1,107 @@
...
+    setPreviousFilters,
+    setSearchValue,
+  } = useContext(FilterContext);
+  const [filterCount, setFilterCount] = useState();
+
+  // Provide indication for the number of filters applied
+  useEffect(() => {
+    let count = 0;
+
+    if (!filtersLayer) {
```
taysea, line 52: is this condition to make sure if doesn't update until the filtersLayer closes? if so, adding a comment would be helpful

aries-site/src/examples/templates/FilterControls/FilterCheckBoxGroup.js
```diff
@@ -0,0 +1,54 @@
...
+    });
+  if (convertToString) options = options.map(option => option.toString());
+
+  useEffect(() => {
+    setValue(filters[property] || []);
+  }, [filters, property, setValue]);
+
+  return (
+    <Box flex={false}>
+      <FormField htmlFor={property} name={property} label={header}>
```
taysea, line 29: Any opportunity to change `header` --> `label` ? Seems a little misleading since header is typically a layout element

aries-site/src/examples/templates/filtering/FilteringCards2/FilteringCards2.js
```diff
@@ -0,0 +1,158 @@
...
+    filtersLayer,
+    getFilteredResults,
+    searchValue,
+  } = useContext(FilterContext);
+
+  /* Configure which attributes should be made available for the user 
+     to filter upon */
+  useEffect(() => {
+    const attributes = [
+      { property: 'role', header: 'Role', filterType: 'checkboxgroup' },
```
taysea, line 60: same as comment above, could we change `header` to `label` ?

taysea, line 60: Any reason to have the component all lowercase as opposed to camel case or matching the component name "CheckBoxGroup"?

aries-site/src/examples/templates/FilterControls/FilterContext.js
```diff
@@ -0,0 +1,124 @@
+// Copyright 2021 - Hewlett Packard Enterprise Company
+
+import { createContext, useState } from 'react';
+
+const useFilter = () => {
```
taysea, line 5: Any reason to call is useFilter here but then export as FilterContext below? Seems like we could just call this FilterContext

aries-site/src/examples/templates/FilterControls/FilterRangeSelector.js
```diff
@@ -0,0 +1,54 @@
...
+
+  return (
+    <Box flex={false}>
+      <FormField htmlFor={property} name={property} label={header}>
+        <Stack>
+          <Box background="border" height="3px" direction="row" />
+          <RangeSelector
+            id={property}
+            name={property}
+            min={inputProps.min || 0}
```
taysea, line 28: Should we spread any other inputProps below, in case user has provided other's that we're not specifically accounting for?
{...inputProps}

aries-site/src/examples/templates/FilterControls/Filters.js
```diff
@@ -0,0 +1,107 @@
...
+
+const FilterButton = styled(Button)`
+  border: 1px solid
+    ${({ theme }) => theme.global.colors.border[theme.dark ? 'dark' : 'light']};
+  &:hover {
+    background: transparent;
+  }
+`;
+
+const ClearFiltersButton = styled(Button)`
```
taysea, line 19: With the recent theme discussions on the above "Filter Button", do you think this styled Button that looks like anchor should be considered as a kind?

aries-site/src/examples/templates/FilterControls/Filters.js
```diff
@@ -0,0 +1,107 @@
...
+    }
+  }, [filters, filtersLayer, setFilterCount]);
+
+  return (
+    <>
+      <Box direction="row" align="center" gap="xsmall">
+        <Stack anchor="top-right">
+          <Box pad={{ top: 'small', right: 'small' }}>
+            <FilterButton
+              icon={<Filter />}
```
taysea, line 68:
```suggestion
              badge={isFiltered ? filterCount : undefined}
              icon={<Filter />}
```
We should be able to do something like this now and remove all the stack stuff

taysea (commented):
I'm liking this direction vs what we currently have. The context helps a lot. I think something I'm thinking about is there are still lots of files to consume and some things that are fairly critical to get right such as the attributes setup and then building the specialized filter version of that component. I'm thinking this would be an important candidate for the step-by-step: "User needs to filter dataset in order to perform task" and then walk through some of the key elements here.

aries-site/src/examples/templates/FilterControls/FilterContext.js
halocline: You are right, there is a case where there are filters applied, but no results due to the criteria applied.

aries-site/src/examples/templates/FilterControls/Filters.js
halocline: Adding.

aries-site/src/examples/templates/FilterControls/FilterCheckBoxGroup.js
halocline: yes, good call. This naming carried through from a table column definition use case, but doesn't make sense here. Updating.

aries-site/src/examples/templates/filtering/FilteringCards2/FilteringCards2.js
halocline: I'll change to matching component name.

halocline:
> I'm liking this direction vs what we currently have. The context helps a lot. I think something I'm thinking about is there are still lots of files to consume and some things that are fairly critical to get right such as the attributes setup and then building the specialized filter version of that component. I'm thinking this would be an important candidate for the step-by-step: "User needs to filter dataset in order to perform task" and then walk through some of the key elements here.

Agree. I think we should brainstorm on how we might make the "configuration" interface more user friendly.

aries-site/src/examples/templates/FilterControls/Filters.js
halocline: I'd like to discuss this further. I definitely think it could be a candidate. I'd also like to spend time thinking of any cons associated as well.

I have seen designs which are using anchor styling, but function more like an action button (e.g. "clear filters," and circumstances like help text which opens a side drawer with help content.

aries-site/src/examples/templates/FilterControls/FilterRangeSelector.js
halocline: good call - I like

aries-site/src/examples/templates/FilterControls/FilterContext.js
halocline: Re-worked this. Introduced a useFilters() hook and a FiltersProvider. 

From a caller point of view, they now only have to:
```
const filtersConfig = [
    { property: 'role', label: 'Role', filterType: 'CheckBoxGroup' },
    { property: 'status', label: 'Status', filterType: 'CheckBoxGroup' },
    {
      property: 'location',
      label: 'Location',
      filterType: 'CheckBoxGroup',
    },
    {
      property: 'hoursAvailable',
      label: 'Remaining Hours Available',
      filterType: 'RangeSelector',
      inputProps: {
        min: 0,
        max: 40,
        valueRange: '0 - 40 hours',
      },
    },
    { property: 'name', label: 'Name', filterType: 'CheckBoxGroup' },
  ];

  return (
    <>
      <Heading level={2} margin={{ bottom: 'small', top: 'none' }}>
        Users
      </Heading>
      <FiltersProvider>
        <FilterControls data={users} filters={filtersConfig} />
        <Users />
      </FiltersProvider>
    </>
  );
```

The rest of the work is handled by the FilterControls components.

aries-site/src/examples/templates/FilterControls/filterTypes/FilterCheckBoxGroup.js
```diff
@@ -0,0 +1,54 @@
+import { useEffect, useState } from 'react';
+import PropTypes from 'prop-types';
+import { Box, CheckBoxGroup, FormField } from 'grommet';
+
+import { useFilters } from '..';
+
+export const FilterCheckBoxGroup = ({ attr }) => {
+  const { data, filters, setFilters, getFilterOptions } = useFilters();
+  const { property, label, inputProps, sort = true, convertToString } = attr;
```
britt6612, line 9: Not sure if I'm following why and where these are coming from? Are they from the styled label?

britt6612:
> > I'm liking this direction vs what we currently have. The context helps a lot. I think something I'm thinking about is there are still lots of files to consume and some things that are fairly critical to get right such as the attributes setup and then building the specialized filter version of that component. I'm thinking this would be an important candidate for the step-by-step: "User needs to filter dataset in order to perform task" and then walk through some of the key elements here.
> 
> Agree. I think we should brainstorm on how we might make the "configuration" interface more user friendly.

I agree with you both.. This was definitely a lot for myself going through this. When I first saw the files even knowing where the `context` was being defined and set so where in these files should I look first... going through  a step by step can clear some of that up.. Once I figured out how everything was connected it was easier to comprehend. I think to clean up the files even more we should decide on what to do with those `styled-buttons` we don't want to have other teams use those. Overall nice work Matt. :)

aries-site/src/examples/templates/FilterControls/filterTypes/FilterCheckBoxGroup.js
halocline: Good question. If I understand it correctly, this could/should be documented/commented.

A caller's data might have a large number of columns, however, from a filtering perspective they might want to limit/focus which of those columns should be presented as a filter.

The proposal is for the caller to specify which data attributes in their data set by which they'd like to filter. They would specify the filter configuration and pass it to a `filters` prop: `<FilterControls filters={filtersConfig} />`. The configuration properties would be:
- property = field/column of interest in the data set
- label = user facing label to present as the filter's FormField label
- filterType = type of input component to use as the filter
- inputProps = any props supported by the component specified by the filterType
- sort = boolean for whether the filter options should be sorted (e.g. options in a checkboxgroup or select)
- convertToString = present numeric values in a CheckBoxGroup (perhaps should be named numericToString)

aries-site/src/examples/templates/FilterControls/SearchFilter.js
```diff
@@ -0,0 +1,65 @@
...
+    ${({ theme }) => theme.global.colors.border[theme.dark ? 'dark' : 'light']};
+  &:hover {
+    background: transparent;
+  }
+`;
+
+export const SearchFilter = () => {
+  const size = useContext(ResponsiveContext);
+  const filterContext = useFilters();
+  const {
```
taysea, line 22: any reason we can't just destructure directly out from `useFilters`?

PR #1623, opened 2021-05-24
Update icons.mdx

ecircenis:
We have too many icons in grommet and many of them are very obscure.  We really ought to be defining a core set of icons that can be used without needing any textual assistance. Barring that, we should give better guidance (as in this request) about only using commonly understood icons and discouraging use of obscure icons.

### DIFF
--- a/aries-site/src/pages/foundation/icons.mdx
+++ b/aries-site/src/pages/foundation/icons.mdx
@@ -8,7 +8,7 @@ import {
   IconTextExample,
 } from '../../examples';
 
-Icons can be used to represent information to users. such as actions a user can perform and represent different types of functionality.
+Icons can be used to represent information to users, such as actions a user can perform or different types of functionality.
 
 It is important that the use of an icon is clear. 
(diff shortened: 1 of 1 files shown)


aries-site/src/pages/foundation/icons.mdx
```diff
@@ -8,8 +8,11 @@ import {
   IconTextExample,
 } from '../../examples';
 
-Icons can be used to represent information to users. such as actions a user can perform and represent different types of functionality. It is important that the use of an icon is clear. In some cases, it may be beneficial to pair icons with text to ensure the use of the
-icon is properly understood.
+Icons can be used to represent information to users. such as actions a user can perform and represent different types of functionality.
```
taysea, line 11:
```suggestion
Icons can be used to represent information to users, such as actions a user can perform or different types of functionality.
```

taysea:
Handling this PR in https://github.com/grommet/hpe-design-system/pull/1624 bc of yarn conflict. I've migrated all the copy changes over.

PR #1635, opened 2021-06-02
Refactor table customization with useFilter

MikeKingdom:
#### What does this PR do?
Re-factor the table customization template example to use the `<FilterControls>` and `useFilter()` from the Filtering template example.

TableCustomizationExample.js


aries-site/src/examples/templates/table-customization/components/TableCustomizationExample.js
```diff
@@ -126,109 +97,59 @@ export const TableCustomizationExample = () => {
...
+const Results = ({ columns }) => {
   const [select, setSelect] = useState([]);
+  const { filteredResults } = useFilters();
 
   return (
-    <Box fill="vertical" overflow="auto">
+    <Box fill overflow="auto">
       <DataTable
-        data={data}
+        data={filteredResults}
```
halocline, line 142: In fullscreen, dark mode:

Screen Shot 2021-06-03 at 11 22 04 AM

We could do:

```suggestion
        background="background"
        data={filteredResults}
```

Or:

```suggestion
        background="background-front"
        data={filteredResults}
```
And add `background="background-front"`on Line 94-ish.

halocline (commented):
One minor tweak to background color. Everything else looks great. Plus eliminating 700 lines of code is not too shabby.

@taysea, we will want to incorporate your <SelectionSummary /> implementation for Table into this example once it is ready.

aries-site/src/examples/templates/table-customization/components/TableCustomizationExample.js
MikeKingdom: why the contrasting color for the table? That looks odd and distracting to me, especially with the checkbox column background being different

MikeKingdom: oh or you are arguing to change the example background to `background` so it has contrast in dark mode but then you also have to change the DataTable since the general background bleeds through into that?

halocline: The screenshot is how it currently renders when in fullscreen dark mode (which doesn't look correct).

Trying to address so that it looks correct. I don't have a preference towards either approach, just as long as the background is uniform across the entire table.

halocline: > but then you also have to change the DataTable since the general background bleeds through into that?

Right. We can either just apply "background" to the DataTable and all looks good. Or we could apply "background-front" to the DataTable, but I believe we would also need to apply "background-front" to the example Box as well to make things look correct.

I am good with either approach.

MikeKingdom: The whole thing is lighter color for me. Not sure why the difference (safari?) but I'll do the above.

taysea (approved):
LGTM. We can incorporate my table selection work once I have it done, but don't think it should block merging this PR.

Issue #1631, opened 2021-05-28
Filtering - Apply FilterControl to remaining filtering examples


halocline:
Apply the new FilterControls module to the following filtering page examples:

- [x] https://design-system.hpe.design/templates/filtering#filters-in-a-center-layer
- [x] https://design-system.hpe.design/templates/filtering#filtering-with-rangeselector (created sub-issue https://github.com/grommet/hpe-design-system/issues/1672)
- [x] https://design-system.hpe.design/templates/filtering#filtering-with-select (https://github.com/grommet/hpe-design-system/issues/1673)

PR #1636, opened 2021-06-02
Filtering refactor - Center Layer Example

halocline:
#### What does this PR do?

- Implements the refactor of the center layer filter example from https://github.com/grommet/hpe-design-system/issues/1631
- Extends useFilters() and FiltersLayer.js with `layerProps` to allow for customizing the Layer
- Extends Filter inputs to allow [`contentProps`](https://v2.grommet.io/formfield#contentProps) to be passed to containing FormField
- Minor cleanups for FilteringCards.js

[FilteringLayer.js Preview](https://deploy-preview-1636--keen-mayer-a86c8b.netlify.app/templates/filtering#filters-in-a-center-layer)

Manual

[FilteringLayer.js Preview](https://deploy-preview-1636--keen-mayer-a86c8b.netlify.app/templates/filtering#filters-in-a-center-layer)

https://github.com/grommet/hpe-design-system/issues/1631

https://github.com/grommet/hpe-design-system/issues/1631

Yes.

Backwards.


aries-site/src/examples/templates/FilterControls/FiltersLayer.js
```diff
@@ -35,21 +38,22 @@ export const FiltersLayer = () => {
...
         pad={{ vertical: 'medium' }}
         width={{ min: size !== 'small' ? 'medium' : undefined }}
+        {...containerProps}
       >
         <Header pad={{ horizontal: 'medium' }}>
           <Heading margin="none">Filters</Heading>
           <Button icon={<FormClose />} onClick={() => closeLayer()} />
         </Header>
-        <Box pad="medium" overflow="auto" flex>
+        <Box pad="medium" overflow="auto" flex {...contentProps}>
```
taysea, line 56: Should content props be here? I also see it on the contained FormField which makes more sense to me

taysea: Or vice versa, but it was confusing to me that it's on both

halocline: I can see the confusion. The same naming is coming from two different instances and have two different targets.

In this instance, `contentProps` is coming from the `layerProps` object. The use in the FormField comes from the `filtersConfig` array objects, but also uses the name `contentProps` which is how the Grommet docs for FormField refer to it.

Open to suggestions.

taysea: Got it, thanks for clarifying. Skimming the code I got a little lost. I like the idea of keeping with Grommet convention for FormField. For it be too verbose to use `formFieldContainerProps` or `inputsContainer` or something like this?

halocline: Adjusted the name so that the main content Box uses `mainProps`. Does that work any better for you?

halocline: After discussion, reverting back to `contentProps`.

aries-site/src/examples/templates/filtering/FilteringLists/FilteringLists.js
```diff
@@ -0,0 +1,139 @@
...
+    // containing the layer's header, main content, and footer
+    containerProps: {
+      fill: false,
+      pad: { horizontal: 'medium', vertical: 'small' },
+      width: 'large',
+    },
+    // contentProps accept any Grommet Box props to customize the Box
+    // containing the FormField filters.
+    contentProps: {
+      direction: 'row',
```
taysea, line 62: Think we'll need to use responsive context so this can be aligned in a column on mobile:
Screen Shot 2021-06-03 at 11 16 27 AM

taysea (commented):
Just one small comment in the code.

Re: snapshot changes, it looks like the global header injected something that didn't render its styles in time for when the snapshot was taken. Once we make the code adjustment and the snapshots re-run, let's see if the snapshots resolve themselves.

taysea (commented):
Only thing I noted is that we'll want to improve them mobile layout when search + clear filters button is present. I am comfortable if you'd rather take that in a separate PR as an enhancement to FilterControls.

Screen Shot 2021-06-16 at 4 06 33 PM

Other than that looks good.

PR #1637, opened 2021-06-02
Use grommet singleId prop

taysea:
#### What does this PR do?
Leveraging grommet's new `options` prop to only apply a single id to layer. This allows use to enable the accessibility checks we had previously had to disable.

aries-site/src/layouts/main/ThemeMode.js

All automatic tests pass -- we have an `id` on SkipLinks which is built with Layer and all is working as expected.

Tab when page loads and SkipLinks should appear.

Backwards compatible.

### DIFF
--- a/aries-site/src/layouts/main/ThemeMode.js
+++ b/aries-site/src/layouts/main/ThemeMode.js
@@ -16,6 +16,11 @@ export const ThemeMode = ({ children, ...rest }) => {
   const body = (
     <Grommet
       background="background-back"
+      options={{
+        layer: {
+          singleId: true,
+        },
+      }}
       theme={aries}
       themeMode={themeMode}
       full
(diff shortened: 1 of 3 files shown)


halocline (approved):

halocline:
@taysea Can you checkout the one remaining snapshot change? I approved the others due to Global Header Service, but this one looks like a card height change and I'm not sure where that is coming from.

taysea:
> @taysea Can you checkout the one remaining snapshot change? I approved the others due to Global Header Service, but this one looks like a card height change and I'm not sure where that is coming from.

Yes, let's keep an eye on the next PR that comes up with snapshots. I've noticed upon occasion Applitools will take a screenshot potentially when the layout is still rendering and the height for a Card in a Grid hasn't fully "gotten into place" -- if we see the same snapshot flagged in future let's look deeper into it

PR #1639, opened 2021-06-03
Update links for filter context

taysea:
#### What does this PR do?
Fixing links to FiltersContext and FIlterControls that were resulting in 404 errors on the site.


halocline (approved):
Thank you @taysea , I had included those in my outstanding PR, but apparently was too slow for the demand!

PR #1647, opened 2021-06-07
Use default button in filtering

### LINKED ISSUE #1642: Replace anchor button with a default button in Filtering examples
Screen Shot 2021-06-04 at 11 52 25 AM
After speaking with the CCS team, they pointed out how we're using an anchor rather than a default button when filtering. They bring up a valid point that we advise to use an anchor for inline text links and to point to other resources rather than to clear a filter.

Please change in both:

- [ ] Design System Site
- [ ] Figma

taysea:
#### What does this PR do?
Button should be used in favor of Anchor for interactions that don't involve navigation. Updating to use default button instead of button that appears like anchor.

aries-site/src/examples/templates/FilterControls/Filters.js

Local in filtering example.

Filter something and look at "Clear Filters" button

Closes #1642 

Should inform teams we are working with about this new pattern since it will not update automatically in their apps.

### DIFF
--- a/aries-site/src/examples/templates/FilterControls/Filters.js
+++ b/aries-site/src/examples/templates/FilterControls/Filters.js
@@ -1,25 +1,9 @@
 import { useContext, useEffect, useState } from 'react';
-import styled from 'styled-components';
 import { Box, Button, ThemeContext } from 'grommet';
 import { Filter } from 'grommet-icons';
 
 import { useFilters } from '.';
 import { FiltersLayer } from './FiltersLayer';
-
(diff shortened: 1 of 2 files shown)


halocline (approved):
I am fine with this if this is the design direction in which we are headed. Also, confirming that this work will be coordinated with Figma updates as well, correct?

taysea:
> I am fine with this if this is the design direction in which we are headed. Also, confirming that this work will be coordinated with Figma updates as well, correct?

Yes, this will result in Figma changes.

PR #1648, opened 2021-06-07
Update wizard footer button

### LINKED ISSUE #1643: Align buttons in Wizard to align to desktop view when in mobile
In the Design System site and Figma, we have the button taking up the available real estate within mobile view. Let's align the buttons in mobile view to what they would look like in desktop view.

Here's what it looks like on the design system site: 
Screen Shot 2021-06-04 at 1 08 31 PM

Here's what it looks like on Figma:
Screen Shot 2021-06-04 at 1 08 08 PM

It should look like this:
Screen Shot 2021-06-04 at 1 14 43 PM

- [ ] Fix Design System Site
- [ ] Fix in Figma

taysea:
#### What does this PR do?
Wizard designs are updating so that Next button does not take up full width on mobile, instead it will justify to the end like it does on mobile.

aries-site/src/examples/templates/wizard/components/StepFooter.js

Tested local, see screenshot
Closes #1643 

Screen Shot 2021-06-07 at 11 03 50 AM

Should update teams that this is changing, since it won't automatically change in their apps.


halocline (approved):

aries-site/src/tests/accessibility/axe.js
```diff
@@ -25,7 +25,10 @@ test('should check Home page', async t => {
   if (t.browser.name === 'Chrome') {
     const axeContext = {
       // excluding the HPE Slim header that gets a complaint about link text
-      exclude: ['.hpehf-close-btn'],
+      exclude: [
+        '.hpehf-close-btn',
+        '.hpehf-help-and-support > .hpehf-help-menu-dropdown-list',
```
taysea, line 30: Need to look into this further -- tried excluding these elements from global header bc it's causing accessibility failure but it didn't seem to resolve.

Issue #1676, opened 2021-06-22
Navigate Design System site + Figma and find misalignments/bugs


vavalos5:
Go into the design system site and play around with the mouse and keyboard interactions to see if anything doesn't seem right. Go into Figma files and check if everything looks aligned.

taysea:
## Dev changes
- [ ] Change accordion border color to "border" in theme
- [ ] DateInput uses `Close` icon on site, but Figma calls for `FormClose`

## Design changes
- [ ] DateInput is a heading and should be `text-strong`, Figma is using `text` (should change to `text-strong`)
- [ ] Menu has no "selected" state, should remove
- [ ] Table designs showing Data Table with Search Value need to be updated to match the filtering styling, currently using wrong/outdated text colors

Clarifications:
- [ ] Select in Figma uses "Form value" font-weight -- is this intended?
- [ ] Need to revisit this pattern in Table: https://www.figma.com/file/XEoIWIlLfcARRIFhkpFCch/HPE-Table?node-id=2390%3A13926

ericsoderberghp:
Is there any reason not to add the "Batch Actions" in Figma to the design system site?

taysea:
Good question. I think we should if this is a pattern that is finalized. For some reason, I was under the impression that it was up for discussion? Which is why I listed it under "Clarifications". If this pattern is final, we will also need to update the Filtering examples to match.

vavalos5:
I too somewhat recall that we were going to discuss "Batch Actions" again. 🤔 But if we're ok with the approach we've got, then it is finalized in Figma

ericsoderberghp:
I was going off the fact that the Batch Actions are published to the "Finished" page in Figma. I think the proposal is a reasonable starting point. I could see publishing it in an "out for review" context, with a request for feedback.

taysea:
Closing this issue since related issues have been opened to the remaining to-do items.

PR #1714, opened 2021-07-06
Exploration/Feedback Requested — Show updated batch action behavior

taysea:
[Preview](https://deploy-preview-1714--keen-mayer-a86c8b.netlify.app/templates/filtering#filtering-with-selectable-results)

#### What does this PR do?
Demonstrating a version of batch actions to align with what is shown in [Figma](https://www.figma.com/file/XEoIWIlLfcARRIFhkpFCch/HPE-Table?node-id=2390%3A13926). Instead of having the search input and filter button persistent, instead when selections have been made those elements are replaced by a grey selections summary box.

Some considerations in my own experimentation with this pattern. What if I'm trying to use search as a mechanism to narrow in on results, make a selection, make a new search, add to my selections? With this pattern that is not possible. I am not familiar with users' typical searching/selecting behavior so maybe this should not be of concern, but figured I'd raise it as it was something I experienced when testing my work.

Related to #1676

### DIFF
--- a/aries-site/src/examples/templates/FilterControls/FilterControls.js
+++ b/aries-site/src/examples/templates/FilterControls/FilterControls.js
@@ -42,28 +42,32 @@ export const FilterControls = ({
     syncFilteredResults();
   }, [syncFilteredResults]);
 
-  return !selected.length ? (
-    <Box gap="xsmall" flex={false}>
-      <Box direction="row" align="end" gap="small">
-        {searchFilter && (
-          <SearchFilter placeholder={searchFilter.placeholder} />
-        )}
-        <Filters />
-      </Box>
-      <ResultsSummary />
-    </Box>
-  ) : (
-    <Box
-      direction="row"
-      background="background-back"
-      fill="horizontal"
-      flex={false}
-      pad={{ horizontal: 'medium', vertical: 'xsmall' }}
-      justify="between"
-      wrap
-    >
-      <ResultsSummary />
-      {actions}
+  return (
+    <Box height={{ min: 'xsmall' }} fill="horizontal" flex={false}>
+      {!selected.length ? (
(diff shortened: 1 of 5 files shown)


ericsoderberghp:
> Some considerations in my own experimentation with this pattern. What if I'm trying to use search as a mechanism to narrow in on results, make a selection, make a new search, add to my selections? With this pattern that is not possible. I am not familiar with users' typical searching/selecting behavior so maybe this should not be of concern, but figured I'd raise it as it was something I experienced when testing my work.

Good comments. I seem to recall we felt that it was perilous to leave the search and filter active as it would be ambiguous what would happen to items selected that were no longer in the new search/filter. Would they still be selected even though not visible? How would the user know that? I think we were shooting for what seemed like and 80% case where the user could filter+search to get the set they wanted and then perform and action. This might be good guidance to add, as something to get feedback on.

ericsoderberghp:
What about exploring a "feedback requested" or "actively exploring" annotation to try to highlight that we want to gather reactions?

taysea:
Feedback from the HPE community in Slack: https://hpe.slack.com/archives/C04LMJWQT/p1625681947280800

## Areas of concern/Points of consideration
- **Action button(s) alignment:** Many folks are interested in the action buttons being left aligned. Because the main "next step" of selecting data is to take action against those selections, the feeling is that having the action buttons left aligned will make the buttons more visible, easier to access. The selection count would become a subtext below the actions and the "clear selection" button would come at the end of any actions.
- **What happens to persistent actions that appeared above the table when this banner appears?** Do they become part of the banner?  It seems that other UIs such as Gmail will hide persistent actions once a selection is made.
- **What is the 80/20 scenario for people searching/filtering, selecting, and then searching/filtering again?**

taysea:
Closing for now, not currently relevant.

PR #1716, opened 2021-07-07
Update reordering columns icon - Table Customization

### LINKED ISSUE #1707: Update Table re-ordering icon and position
Based on feedback, the "pencil" icon is not clear that it will be editing the table configuration. also, because it is affecting the view/data in the table, it should be placed next to search/filter which do similar functions:
Screen Shot 2021-07-02 at 10 45 49 AM

taysea:
#### What does this PR do?
Upon discussion and feedback, we have chosen to change the icon to the "Configure" icon.

aries-site/src/examples/templates/table-customization/components/TableCustomizationExample.js

Locally.

Look at deploy.

Closes #1707 

Changes icon/button placement in pattern.

### DIFF
--- a/aries-site/src/examples/templates/FilterControls/FilterControls.js
+++ b/aries-site/src/examples/templates/FilterControls/FilterControls.js
@@ -42,7 +42,7 @@ export const FilterControls = ({
 
   return (
     <Box gap="xsmall" flex={false}>
-      <Box direction="row" align="end" gap="small">
+      <Box direction="row" align="start" gap="small">
         {searchFilter && (
           <SearchFilter placeholder={searchFilter.placeholder} />
         )}
--- a/aries-site/src/examples/templates/FilterControls/FilterControls.js
+++ b/aries-site/src/examples/templates/FilterControls/FilterControls.js
@@ -1,6 +1,6 @@
-import { useEffect } from 'react';
+import { useContext, useEffect } from 'react';
 import PropTypes from 'prop-types';
-import { Box } from 'grommet';
+import { Box, ResponsiveContext } from 'grommet';
 
 import { Filters, ResultsSummary, SearchFilter, useFilters } from '.';
 
@@ -19,6 +19,7 @@ export const FilterControls = ({
     setPrimaryKey,
     syncFilteredResults,
   } = useFilters();
+  const size = useContext(ResponsiveContext);
 
   useEffect(() => {
     setData(data);
@@ -43,7 +44,13 @@ export const FilterControls = ({
 
   return (
     <Box flex={false} fill="horizontal">
-      <Box direction="row" align="start" justify="between" wrap>
+      <Box
+        direction="row"
+        align="start"
+        justify="between"
+        gap="small"
+        wrap={size === 'small'} // so search input has room to grow on mobile
+      >
         <Box
           direction="row"
           align="start"
--- a/aries-site/src/examples/templates/FilterControls/FilterControls.js
+++ b/aries-site/src/examples/templates/FilterControls/FilterControls.js
@@ -5,6 +5,7 @@ import { Box } from 'grommet';
 import { Filters, ResultsSummary, SearchFilter, useFilters } from '.';
 
 export const FilterControls = ({
+  actions,
   data,
   filters,
   layerProps,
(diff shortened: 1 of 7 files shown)


ericsoderberghp:
hmm....
I'm wondering whether that icon is right. What % people who have never interacted with it before do you think will know what it does? I'm wondering if "Configure" or "Settings" would be better, or maybe even "Compare" or "Transaction" might work.

taysea:
>  I'm wondering if "Configure" or "Settings" would be better, or maybe even "Compare" or "Transaction" might work.

I have used "Configure" and added tooltip text of "Configure columns"

aries-site/src/examples/templates/table-customization/components/TableCustomizationExample.js
```diff
@@ -100,38 +100,42 @@ export const TableCustomizationExample = () => {
...
+                    onOpen={() => setOpen(true)}
+                    dropContent={
+                      <ColumnSettings
+                        columns={COLUMNS}
+                        visibleColumns={visibleColumns}
+                        setVisibleColumns={setVisibleColumns}
+                        open={open}
+                      />
+                    }
+                    tip="Configure columns"
```
ericsoderberghp, line 130:
```suggestion
                    tip="configure columns"
```

ericsoderberghp: to align more with "open filters"

taysea: Wondering if both should be capitalized? "Open filters" and "Configure columns"?

ericsoderberghp: https://github.com/grommet/grommet/blob/master/src/js/languages/default.json has mostly lowercase. I do think it would be good to be consistent either way.

taysea: Some of those are only read allowed by screen reader but not visible in the UI, where as these tips are visible when the user hovers over the button. It feels that since the tooltip in the UI is a message, it would make sense to sentence case them.

ericsoderberghp: works for me

ericsoderberghp:
Screen Shot 2021-07-13 at 9 48 11 AM

responsive layout issue

ericsoderberghp:
Screen Shot 2021-07-14 at 9 44 36 AM

I'm wondering if we should flex the Box containing the search input to keep the controls on the same line, instead of wrapping.

ericsoderberghp:
Also, I seem to recall someone wondering whether the control to edit the table columns should be on the left, near the filter control, since it's a bit closer in action to changing what the table shows as opposed to taking action on the contents of the table. Thoughts?

taysea:
> Also, I seem to recall someone wondering whether the control to edit the table columns should be on the left, near the filter control, since it's a bit closer in action to changing what the table shows as opposed to taking action on the contents of the table. Thoughts?

I've added a comment to the code around why we are keeping it right-aligned for now with the other actions.

taysea:
> I'm wondering if we should flex the Box containing the search input to keep the controls on the same line, instead of wrapping.

This will be the mobile behavior. You have to click "Mobile" to see the full responsive behavior since the examples are designed to always show the selected screen size layout.

Before search is clicked:
Screen Shot 2021-07-15 at 2 30 30 PM

After search is clicked:
Screen Shot 2021-07-15 at 2 30 36 PM

aries-site/src/examples/templates/FilterControls/FilterControls.js
```diff
@@ -41,19 +42,28 @@ export const FilterControls = ({
...
 
   return (
-    <Box gap="xsmall" flex={false}>
-      <Box direction="row" align="end" gap="small">
-        {searchFilter && (
-          <SearchFilter placeholder={searchFilter.placeholder} />
-        )}
-        <Filters />
+    <Box flex={false} fill="horizontal">
+      <Box direction="row" align="start" justify="between" wrap>
```
ericsoderberghp, line 46: I still find the `wrap` to not look so good on my small laptop, as shown:

Screen Shot 2021-07-15 at 2 54 42 PM

Can we remove the wrap here and ensure the SearchFilter is within a `<Box flex="shrink" />`?

aries-site/src/examples/templates/FilterControls/FilterControls.js
```diff
@@ -41,19 +42,28 @@ export const FilterControls = ({
...
+        <Box
+          direction="row"
+          align="start"
+          gap="small"
+          margin={{ bottom: 'xsmall' }}
```
ericsoderberghp, line 51: Should this Box have `flex`?

taysea: I am no noticing any difference when having `flex` vs not having `flex.

ericsoderberghp: OK. Maybe that's a default.

ericsoderberghp (approved):

PR #1717, opened 2021-07-07
Add aria-describedBy to the DataTable examples

### LINKED ISSUE #1683: DataTable Accessibility - screen reader feedback
The following comment came from Bill Tipton regarding the Design-System Table behavior:

>when I was exploring the table on  the HPE Design System at:
https://design-system.hpe.design/
 
> I did notice it had column headings although no row headings (just looked) allowing me to tie the data in the cell to the column heading which helped me understand what the data represented based on the column heading. However, even with this information I pretty much had no idea what data the complete table was supposed to be representing. Feedback: Maybe “Storage Pools” was the Caption for the table? If not if you added a caption tag as described at the link below this would tell screen reader users what the data in the table overall is supposed to be displaying to the user.
https://www.w3.org/WAI/tutorials/tables/caption-summary/
 
>Storage Pools was a Heading just above the table which I am assuming describes what the data in the table represents.
If this is true in my opinion you would not need the Caption Tag since that would be a duplicate of the Heading.

MikeKingdom:
#### What does this PR do?
This change helps screen readers identify the contents of a table by linking the heading for the table to table itself. This is one of the alternatives from the [WAI table recommendations](https://www.w3.org/WAI/tutorials/tables/caption-summary/).

TableExample.js

Tested with NVDA screen reader

Visit the Table component and filtering example pages with a screenreader enabled like NVDA, JAWS or VoiceOver.

This is intended to help developers that copy these examples address accessibility as well.

Fixes #1683 

Not necessary

### DIFF
--- a/aries-site/src/examples/components/pagination/PaginationTableExample.js
+++ b/aries-site/src/examples/components/pagination/PaginationTableExample.js
@@ -541,11 +541,16 @@ export const PaginationTableExample = () => {
 
   return (
     <Box pad="small">
-      <Heading level={3} margin={{ bottom: 'small', top: 'none' }}>
+      <Heading
+        id="storage-pools-heading"
+        level={3}
+        margin={{ bottom: 'small', top: 'none' }}
+      >
         Storage Pools
       </Heading>
       <Box>
         <DataTable
+          aria-describedby="storage-pools-heading"
           data={data}
           columns={[
             {
--- a/aries-site/src/examples/components/table/TableExample.js
+++ b/aries-site/src/examples/components/table/TableExample.js
@@ -294,11 +294,16 @@ export const TableExample = () => {
 
   return (
     <>
-      <Heading level={3} margin={{ bottom: 'small', top: 'none' }}>
+      <Heading
+        id="storage-pools-heading"
+        level={3}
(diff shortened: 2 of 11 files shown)


ericsoderberghp (approved):

ericsoderberghp:
Might be good to mention this either here https://design-system.hpe.design/foundation/accessibility and/or here https://design-system.hpe.design/components/table as well.

MikeKingdom:
> Might be good to mention this either here https://design-system.hpe.design/foundation/accessibility and/or here https://design-system.hpe.design/components/table as well.

Looking at the accessibility section, I'm not seeing how it would fit in there since that doesn't really go into this type of detail.

It certainly seems like I could add something to the https://design-system.hpe.design/components/table 'Guidance' section though. Thoughts?

ericsoderberghp:
Let's start by adding an "Accessibility" section to the Table page. We can comment on the fact that `scope="row"` is automatically added. And we can direct callers to have an associated Heading with an `id` that is referenced by `aria-describedBy`.

taysea (approved):
Thanks for adding this!

MikeKingdom:
Thanks!

PR #1789, opened 2021-08-17
put in a option for fire fox

britt6612:
#### What does this PR do?
This PR shows a preview of what `datatable` header would look like in FireFox when you scroll
aries theme 
This would be taken out and put into the theme 
visual 
http://localhost:3000/components/table
go to the link above in firefox
https://github.com/grommet/grommet-theme-hpe/issues/200
issue 200
no
backwards compatible

### DIFF
--- a/aries-site/src/themes/aries.js
+++ b/aries-site/src/themes/aries.js
@@ -3,6 +3,40 @@ import { deepMerge } from 'grommet/utils';
 
 export const aries = deepMerge(hpe, {
   defaultMode: 'dark',
+  dataTable: {
+    pinned: {
+      header: {
+        extend: ({ theme }) => `
+        backdrop-filter: blur(8px);
+        @-moz-document url-prefix() {
+        background: ${
+          theme.dark ? 'rgba(64, 75, 92, 0.95)' : 'rgba(255, 255, 255, 0.95)'
+        };
+        }
+          `,
+      },
+      body: {
+        extend: ({ theme }) => `
+        backdrop-filter: blur(8px);
+        @-moz-document url-prefix() {
+        background: ${
+          theme.dark ? 'rgba(64, 75, 92, 0.95)' : 'rgba(255, 255, 255, 0.95)'
+        };
+        }
+          `,
+      },
+      footer: {
+        extend: ({ theme }) => `
+        backdrop-filter: blur(8px);
+        @-moz-document url-prefix() {
+        background: ${
+          theme.dark ? 'rgba(64, 75, 92, 0.95)' : 'rgba(255, 255, 255, 0.95)'
+        };
+        }
+          `,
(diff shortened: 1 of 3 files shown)


aries-site/src/themes/aries.js
```diff
@@ -3,6 +3,40 @@ import { deepMerge } from 'grommet/utils';
...
+  dataTable: {
+    pinned: {
+      header: {
+        extend: ({ theme }) => `
+        backdrop-filter: blur(8px);
+        @-moz-document url-prefix() {
+        background: ${
+          theme.dark ? 'rgba(64, 75, 92, 0.95)' : 'rgba(255, 255, 255, 0.95)'
+        };
+        }
```
halocline, line 15: Can you add a comment for why this is needed and also provide a link to MDN docs / can-i-use docs for support of `backdrop-filter`? Sounds like FireFox should be supporting shortly - currently an opt in users can enable, but not yet a default.

aries-site/src/themes/aries.js
```diff
@@ -3,6 +3,40 @@ import { deepMerge } from 'grommet/utils';
...
 export const aries = deepMerge(hpe, {
   defaultMode: 'dark',
+  dataTable: {
+    pinned: {
+      header: {
+        extend: ({ theme }) => `
+        backdrop-filter: blur(8px);
+        @-moz-document url-prefix() {
+        background: ${
+          theme.dark ? 'rgba(64, 75, 92, 0.95)' : 'rgba(255, 255, 255, 0.95)'
```
halocline, line 13: Can we translate these to the hexadecimal equivalents for consistency?

halocline: How were these values arrived at? Did you have a chance to run these by @vavalos5 for a quick 👍 👎 ?

britt6612: I will change these and no I will run them by her I tried to use the `background-contrast` but it wasn't enough `opactity` still looked pretty bad

aries-site/src/themes/aries.js
britt6612: yes I read that Firefox had `backdrop-filter` on its list of priority

aries-site/src/themes/aries.js
britt6612: @vavalos5 and I spoke and she is okay with these changes she just mentioned to make sure we stay on top of it when it is available for firefox so we can take this out!

aries-site/src/themes/aries.js
britt6612: I will change these values on the PR for the theme and add the comment there then we can strip this out

taysea:
Closing this PR since the theme PR has been merged.

PR #1819, opened 2021-08-31
DO NOT MERGE Create icon deploy for grommet-icons updates

taysea:
#### What does this PR do?

Preview

We are updating the icons to use "stroke" and "stroke-width" in cases where they previously were using "fill". The updates are happening in grommet-icons, but this is just a deploy to show the new vs old for visual confirmation since snapshots changed in grommet-icons due to fill --> stroke.


halocline:
@taysea I believe this can be closed now that the update icons are within Grommet-Icons.

Issue #1828, opened 2021-09-03
Site Search - Create branch demonstrating "immediate wins"


halocline:
Parent issue: https://github.com/grommet/hpe-design-system/issues/1787

- [x] Create a "Search" branch on hpe-design-system
- [x] Provide a proof of concept to demonstrate the following:
  - [x] Remove query constraint (i.e. allow user to type anything they want and hit Enter)
  - [x] Search Results page/view/component
    - Component will accept an array of search results objects
    - Search results object should resemble:
```
{
  query: string,
  matchedText: string,
  matchType: string,
  url: string,
  preview: string | JSX.element,
}
```
  - [x] Add and populate the following items to a 4-5 `structure.js` page entries:
    - [x] Title synonyms
    - [x] Tags
  - [x] Expand metadata search matches including:
    - [x] Title (currently implemented)
    - [x] Section titles (currently implemented)
    - [x] Title synonyms
    - [x] Description
    - [x] Tags
- [x] Share link to demo branch ([Demo branch](https://deploy-preview-1842--keen-mayer-a86c8b.netlify.app/))
- [x] Create issue to incorporate UX feedback https://github.com/grommet/hpe-design-system/issues/1848
- [x] Create issue to populate expanded meta data (e.g. tags for each page, add page sections where appropriate) https://github.com/grommet/hpe-design-system/issues/1849
- [x] Create issue to demonstrate prototype for "content/document" search (i.e. searching a page's actual content) https://github.com/grommet/hpe-design-system/issues/1850

PR #1842, opened 2021-09-09
Site Search

halocline:
## This is a prototype / work in progress

https://user-images.githubusercontent.com/1756948/132775472-9595170f-b9c6-4d50-81b7-ab9b5c887054.mov

**Not included:**
- Not ready for code review. Needs refactoring. Purpose is to demonstrate UX prototype.
- Currently only searching page metadata. Have not yet implemented searching of page content.

**This [demo branch](https://deploy-preview-1842--keen-mayer-a86c8b.netlify.app/) is intended to demonstrate the following:**
- [ ] Removes query constraint (i.e. allow user to type anything they want and hit Enter)
- [ ] Expands types of search matches to include:
  - [ ] Page titles
  - [ ] Page section titles (e.g. h2, h3s)
  - [ ] Page description
  - [ ] Page tags (i.e. can be used for synonyms such as tile vs. card)
  - NOT YET INCLUDED - Page's entire contents (I have a proof of concept for this, but is not included in this branch yet)
- [ ] Provides a proposed "expanded search results view" which features:
  - [ ] Dedicated, focused search interface
  - [ ] Expanded results for which each result contains:
    - [ ] Page / section title
    - [ ] url
    - [ ] "preview content" - currently this just displays a page's description, but the plan is to replace that with the matched subsection of the page's content once implemented
    - [ ] tags
  - [ ] Query-match text highlighting

**Areas to concentrate feedback**
- [ ] General user experience / flow
- [ ] Design / layout treatments for search results interface
- [ ] Other

Closes https://github.com/grommet/hpe-design-system/issues/1828

### DIFF
--- a/aries-site/src/data/structures/components.js
+++ b/aries-site/src/data/structures/components.js
@@ -70,7 +70,15 @@ export const components = [
       },
     },
     relatedContent: ['Lists', 'Dashboards', 'Navigation'],
-    sections: [],
+    sections: [
+      'Dashboard Example',
+      'Navigation Example',
+      'Card as a Dashboard Element',
+      'Card with an Image',
+      'Card with a Footer',
+      'More Examples of a Card',
+    ],
+    tags: ['tile', 'widget', 'container', 'card'],
   },
   {
     name: 'Tabs',
@@ -153,7 +161,7 @@ export const components = [
       'A Tip is used to specify extra information when the user moves the mouse pointer over the element.',
     seoDescription:
       'A Tip is used to specify extra information when the user moves the mouse pointer over the element.',
-    sections: [],
+    sections: ['Tip with icons', 'Tip with exit', 'Tip Content Length'],
     preview: {
       image: {
         src: {
@@ -164,6 +172,7 @@ export const components = [
       },
     },
     relatedContent: ['Layer', 'Button', 'Header', 'Persistent Navigation'],
+    tags: ['tooltip', 'tool tip', 'hover', 'hover text', 'hint'],
   },
   {
     name: 'Search',
@@ -244,6 +253,7 @@ export const components = [
     seoDescription:
       'The accordion affords content to be delivered progressively.',
     sections: [],
+    tags: ['accordian', 'acordion', 'collapse', 'panel', 'expand'],
   },
   {
     name: 'Header',
@@ -355,7 +365,8 @@ export const components = [
   {
     name: 'Notification',
     category: 'Visualizations',
-    description: 'Notifications deliver transparent clarity for task and system statuses.',
+    description:
+      'Notifications deliver transparent clarity for task and system statuses.',
     preview: {
       image: {
         src: {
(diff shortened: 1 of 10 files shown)


taysea (commented):
This is a huge improvement to what we have. Here are some reactions to the experience:
- I was confused when the suggestions went away because I thought that meant there was nothing that would match, but when I hit enter and got a modal pop-up I was like "Oh, I wish I had seen this the whole time" -- that being said, wondering if we could have the modal open from the get-go OR have the dropdown suggestions include the details that the modal provides
- I am worried about the blue highlight for contrast accessibility -- would just doing bold text be enough? or bold + underline?

halocline:
> * I was confused when the suggestions went away because I thought that meant there was nothing that would match, but when I hit enter and got a modal pop-up I was like "Oh, I wish I had seen this the whole time" -- that being said, wondering if we could have the modal open from the get-go OR have the dropdown suggestions include the details that the modal provides
> * I am worried about the blue highlight for contrast accessibility -- would just doing bold text be enough? or bold + underline?

Great comments. Keep'em coming.

britt6612:
Nice Job! I really like the modal aspect when you type `tile` and then its highlighted in the results..
I would say adding something for when something does not exist.. So if I type `anything` maybe having "no results found" or something in the modal to let the user know

britt6612:
Im also wondering if we should add something at the end of the search that says `view more` or `see detailed results` on a longer result like `colors` I guess don't really expect the layer to open so when I hit `enter` and it  does Im a little taken back but just my thoughts
https://user-images.githubusercontent.com/42451602/132778683-80051061-d04b-45f8-a325-aa6643108d89.mov

bahriaditi:
Thanks for sharing your work in progress @halocline!

+1 to @taysea's comment: 

> "have the dropdown suggestions include the details that the modal provides"

Attaching an example of this from [Shopify's design system](https://polaris.shopify.com/). [Side note: They seem to be using a third party search API]

I also like the idea of bolding input letters as they are typed in the dropdown results – nice way to give feedback and makes me feel heard as a user.

Screen Shot 2021-09-09 at 8 47 15 PM

halocline:
From [Slack thread](https://hpe.slack.com/archives/GPPLUK6LR/p1631232172098200?thread_ts=1631230705.097100&cid=GPPLUK6LR):
I think this is awesome and something we’ve definitely been needing. Great enhancement.
Here are some thoughts I had:
- If the user clicks out of the modal, will this result in the modal closing? Resulting in loss of what was searched? ( Can create a frustrating experience)
- If there are tons of results for let’s say “button”, will a modal limit the amount of space needed to fit more content to be more easily accessible to a user. The modal is limited so, it’s likely that a user will have to scroll
- I’ve seen that the most common experience for this type of behavior is implementing it in a new page. Is this possible?

halocline:
Closing in favor of https://github.com/grommet/hpe-design-system/pull/1871

PR #1853, opened 2021-09-13
add new prop to align list

### LINKED ISSUE #1790: Order of Items Example - List data should be top-aligned
Slack thread - https://hpe.slack.com/archives/C04LMJWQT/p1629231480089100

image

britt6612:
#### What does this PR do?
utilizing the new `defaultItemProps` that was added to grommet now we can use this to align our `list` onOrder example
templates/list-views/ListOrderExample.js 
site
check out the example and see it is  align start
closes #1790 
yes
backwards compatible

### DIFF
--- a/aries-site/src/pages/templates/lists.mdx
+++ b/aries-site/src/pages/templates/lists.mdx
@@ -187,11 +187,20 @@ With long sets of data, pagination can help to create a more efficient means of 
 
 ### Item order
 
+For **multi-line** list items, they should be set to top aligned for easy scanning across a row.
+In order to align start use the prop `defaultItemsProp`.
+
+```
+<List
+defaultItemsProp={{align: 'start'}}
+...any additional props
+<List/>
+```
+
+
 Allowing the user to re-order the list by dragging items or pressing move up and down controls.
-When using OnOrder if there is multi-line content then this should be top aligned for easy scanning
-across a row. In order to aliign start you can add the prop `defaultItemsProp` to `DataTable`
-to be set as `align: 'start'`. Choose `Actions->Re-Order Items` to
-enable drag and drop of list items and move up/down controls. Press `Done` when finished.
+Choose `Actions->Re-Order Items` to enable drag and drop of list items and move up/down controls.
+Press `Done` when finished.
 
 <Example
   background="background-back"
--- a/aries-site/src/pages/templates/lists.mdx
+++ b/aries-site/src/pages/templates/lists.mdx
@@ -187,7 +187,10 @@ With long sets of data, pagination can help to create a more efficient means of 
 
 ### Item order
 
-Allowing the user to re-order the list by dragging items or pressing move up and down controls. Choose `Actions->Re-Order Items` to
+Allowing the user to re-order the list by dragging items or pressing move up and down controls.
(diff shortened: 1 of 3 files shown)


taysea (commented):
This looks good implementation-wise -- just wondering but could be separate work, are we recommending that all lists do this? If so, is this something that should be enforced from the theme level?

alternately, should we add guidance to this example for developers about using defaultItemProps align: 'start'?

ericsoderberghp:
> This looks good implementation-wise -- just wondering but could be separate work, are we recommending that all lists do this? If so, is this something that should be enforced from the theme level?
> 
> alternately, should we add guidance to this example for developers about using defaultItemProps align: 'start'?

I don't think this will be the case for all lists, just those with multi-line content in a list item.

britt6612:
> > This looks good implementation-wise -- just wondering but could be separate work, are we recommending that all lists do this? If so, is this something that should be enforced from the theme level?
> > alternately, should we add guidance to this example for developers about using defaultItemProps align: 'start'?
> 
> I don't think this will be the case for all lists, just those with multi-line content in a list item.

 added guidance to only add if there is multi-line wonder if I should bold that to make stand out more.

aries-site/src/pages/templates/lists.mdx
```diff
@@ -187,8 +187,20 @@ With long sets of data, pagination can help to create a more efficient means of
 
 ### Item order
 
-Allowing the user to re-order the list by dragging items or pressing move up and down controls. Choose `Actions->Re-Order Items` to
-enable drag and drop of list items and move up/down controls. Press `Done` when finished.
+For **multi-line** list items, they should be set to top aligned for easy scanning across a row.
```
taysea, line 190: Docs wording suggestion:

List supports `onOrder` property which allows the user to re-order the list by dragging items or pressing up and down controls.

Choose `Actions -> Re-Order Items` to enable drag and drop of list items and move up/down controls. While in re-ordering mode, take note of how the row content is top-aligned. For **multi-line** list items like this use case, list items should be top-aligned for easy scanning across a row.

To implement this, use the prop `defaultItemsProp` on List:

```
<List
defaultItemsProp={{align: 'start'}}
...any additional props
<List/>
```

aries-site/src/pages/templates/lists.mdx
```diff
@@ -187,8 +187,20 @@ With long sets of data, pagination can help to create a more efficient means of
...
 
-Allowing the user to re-order the list by dragging items or pressing move up and down controls. Choose `Actions->Re-Order Items` to
-enable drag and drop of list items and move up/down controls. Press `Done` when finished.
+List supports `onOrder` property which allows the user to re-order the list by dragging items or pressing up and down controls.
+
+Choose `Actions -> Re-Order Items` to enable drag and drop of list items and move up/down controls. While in re-ordering mode,
+take note of how the row content is top-aligned. For **multi-line** list items like this use case, list items should be top-aligned
+for easy scanning across a row.
+
+To implement this, use the prop `defaultItemsProp` on List:
```
taysea, line 196:
```suggestion
To implement top-alignment for multi-line lists, use the prop `defaultItemsProp` on List:
```

taysea (approved):

halocline (approved):

Issue #1848, opened 2021-09-10
Site Search - Incorporate "search results view" UX feedback


halocline:
Related to #1787
Related to #1828

- [x] Consider feedback regarding how the Search Results View is presented
  - Transition from suggestions to modal view surprising. Consider showing expanded detail within the suggestions as opposed to separate. OR, clicking on search icon launches modal right away.
  - Should it be a dedicated page of results?
- [x] Text highlighting
  - Mostly positive reactions
  - Consider whether color treatment is accessible as opposed to bolding
- [x] Consider adding messaging for cases where lots of results found vs. no results found
- [x] Create issue for exploring search history concept (What is the UX? Do we show most recent searches? How displayed? Store in local storage? Site Search - Explore concept and value of search history #1880
- [x] After considerations, incorporate into revision of prototype
- [x] Share link to revised prototype and collect feedback
  - Iteration A - [Demo](https://deploy-preview-1857--keen-mayer-a86c8b.netlify.app/) | [PR](https://github.com/grommet/hpe-design-system/pull/1857)
  - Iteration B - [Demo](https://deploy-preview-1871--keen-mayer-a86c8b.netlify.app/) | [PR](https://github.com/grommet/hpe-design-system/pull/1871)

PR #1857, opened 2021-09-14
Site search UX - Expanded results in TextInput suggestions drop

halocline:
Provides iteration on site search UX. Related to #1848

[This execution](https://deploy-preview-1857--keen-mayer-a86c8b.netlify.app/) presents expanded search results directly within the search suggestions drop.

Screen Shot 2021-09-15 at 9 42 33 AM

PR #1871, opened 2021-09-15
Site search UX - Expanded results in Layer

halocline:
Provides iteration on site search UX. Related to https://github.com/grommet/hpe-design-system/issues/1848

[This execution](https://deploy-preview-1871--keen-mayer-a86c8b.netlify.app/) presents expanded search results within a focused layer/modal.

**Testing Recommendations**:
Currently, the search matching algorithm is pretty primitive, but provides a base to build from.

The current approach looks for instances where the provided query string has an exact match in a page's title, section headings, or tag metadata. Each match instance of a page's title or section heading will be included in the search results. One and only one tag match instance will be returned in the search results.

Recommended queries:
- 'head'
- 'header'
- 'heading'
- 'tile'
- 'widget'
- 'truncat'
- 'truncated'
- 'truncating'

Screen Shot 2021-09-15 at 9 42 01 AM


britt6612:
One comment with this approach.. Is as a user is searching they might go back to look at the page and with having a layer you need to click out of the modal you are in.. I like the approach of how #1857 is next to the search and your able to search but  scroll if you needed to look at something. Just my 2 cents with the UX side.

halocline:
> One comment with this approach.. Is as a user is searching they might go back to look at the page and with having a layer you need to click out of the modal you are in.. I like the approach of how #1857 is next to the search and your able to search but scroll if you needed to look at something. Just my 2 cents with the UX side.

Can you elaborate on/point to an experience where it would be useful to have search results open while also consuming content on a page? I am trying to think of some, but keep coming back to "task focussed, one thing at a time" but could certainly have blinders on.

britt6612:
> > One comment with this approach.. Is as a user is searching they might go back to look at the page and with having a layer you need to click out of the modal you are in.. I like the approach of how #1857 is next to the search and your able to search but scroll if you needed to look at something. Just my 2 cents with the UX side.
> 
> Can you elaborate on/point to an experience where it would be useful to have search results open while also consuming content on a page? I am trying to think of some, but keep coming back to "task focussed, one thing at a time" but could certainly have blinders on.

Guess I do like the idea of "task focussed, one thing at a time" I just tend to wander off. I have done it before in the past I think I tend to start to "Search" something then double-check the home page if there was something relevant to what I was looking for.. I do think this may be just something  I do and not necessarily the correct thing. Since our home page is pretty condensed in terms of how much information is there. I would agree here with the "task focussed, one thing at a time" Theres  really nothing on our home page that would necessarily help someone with their search

halocline:
> > > One comment with this approach.. Is as a user is searching they might go back to look at the page and with having a layer you need to click out of the modal you are in.. I like the approach of how #1857 is next to the search and your able to search but scroll if you needed to look at something. Just my 2 cents with the UX side.
> > 
> > 
> > Can you elaborate on/point to an experience where it would be useful to have search results open while also consuming content on a page? I am trying to think of some, but keep coming back to "task focussed, one thing at a time" but could certainly have blinders on.
> 
> Guess I do like the idea of "task focussed, one thing at a time" I just tend to wander off. I have done it before in the past I think I tend to start to "Search" something then double-check the home page if there was something relevant to what I was looking for.. I do think this may be just something I do and not necessarily the correct thing. Since our home page is pretty condensed in terms of how much information is there. I would agree here with the "task focussed, one thing at a time" Theres really nothing on our home page that would necessarily help someone with their search

Keep us posted if anything else comes to mind.

aries-site/src/examples/components/card/CardMeterExample.js
```diff
@@ -33,7 +33,7 @@ export const CardMeterExample = () => (
               color: 'graph-1',
               highlight: false,
               label: 'availability',
-              onClick: '() => {}',
+              onClick: () => {},
```
halocline, line 36: opportunistic propTypes validation error --> function, not string

aries-site/src/examples/components/notifications/StateVsStatusDiagram.js
```diff
@@ -4,7 +4,7 @@ import { ToastPreview } from '../../cardPreviews';
 
 const color = 'text-strong';
 const anchor = 'vertical';
-const thickness = 'xxxsmall';
+const thickness = 'hair';
```
halocline, line 7: opportunistic propTypes validation error --> 'hair' is valid value, 'xxxsmall' is not

aries-site/src/layouts/content/SubsectionHeader.js
```diff
@@ -36,6 +36,6 @@ export const SubsectionHeader = ({ children, level }) => {
 };
 
 SubsectionHeader.propTypes = {
-  children: PropTypes.string.isRequired,
+  children: PropTypes.oneOfType([PropTypes.string, PropTypes.array]).isRequired,
```
halocline, line 39: opportunistic propTypes validation error --> some markdown titles are passed as arrays

taysea (commented):
I don't think the "tags" are necessary to have shown -- I think it feels distracting especially since they include misspelled things. I'd prefer if those things could be indexed in the meta data but not displayed and highlighted

halocline:
> I don't think the "tags" are necessary to have shown -- I think it feels distracting especially since they include misspelled things. I'd prefer if those things could be indexed in the meta data but not displayed and highlighted

Cool. I think I feel the same. I had left them in for a) auditing / demonstrating functionality internally, and b) the following scenario:
- If you were to search for "widget", "tile", or "hover" and the tags were not present, would you feel like you were receiving a relevant result? If so, I think we should remove as well.

taysea:
> * If you were to search for "widget", "tile", or "hover" and the tags were not present, would you feel like you were receiving a relevant result? If so, I think we should remove as well.

We could try having text potentially that says "Nothing found for [search term]." and then have the "similar" results appear below that?

aries-site/src/layouts/navigation/HighlightPhrase.js
```diff
@@ -0,0 +1,42 @@
...
+      nextIndex = match.index + match[0].length;
+      nextPhrase.push(
+        <Text
+          key={index}
+          weight="bolder"
+          color={
+            theme.dark
+              ? theme.global.colors.yellow.light
+              : theme.global.colors.blue.dark
+          }
```
britt6612, line 24: why are we using `yellow` for dark then `blue` for light?

britt6612:
Wondering if we can make the `layerContainer` also have rounded corners or ask a designer to me this feels off its very small difference but something that I caught when looking through..

Screen Shot 2021-09-20 at 4 08 37 PM

britt6612:
On certain searches, it tends to go blank 

Screen Shot 2021-09-20 at 4 11 57 PM

britt6612:
One thing to note that we need to add `bolder` to `weight` in grommet for Text propTypes. 

Screen Shot 2021-09-20 at 4 14 26 PM

aries-site/src/layouts/navigation/HighlightPhrase.js
halocline: Colors are selected for ability to effectively highlight with enough contrast for legibility. 

Yellow on black has strong contrast. Yellow on white, not so much, so need another another color to use. I could not find a color variant which performed well in both contexts, so used two colors.

Open to alternative suggestions.

aries-site/src/layouts/navigation/SearchResults.js
```diff
@@ -0,0 +1,107 @@
...
+            size="xlarge"
+          />
+        </Keyboard>
+      </Box>
+      <Box overflow="auto" pad="xsmall">
+        <List
+          data={results}
+          onClickItem={onSelect}
+          border={{ side: 'bottom', color: 'border-weak' }}
+          // paginate={results.length > step ? { page } : false}
```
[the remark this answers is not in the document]
halocline: Related to [this comment](https://github.com/grommet/hpe-design-system/pull/1871#issuecomment-923373062). 

This should be removed. It was exploring an approach to address the 'blank page' mentioned above. 

This needs more exploration as Pagination within a List is giving some unexpected results when List's data is dynamic. We may need to implement as a controlled pagination instead of using that supplied by List.

halocline:
> Wondering if we can make the `layerContainer` also have rounded corners or ask a designer to me this feels off its very small difference but something that I caught when looking through..
> 
> Screen Shot 2021-09-20 at 4 08 37 PM

Good catch.

halocline:
> Wondering if we can make the `layerContainer` also have rounded corners or ask a designer to me this feels off its very small difference but something that I caught when looking through..
> 
> Screen Shot 2021-09-20 at 4 08 37 PM

Good catch.

halocline:
> One thing to note that we need to add `bolder` to `weight` in grommet for Text propTypes.
> 
> Screen Shot 2021-09-20 at 4 14 26 PM

Yes. On my todos.

halocline:
> On certain searches, it tends to go blank

Yes. A result exists, but Pagination is not returning to the first page of results. This may be a bug in List's `paginate` rather than this particular implementation, but haven't confirmed yet.

halocline:
> > * If you were to search for "widget", "tile", or "hover" and the tags were not present, would you feel like you were receiving a relevant result? If so, I think we should remove as well.
> 
> We could try having text potentially that says "Nothing found for [search term]." and then have the "similar" results appear below that?
> 
> This might give the user a bit more closure knowing that we did "hear them" with their search term and are suggesting other potentially relevant options.

I think a "similar" or "related" results section might work. I will give it a look.

britt6612:
> > One thing to note that we need to add `bolder` to `weight` in grommet for Text propTypes.
> > Screen Shot 2021-09-20 at 4 14 26 PM
> 
> Yes. On my todos.
> 
> Actually, @britt6612, if you have bandwidth, would you be willing to file an issue and PR on Grommet to resolve this? That way I can incorporate the rest of this feedback and the rest of the metadata work?

yes of course I can do that real quick!

britt6612:
> > On certain searches, it tends to go blank
> 
> Yes. A result exists, but Pagination is not returning to the first page of results. This may be a bug in List's `paginate` rather than this particular implementation, but haven't confirmed yet.

okay sounds good! maybe something we can debug find out!

aries-site/src/layouts/navigation/SearchResults.js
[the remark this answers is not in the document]
britt6612: Thanks for clarifying

aries-site/src/layouts/navigation/HighlightPhrase.js
britt6612: Let me look at the colors and see Ill suggest if I find something else

halocline:
FYI, I am working on adjusting the "search" TestCafe tests to align with this new presentation. Other than that, ready for UX, QA, and code reviews.

aries-site/src/layouts/navigation/Search.js
```diff
@@ -1,114 +1,180 @@
...
+    return {
+      label: <SearchResult query={query} result={nextValue} />,
+      value: nextValue,
+    };
+  });
+
+/* Sort search results by relevancy and/or strength of match type */
+const sortSearch = matches => {
+  const results = matches
+    /* Prioritize title matches over metadata matches */
```
britt6612, line 32: this would also prioritize title matches over content matches?

aries-site/src/layouts/navigation/SearchResults.js
```diff
@@ -0,0 +1,168 @@
...
+  Keyboard,
+  List,
+  ResponsiveContext,
+  Pagination,
+  Text,
+} from 'grommet';
+import { FormClose } from 'grommet-icons';
+
+import { SearchInput, SearchResult } from '.';
+
```
britt6612, line 15: was this added so that when you searched it would show on first page?

aries-site/src/layouts/navigation/HighlightPhrase.js
britt6612: I looked through the colors that we offer and had tried `orange` and `teal` and your correct the contrast was not as good so I think we are okay with keeping as is with the `blue` and `yellow` I was just looking for a consistent color but did not find a great one for both

aries-site/src/layouts/navigation/Search.js
```diff
@@ -1,114 +1,180 @@
...
+  results &&
+  results.map(result => {
+    const nextValue = {
+      ...result.value,
+      url: nameToPath(result.value.title),
+    };
 
-export const Search = ({ focused, setFocused }) => {
+    return {
+      label: <SearchResult query={query} result={nextValue} />,
```
britt6612, line 24: The Search result looks good, however with using `voiceover` it is unclear after you hit enter for whatever keyword you typed it doesn't tell you what the result is.. I think this is something in Grommet we can look at

taysea:
It seems unexpected that the "Similar to..." text would be included in the list item. I would expect it to be text and then a list below it. Understand that could cause rework and I wouldn't hold up this PR to resolve it if it's going to take time.
 
Screen Shot 2021-09-22 at 2 20 01 PM

britt6612:
Screen Shot 2021-09-22 at 3 24 02 PM

When I was searching `header` on the last page I got this related content on the bottom is this something that should be on every search result? Its not consistent across every search.

aries-site/src/layouts/navigation/Search.js
```diff
@@ -1,114 +1,180 @@
...
     }
   };
 
   const onSelect = event => {
-    const href = nameToPath(event.suggestion);
+    const href = nameToPath(event.item.value.title);
     if (internalLink.test(href)) {
       router.push(href);
-      setFocused(false);
+      setOpen(false);
```
taysea, line 150:
```suggestion
      onClose();
```
would we want to do this here?

aries-site/src/layouts/navigation/Search.js
halocline: TBD. Not necessarily. I think this sorting is temporary and can be enhanced quite a bit. I have a concept of a "relevancy score" which should drive the ultimate sort order, but that's a future scope of work.

aries-site/src/layouts/navigation/SearchResults.js
halocline: Not sure which line you have tagged here, but guessing:
```
const defaultPage = { begin: 0, end: 4 };
```
?

If so, these are required to make the controlled paginated results work correctly.

aries-site/src/layouts/navigation/Search.js
halocline: Glad you are testing screen readers. I am going to add an issue to cover an accessibility audit specific to the search behavior. Would be great to nail it.

aries-site/src/layouts/navigation/SearchResults.js
britt6612: Yes it was those line I was just curious how you fixed that!

aries-site/src/layouts/navigation/Search.js
britt6612: Sounds good! Don’t want to hold up this PR we can track in another issue!

halocline:
> It seems unexpected that the "Similar to..." text would be included in the list item. I would expect it to be text and then a list below it. Understand that could cause rework and I wouldn't hold up this PR to resolve it if it's going to take time.
> 

I took a stab at an approach. There are some usability aspects that get introduced and I am not in love with, but I think they are temporary. When we introduce content matching and additional match types, we can also introduce a relevancy score which should make this whole idea of "similar results" a non-factor.

aries-site/src/layouts/navigation/Search.js
halocline: done

aries-site/src/layouts/navigation/SearchResults.js
halocline: Since we are using controlled pagination, then we also need to control which subset of results are passed to List's data. This is the mechanism to do those slices of results.

halocline:
> When I was searching `header` on the last page I got this related content on the bottom is this something that should be on every search result? Its not consistent across every search.

Right now, search results are divided into 1) title + section header matches, and 2) tag matches. The "Similar results" messaging is being inserted between 1 & 2 because folks felt showing the "tags: ..." was confusing / unnecessary.

I think this all goes away once we do content matching + introduce a relevancy score.

halocline:
> FYI, I am working on adjusting the "search" TestCafe tests to align with this new presentation. Other than that, ready for UX, QA, and code reviews.

Modified tests are in place.

aries-site/src/layouts/navigation/Search.js
britt6612: sounds good this was more curiosity on my end

britt6612:
Not sure what happened to this alignment

Screen Shot 2021-09-23 at 7 43 57 PM

halocline:
> Not sure what happened to this alignment
> 
> Screen Shot 2021-09-23 at 7 43 57 PM

Looks like one of these needs a flex={false} --- Are you on a shorter screen (e.g like a laptop?) -- good catch.

britt6612:
> > Not sure what happened to this alignment
> > Screen Shot 2021-09-23 at 7 43 57 PM
> 
> Are you on a shorter screen (e.g like a laptop?) -- good catch. Have a fix.

yes just on my mac! I saw it on the little screen

britt6612 (approved):
Besides the few things, I pointed out everything else looks really solid!!

halocline:
> > > Not sure what happened to this alignment
> > > Screen Shot 2021-09-23 at 7 43 57 PM
> > 
> > 
> > Are you on a shorter screen (e.g like a laptop?) -- good catch. Have a fix.
> 
> yes just on my mac! I saw it on the little screen

Fixed.

britt6612:
> > > > Not sure what happened to this alignment
> > > > Screen Shot 2021-09-23 at 7 43 57 PM
> > > 
> > > 
> > > Are you on a shorter screen (e.g like a laptop?) -- good catch. Have a fix.
> > 
> > 
> > yes just on my mac! I saw it on the little screen
> 
> Fixed.

I just double-checked this and yes it looks good!

PR #1875, opened 2021-09-15
change height if example is in full screen

### LINKED ISSUE #1826: Update Table examples so that height is not always limited, add guidance around limiting Table height
Right now, Table examples always have a fixed height to demonstrate scroll behavior but this is becoming deceiving for developers as they implement and interpret that the height always needs to be limited.

- [x] Update examples so height is not restricted in full screen mode / think about best way to represent table scroll behavior in general
- [x] Add guidance around when to limit/not limit table height
- [x] Add guidance around how/when DataTable should scroll

britt6612:
#### What does this PR do?
This PR take shows the `table` examples when in full screen to not be shortened by the height set. 
table/examples
design system
look at the examples in full screen 
Some user may think they need to always restrict the height for the `table` however this is not true so we need to add guidance on when and when not to restrict the height. 
step one of #1826 
closes #1826
Need to add step 2 & 3 to close issue 
yes
backwards compatible

### DIFF
--- a/aries-site/src/examples/components/table/TableExample.js
+++ b/aries-site/src/examples/components/table/TableExample.js
@@ -304,6 +304,7 @@ export const TableExample = ({ designSystemDemo }) => {
       <Box
         // Height is restricted to keep inline doc page examples more compact.
         // In production, DataTable height should follow height guidelines.
+        // https://design-system.hpe.design/components/table#setting-the-height-of-a-table
         height={designSystemDemo ? undefined : 'small'}
         overflow="auto"
       >
--- a/aries-site/src/examples/components/table/TableExample.js
+++ b/aries-site/src/examples/components/table/TableExample.js
@@ -288,7 +288,7 @@ const handleClickRow = obj => {
   `);
 };
 
-// designSystemDemo is used for DS site can be removed in production.
+// designSystemDemo is used for DS site only, can be removed in production.
 export const TableExample = ({ designSystemDemo }) => {
   const size = React.useContext(ResponsiveContext);
 
--- a/aries-site/src/examples/components/table/TableExample.js
+++ b/aries-site/src/examples/components/table/TableExample.js
@@ -1,4 +1,5 @@
 import React from 'react';
+import PropTypes from 'prop-types';
 import {
   Box,
   DataTable,
@@ -287,7 +288,8 @@ const handleClickRow = obj => {
   `);
 };
 
-export const TableExample = () => {
+// designSystemDemo is used for DS site can be removed in production.
+export const TableExample = ({ designSystemDemo }) => {
   const size = React.useContext(ResponsiveContext);
 
   return (
@@ -299,7 +301,12 @@ export const TableExample = () => {
       >
         Storage Pools
       </Heading>
-      <Box height="medium" overflow="auto">
+      <Box
+        // Height is restricted to keep doc pages more compact.
+        // In production dataTable hight should follow height guidelines
+        height={designSystemDemo ? undefined : 'medium'}
+        overflow="auto"
+      >
         <DataTable
           aria-describedby="storage-pools-heading"
           data={data}
@@ -321,4 +328,8 @@ export const TableExample = () => {
       </Box>
     </>
   );
+};
+
+TableExample.propTypes = {
+  designSystemDemo: PropTypes.bool,
 };
--- a/aries-site/src/examples/components/table/TableExample.js
+++ b/aries-site/src/examples/components/table/TableExample.js
@@ -303,7 +303,7 @@ export const TableExample = ({ designSystemDemo }) => {
       </Heading>
       <Box
         // Height is restricted to keep doc pages more compact.
-        // In production dataTable hight should follow height guidelines
+        // In production, DataTable height should follow height guidelines.
         height={designSystemDemo ? undefined : 'small'}
         overflow="auto"
       >
--- a/aries-site/src/examples/components/table/TableExample.js
+++ b/aries-site/src/examples/components/table/TableExample.js
@@ -304,7 +304,7 @@ export const TableExample = ({ designSystemDemo }) => {
       <Box
         // Height is restricted to keep doc pages more compact.
         // In production dataTable hight should follow height guidelines
-        height={designSystemDemo ? undefined : 'medium'}
+        height={designSystemDemo ? undefined : 'small'}
         overflow="auto"
       >
         <DataTable
--- a/aries-site/src/examples/components/table/TableExample.js
+++ b/aries-site/src/examples/components/table/TableExample.js
@@ -302,7 +302,7 @@ export const TableExample = ({ designSystemDemo }) => {
         Storage Pools
       </Heading>
       <Box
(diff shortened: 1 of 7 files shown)


aries-site/src/examples/components/table/TableExample.js
```diff
@@ -287,7 +288,8 @@ const handleClickRow = obj => {
   `);
 };
 
-export const TableExample = () => {
+// designSystemDemo is used for DS site can be removed in production.
```
taysea, line 291:
```suggestion
// designSystemDemo is used for DS site only, can be removed in production.
```

aries-site/src/examples/components/table/TableExample.js
```diff
@@ -299,7 +301,12 @@ export const TableExample = () => {
       >
         Storage Pools
       </Heading>
-      <Box height="medium" overflow="auto">
+      <Box
+        // Height is restricted to keep doc pages more compact.
```
taysea, line 305:
```suggestion
        // Height is restricted to keep inline doc page examples more compact.
```

aries-site/src/examples/components/table/TableExample.js
```diff
@@ -299,7 +301,12 @@ export const TableExample = () => {
...
+        // In production dataTable hight should follow height guidelines
```
taysea, line 306:
```suggestion
        // In production, DataTable height should follow height guidelines.
```

aries-site/src/examples/components/table/TableFixedHeaderExample.js
```diff
@@ -227,7 +228,7 @@ const columns = [
   },
 ];
 
-export const TableFixedHeaderExample = () => (
+export const TableFixedHeaderExample = ({ designSystemDemo }) => (
```
taysea, line 231:
```suggestion
// designSystemDemo is used for DS site only, can be removed in production.
export const TableFixedHeaderExample = ({ designSystemDemo }) => (
```

aries-site/src/pages/components/table.mdx
```diff
@@ -175,6 +175,16 @@ accessible in a tool specifically designed for analysis.
   - Align it horizontally with the rest of its layout context. Typically, this is aligned to the start of the content as opposed to stretched.
   - Align any header above the table, including heading, search, filter, actions, in the same way. This keeps an actions menu or button in the context of the table and prevents them from being orphaned. However, don't force the header above the table to align its width to the table, as that might compress the contents of the header too much.
 
+  ### Setting the height of a table
+  
+- Tables should be placed within the main content of the page, it is reccommended to use
```
taysea, line 180: Wording suggestion:


- Let the data set the height of the Table. The Table will grow in height as the length of data increases.
- If it is expected a user will want to navigate to a specific portion of the data, pagination should be used. Generally, it is recommended that a page has a `step` of 20 records. For more information on pagination, read [Pagination guidance](https://design-system.hpe.design/components/pagination).
- Do not restrict the height of a table if the page has available space. This may introduce unnecessary scrolling.
- If the height of a Table needs to be restricted, set `overflow=“auto”` on its parent Box to allow the user to scroll through the data.

taysea (approved):

britt6612:
Sent these changes to @vavalos5 she was good with them as well!

halocline:
Screen Shot 2021-09-17 at 1 31 05 PM

halocline:
Inconsistencies for where the tables appear within fullscreen mode; some are displayed vertically aligned top (my preference), others centered. I would like these to be consistent.
Screen Shot 2021-09-17 at 1 36 24 PM
Screen Shot 2021-09-17 at 1 36 46 PM

britt6612:
> Inconsistencies for where the tables appear within fullscreen mode; some are displayed vertically aligned top (my preference), others centered. I would like these to be consistent.
> Screen Shot 2021-09-17 at 1 36 24 PM
> Screen Shot 2021-09-17 at 1 36 46 PM

Yes some of these in the code are just using a fragment `<></>` and some are using a `box` I was not sure about changing this if we want to add the  `box` with `small pad` like the pagination example that's fine.

aries-site/src/pages/components/table.mdx
```diff
@@ -175,6 +175,13 @@ accessible in a tool specifically designed for analysis.
   - Align it horizontally with the rest of its layout context. Typically, this is aligned to the start of the content as opposed to stretched.
   - Align any header above the table, including heading, search, filter, actions, in the same way. This keeps an actions menu or button in the context of the table and prevents them from being orphaned. However, don't force the header above the table to align its width to the table, as that might compress the contents of the header too much.
 
+  ### Setting the height of a table
+
+- Let the data set the height of the Table. The Table will grow in height as the length of data increases.
+- If it is expected a user will want to navigate to a specific portion of the data, pagination should be used. Generally, it is recommended that a page has a step of 20 records. For more information on pagination, read [Pagination guidance.](https://design-system.hpe.design/components/pagination#when-to-use)
```
halocline, line 181: When internally navigating, avoid using full urls. This will cause the entire page to reload rather than a re-render.

'step' is a bit of Grommet terminology and may not be commonly understood.

```suggestion
- If it is expected a user will want to navigate to a specific portion of the data, pagination should be used. Generally, it is recommended that each page of table results be limited to 20 records. For more information on pagination, read [Pagination guidance].(/components/pagination#when-to-use)
```

halocline:
> Yes some of these in the code are just using a fragment `<></>` and some are using a `box` I was not sure about changing this if we want to add the `box` with `small pad` like the pagination example that's fine.

🤔  TableExample.js uses a fragment, yet it is top-aligned...

britt6612:
> > Yes some of these in the code are just using a fragment `<></>` and some are using a `box` I was not sure about changing this if we want to add the `box` with `small pad` like the pagination example that's fine.
> 
> 🤔 TableExample.js uses a fragment, yet it is top-aligned...

I'm looking now strange 🤔 maybe somewhere else Ill look through now!

halocline:
While we are here, I think we should cleanup how TableSingleSelectExample.js is implemented. The manual repetition of `<DetailsPage ...` is probably not something we'd promote. Rather, we should be mapping through the records to generate those component instances.

Also, there is a typo `setPageDetals` in that file as well.

britt6612:
@halocline I took out the object of `contact` let me know if this is okay, or if I should map it twice..

halocline:
> @halocline I took out the object of `contact` let me know if this is okay, or if I should map it twice..

For the sake of this example, I think that is fine to keep things simple. In a real implementation, we'd want to specify the specific properties in which we were interested and do some transformation ahead of time, but in this case I think we can do the simplified story.

britt6612:
> > @halocline I took out the object of `contact` let me know if this is okay, or if I should map it twice..
> 
> For the sake of this example, I think that is fine to keep things simple. In a real implementation, we'd want to specify the specific properties in which we were interested and do some transformation ahead of time, but in this case I think we can do the simplified story.

Sounds good kinda figured to keep it simple therefore I took it out!

halocline (approved):

PR #1956, opened 2021-10-15
Propose enhancement to support do and don't for examples

taysea:
#### What does this PR do?

Proposing an enhancement to our Example layout to introduce the option do create a "Do" and "Don't" pair of examples. Various stakeholders have expressed interest in increased visual examples for do and don't patterns. While I don't think it is expected we can be exhaustive in all the do's and don'ts, this could be a nice way to call out major don't.

Use "Anchor" page to review the layout proposal (would not be the first example on the page, but I just placed it here for ease).

aries-site/src/layouts/content/Example/Container.js

### DIFF
--- a/aries-site/src/layouts/content/Example/Container.js
+++ b/aries-site/src/layouts/content/Example/Container.js
@@ -1,9 +1,10 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import { Box, defaultProps } from 'grommet';
+import { Box, defaultProps, Paragraph, Text } from 'grommet';
 
 // This creates the overall canvas/frame that the Example sits inside of
 export const Container = ({
+  bestPractice: bestPracticeProp,
   designer, // link to grommet designer example
   docs, // link to grommet doc for component
   figma, // link to figma design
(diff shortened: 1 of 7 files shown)


aries-site/src/layouts/content/Example/Container.js
```diff
@@ -27,31 +28,71 @@ export const Container = ({
...
+        pad={
+          template || screenContainer
+            ? { horizontal: 'large', top: 'large' }
+            : 'large'
+        }
+        round={
+          !horizontalLayout &&
+          (designer || docs || figma || guidance || screenContainer || template)
+            ? { corner: 'top', size: 'small' }
+            : 'small'
```
halocline, line 82: For these smaller boxes, wondering if we can dial back the corners back one notch?

```suggestion
            : 'xsmall'
```

aries-site/src/layouts/content/Example/Container.js
```diff
@@ -27,31 +28,71 @@ export const Container = ({
   else if (!plain) height = { min: 'medium' };
   else height = undefined;
 
+  let bestPractice;
+  let background;
+  let label;
+  if (bestPracticeProp) {
+    if (bestPracticeProp.type === 'do') {
+      background = 'brand';
```
halocline, line 36: Resolves to same color value, but semantically brand doesn't feel as appropriate here.

```suggestion
      background = 'green!';
```

aries-site/src/layouts/content/Example/Container.js
halocline: I guess this would affect the other containers as well... 

Would we be better off having a specialized DoDontContainer? It could consume Container as an import and pass a round prop in.

ericsoderberghp:
> Like this idea of showing both of these Do's and Dont's. I know for this purpose you just put it on the top of the page but wondering if this would go under Guidelines, but then this can also make the page feel very long. I was wondering if maybe there is a way we can view the Do's and Dont's with a collapsible button similar to how we have "show code"
> 
> Unless we plan on just showing 1 examples but I know for the NameValueList there was at least 3 good examples of do's and don't so just thinking of a way to show this information but also in a neat way to not overwhelm the user with a long guidance page

Agreed on not putting it towards the top of the page. It feels like something that should go later on. I don't know that I'd put it under a control to reveal it though.

PR #1978, opened 2021-10-21
Voice tone updates

CitizenFox:
#### What does this PR do?

The goal of this PR is to rebrand our 'Voice and Tone' as 'Content' and reorganize the page into 5 smaller, more-consumable pages of related content. These will include:
- Content: The main hub which all other content-related pages will eventually live under.
- Guidelines: A reference for concrete dos and don'ts.
- Best Practices: A reference for gray-recommendations, shoulds and shouldn'ts.
- Implementation: Implementation practices for designers and developers to create good UX, but not necessarily focused on creating UX content.
- Case Study: A study in developing better content. 

One thing to note is that all 4 Content subpages will eventually be moved from the Foundation-level to live under Content (which will continue to live under Foundation). 

Start with the main [Content](https://deploy-preview-1978--keen-mayer-a86c8b.netlify.app/foundation/contentt) page and then see the [Content References](https://deploy-preview-1978--keen-mayer-a86c8b.netlify.app/foundation/content) at the end of the page.


aries-site/src/data/structures/foundation.js
```diff
@@ -374,42 +372,41 @@ export const foundation = [
...
+        },
+        alt: 'HPE Voice and Tone Preview with two speech bubbles',
+      },
+    },
+    seoDescription: '',
+    sections: [
+      
+    ],
+    relatedContent: [
+
```
halocline, line 406: @CitizenFox I think you need a similar inclusion to what you added above:
```suggestion
'Our Brand'
```

PR #2212, opened 2022-01-12
DO NOT MERGE - Page layouts temp - demo playground

halocline:
[Deploy Preview](https://deploy-preview-2212--keen-mayer-a86c8b.netlify.app/templates/pagelayoutgrid5)


taysea:
I know this is just a draft, but I like the exploration and wanted to leave a quick thought (you might already be working on it).

Would it be possible to have the "Demo Controls" outside of the width constraints of the selected page layout so they don't jump around as much?

halocline:
> I know this is just a draft, but I like the exploration and wanted to leave a quick thought (you might already be working on it).
> 
> Would it be possible to have the "Demo Controls" outside of the width constraints of the selected page layout so they don't jump around as much?

For sure. This was just a quick add so that Aditi would have the ability to audit, manipulate, and play with how the layouts were responding.

If you think others would find value in the controls, would you be willing to mock up what you are thinking? Just a picture of sketch on a piece of paper would do the trick.

PR #2303, opened 2022-02-01
Apply scale factor to rest of theme attributes

halocline:
[Deploy Preview](https://deploy-preview-2303--keen-mayer-a86c8b.netlify.app/templates/content-layouts#prioritizing-content-needs)

#### What does this PR do?

Applies `screenContainer` scaling to Text, Heading, and other theme attributes so that scaled examples present proportionally.

**Background**
As we populate dashboard examples with hi-fidelity content and present them in a scaled view, we would like all components scaled proportionally. Prior scaling was only applied to size, edgeSize, and breakpoints. This PR applies scale factor broadly.

scaled.js

**General UX Checks**
- [x] Light & dark modes

**Accessibility Checks**
- [x] Keyboard interactions

**Code Quality Checks**

### DIFF
--- a/aries-site/src/layouts/content/Example/Example.js
+++ b/aries-site/src/layouts/content/Example/Example.js
@@ -142,9 +142,8 @@ export const Example = ({
     else if (!fullscreen) {
       const containerWidth = mockBrowserRect.width;
       scaledTheme = screenContainer.scale
-        ? scaled(theme, screenContainer.scale)
+        ? scaled(undefined, screenContainer.scale)
         : theme;
-
       const { breakpoints } = scaledTheme.global;
       let breakpoint;
       Object.entries(breakpoints)
--- a/aries-site/src/layouts/content/Example/Example.js
+++ b/aries-site/src/layouts/content/Example/Example.js
@@ -142,7 +142,7 @@ export const Example = ({
     else if (!fullscreen) {
       const containerWidth = mockBrowserRect.width;
       scaledTheme = screenContainer.scale
-        ? scaled(undefined, screenContainer.scale)
+        ? scaled(screenContainer.scale)
         : theme;
       const { breakpoints } = scaledTheme.global;
       let breakpoint;
--- a/aries-site/src/pages/templates/content-layouts.mdx
+++ b/aries-site/src/pages/templates/content-layouts.mdx
@@ -18,10 +18,11 @@ Specifying layout content in this way will also ensure that it is interepreted a
 <Example 
   showResponsiveControls={['fullScreen']}
   code="https://raw.githubusercontent.com/grommet/hpe-design-system/master/aries-site/src/examples/templates/content-layouts/ResponsiveContentLayoutExample.js"
-  screenContainer={{ scale: 0.5 }}
-  // height anad width value chosen to best visually represent the example
+  // height, width, and screenContainer.scale value chosen to best visually 
+  // represent the example
   height="650px"
   width="100%"
+  screenContainer={{ scale: 0.57 }}
 >
   <ResponsiveContentLayoutExample />
 </Example>
--- a/aries-site/src/themes/scaled.js
+++ b/aries-site/src/themes/scaled.js
@@ -2,11 +2,11 @@ import { generate } from 'grommet/themes/base';
 import { deepMerge } from 'grommet/utils';
 import { aries } from './aries';
 
-export const scaled = (theme = aries, scale = 1, baseSpacing = 24) => {
+export const scaled = (scale = 1, baseSpacing = 24) => {
   // scales everything in base
   const scaledBase = generate(baseSpacing * scale);
   // merges theme, but overwrites scale for particular attributes
-  const source = deepMerge(scaledBase, theme);
+  const source = deepMerge(scaledBase, aries);
   const scaledTheme = { ...source };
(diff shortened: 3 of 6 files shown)


aries-site/src/themes/scaled.js
```diff
@@ -1,26 +1,87 @@
...
-  const source = deepMerge(base, theme);
+export const scaled = (theme = aries, scale = 1, baseSpacing = 24) => {
+  // scales everything in base
+  const scaledBase = generate(baseSpacing * scale);
+  // merges theme, but overwrites scale for particular attributes
+  const source = deepMerge(scaledBase, theme);
   const scaledTheme = { ...source };
+  // define attributes which need to be re-scaled; essentially any
+  // size which is staticly set in hpe/aries theme
   const scaleAttributes = [
```
taysea, line 13: whoa, this is a lot of items. I assume this has do be somewhat manual (which is probably fine since we won't be changing theme structure) -- but any thoughts on how to make sure this scales with any additions/changes to theme?

aries-site/src/layouts/content/Example/Example.js
```diff
@@ -142,9 +142,8 @@ export const Example = ({
     else if (!fullscreen) {
       const containerWidth = mockBrowserRect.width;
       scaledTheme = screenContainer.scale
-        ? scaled(theme, screenContainer.scale)
+        ? scaled(undefined, screenContainer.scale)
```
taysea, line 145: Is there anywhere else we're calling `scaled` with another theme? If not, should we just let `scaled()` use aries theme and remove the option for a theme parameter?

aries-site/src/pages/templates/content-layouts.mdx
```diff
@@ -18,10 +18,11 @@ Specifying layout content in this way will also ensure that it is interepreted a
 <Example 
   showResponsiveControls={['fullScreen']}
   code="https://raw.githubusercontent.com/grommet/hpe-design-system/master/aries-site/src/examples/templates/content-layouts/ResponsiveContentLayoutExample.js"
-  screenContainer={{ scale: 0.5 }}
-  // height anad width value chosen to best visually represent the example
+  // height, width, and screenContainer.scale value chosen to best visually 
+  // represent the example
   height="650px"
   width="100%"
+  screenContainer={{ scale: 0.57 }}
```
britt6612, line 25: how did we go from 0.5 to 0.57?

aries-site/src/themes/scaled.js
halocline: I know. It's unfortunate, and the right question to ask. I am trying to balance forward progress (i.e. ability to get the hi-fidelity dashboards unblocked) with a reasonably sound approach. 

Right now, I think that balance is using a little brute force. Brute force in this case is essentially using the array as a "config file." In a way, our HPE-Theme is essentially a giant config file as well.

That said, the two approaches I have in mind are:
1. Make the hpe-theme dynamic, substituting any static pixel-based values with its own `baseSpacing` and calculation. But this would involve changing the theme and I don't think we are ready for that at this time. However, we shouldn't let go of this thought either as there are potential features we could enable such as the Gmail-like default, comfortable, and compact views which would allow end-users to customize their preferred settings.
2. Alternatively, I was thinking of recursively traversing the theme object and doing some sort of parsing, calculation, and substitution.
3. Lastly, this just came to mind, is that we could convert the theme object to a string, use regexp to target any string containing 'px', parse that, then replace with a string template literal containing our desired calculation.

None of these are the most elegant and will likely have edge cases where we would want to opt-out of the transformation, particularly in some of our `extend` cases implementing CSS directly.

As of now, I am thinking we need to move forward, but can continue thinking on this as I would eventually like to enable the "customize my layout" type functionality.

aries-site/src/pages/templates/content-layouts.mdx
halocline: See comment on line 22.

aries-site/src/themes/scaled.js
halocline: I'm open to other ideas as well.

aries-site/src/layouts/content/Example/Example.js
halocline: We can. I will.

I was originally using this to test out the concept where we could repurpose for other use cases. We can always add back at a latter date.

aries-site/src/themes/scaled.js
taysea: Good thoughts. I definitely agree with the "let's move forward" mindset. I'll leave this comment open here to reference later on.

ericsoderberghp: FYI, the grommet-theme-designer has some of this theme manipulation built into it. It might be interesting to extract functions like this and ship them with grommet to allow re-use. https://github.com/grommet/grommet-theme-designer/blob/37c471971d734e45c0fc226567f51e5734d0f92f/src/Spacing.js#L5

halocline: I think it would be a nice feature to have.

aries-site/src/layouts/content/Example/Example.js
halocline: Done.

taysea (approved):

PR #2331, opened 2022-02-08
Two col dashboard polish

halocline:

**General UX Checks**
- [ ] Light & dark modes

**Accessibility Checks**
- [ ] Keyboard interactions

**Code Quality Checks**

### DIFF
--- a/aries-site/src/examples/templates/dashboards/TwoColumnDashboard.js
+++ b/aries-site/src/examples/templates/dashboards/TwoColumnDashboard.js
@@ -30,8 +30,8 @@ const parentGrid = {
   columns: {
     xsmall: '100%',
     small: '100%',
-    medium: ['auto', 'auto'],
-    large: ['auto', 'medium'],
+    medium: ['auto', ['small', 'medium']],
+    large: ['auto', ['small', 'medium']],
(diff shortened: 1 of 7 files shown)


taysea:
I like the gridAreas, gives a little more visibility to how the content responds.

One thing I noticed -- which has to do with PageContainer, AppContainer, and the gaps is the the elevation at the bottom of the last card is getting cut-off by the gap between the page container and the next element. I think all we need to do is set some default bottom padding on the page container (maybe just match it to the horizontal padding at the various breakpoints?)

Screen Shot 2022-02-07 at 6 42 05 PM

britt6612 (approved):
Looks really nice the resizing looks a lot cleaner

halocline:
> I like the gridAreas, gives a little more visibility to how the content responds.
> 
> One thing I noticed -- which has to do with PageContainer, AppContainer, and the gaps is the the elevation at the bottom of the last card is getting cut-off by the gap between the page container and the next element. I think all we need to do is set some default bottom padding on the page container (maybe just match it to the horizontal padding at the various breakpoints?)
> 
I had noticed that too. We'll get it fixed.

PR #2348, opened 2022-02-11
add background for theme context

### LINKED ISSUE #2325: DataTable - Background color on DataTable header incorrect color
The Design System site is showing a bug on the DataTable header. Here's an image of what it's showing:
Screen Shot 2022-02-07 at 8 49 57 AM

[Please align to Figma](https://www.figma.com/file/XEoIWIlLfcARRIFhkpFCch/HPE-DataTable?node-id=2390%3A5274). Thanks :)

britt6612:
[Deploy Preview](https://deploy-preview-2348--keen-mayer-a86c8b.netlify.app/components/datatable?q=data)

#### What does this PR do?
This PR fixes the `datatable` header color by adding `backgroud` to `theme.context`
Example.js

**General UX Checks**
- [x] Light & dark modes

**Accessibility Checks**
- [x] Keyboard interactions

**Code Quality Checks**

The `Header` of the `pinned` DataTable turned into a grey color. 
When the `Theme.Context` was added to the Example.js was when this regression happened. 
With the  `Theme.Context` wrapper it is looking at the theme background color instead of the 
background color of the example wrapper which is the `background-back`

closes #2325 
no
backwards compatible

### DIFF
--- a/aries-site/src/layouts/content/Example/Example.js
+++ b/aries-site/src/layouts/content/Example/Example.js
@@ -182,7 +182,7 @@ export const Example = ({
       >
         {/* prevent theme from overriding the desired background color */}
         <ThemeContext.Extend
-          value={{ ...(scaledTheme || theme), background: 'background' }}
+          value={{ ...(scaledTheme || theme), background: 'background-front' }}
         >
           <ResponsiveContext.Provider value={viewPort}>
             {cloneElement(children, {
--- a/aries-site/src/layouts/content/Example/Example.js
+++ b/aries-site/src/layouts/content/Example/Example.js
@@ -180,7 +180,7 @@ export const Example = ({
         width={width}
         ref={inlineRef}
       >
-        {/* need to overrid theme background so its not resetting */}
+        {/* prevent theme from overriding the desired background color */}
         <ThemeContext.Extend
           value={{ ...(scaledTheme || theme), background: 'background' }}
         >
--- a/aries-site/src/layouts/content/Example/Example.js
+++ b/aries-site/src/layouts/content/Example/Example.js
@@ -180,6 +180,7 @@ export const Example = ({
         width={width}
         ref={inlineRef}
       >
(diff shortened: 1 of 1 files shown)


taysea (approved):
Looks good code wise, but can we add a comment explaining why we need to do this?

britt6612:
> Looks good code wise, but can we add a comment explaining why we need to do this?

yes I will add

aries-site/src/layouts/content/Example/Example.js
```diff
@@ -180,7 +180,10 @@ export const Example = ({
         width={width}
         ref={inlineRef}
       >
-        <ThemeContext.Extend value={scaledTheme || theme}>
+        {/* need to overrid theme background so its not resetting */}
```
halocline, line 183: overrid --> override

Does this make the intent any more clear? As currently written, I got the "what is happening" but still had trouble following the "why we want it".

```suggestion
        {/* prevent theme from overriding the desired background color */}
```

aries-site/src/layouts/content/Example/Example.js
```diff
@@ -180,7 +180,10 @@ export const Example = ({
         width={width}
         ref={inlineRef}
       >
-        <ThemeContext.Extend value={scaledTheme || theme}>
+        {/* prevent theme from overriding the desired background color */}
+        <ThemeContext.Extend
+          value={{ ...(scaledTheme || theme), background: 'background' }}
```
taysea, line 185:
```suggestion
          value={{ ...(scaledTheme || theme), background: 'background-front' }}
```
Think we need to do `background-front` instead.

taysea: For the way dark vs light mode differ

britt6612:
> I am seeing the data table header as a different color. Is that the intent?
> 
> Screen Shot 2022-02-11 at 11 28 20 AM

good catch it should be `background-front`

PR #2389, opened 2022-02-23
Add Cost card to 4-col Dashboard

taysea:
[Deploy Preview](https://deploy-preview-2389--keen-mayer-a86c8b.netlify.app/templates/dashboards/#four-column-dashboard)

#### What does this PR do?
This adds the [cost card](https://www.figma.com/file/4rdBkUlHd5MCVq3hvUOXHd/Page-Layouts?node-id=2675%3A21000) to the 4 column dashboard. One of the main additions to this PR is enhancements to make the `Legend` items clickable. onClick will return the datum that was clicked. In our case, we use this information to adjust the meter styling and change the cost value shown inside the meter. Because the legend is tabbable, it is accessible to a keyboard/screen reader user. I also explored the idea of adding a "compact" mode to the legend (where just the label is displayed and not the value).

To-do:
- [x] when you click on an item from the legend that results in a shorter value, the "last 30days" text is wrapping. need to figure out why that is.

aries-site/src/examples/templates/dashboards/content/chartCards/CostCompact.js

**General UX Checks**
- [x] Light & dark modes

**Accessibility Checks**
- [x] Keyboard interactions

**Code Quality Checks**

Backwards compatible.

### DIFF
--- a/aries-site/src/examples/templates/dashboards/components/Legend.js
+++ b/aries-site/src/examples/templates/dashboards/components/Legend.js
@@ -1,10 +1,13 @@
-import { useContext } from 'react';
+import { Fragment, useContext, useState } from 'react';
 import PropTypes from 'prop-types';
-import { Box, List, Text, ThemeContext } from 'grommet';
+import { Box, Button, List, Text, ThemeContext } from 'grommet';
 
-export const Legend = ({ values, ...rest }) => {
+export const Legend = ({ mode, onClick, values, ...rest }) => {
+  const [active, setActive] = useState();
   const theme = useContext(ThemeContext);
   const size = theme.global.edgeSize.small;
+
+  const Wrapper = onClick ? Button : Fragment;
 
   return (
     <List
@@ -13,18 +16,42 @@ export const Legend = ({ values, ...rest }) => {
       {...rest}
     >
       {datum => (
-        <Box direction="row" gap="small" justify="between">
-          <Box direction="row" align="center" gap="xsmall">
-            <Box background={datum.color} height={size} width={size} round />
-            <Text>{datum.label}</Text>
+        <Wrapper
+          {...(onClick
+            ? {
+                onClick: () => {
+                  onClick(datum);
+                  setActive(active !== datum ? datum : undefined);
+                },
+                active: active === datum,
+                hoverIndicator: true,
+              }
+            : {})}
+        >
+          <Box
+            direction="row"
+            gap="small"
+            justify="between"
+            pad={
+              onClick
+                ? { horizontal: 'xsmall', vertical: 'xxsmall' }
+                : undefined
+            }
+          >
+            <Box direction="row" align="center" gap="xsmall">
+              <Box background={datum.color} height={size} width={size} round />
+              <Text>{datum.label}</Text>
+            </Box>
+            {mode !== 'compact' && <Text>{datum.displayValue}</Text>}
           </Box>
-          <Text>{datum.displayValue}</Text>
-        </Box>
+        </Wrapper>
       )}
     </List>
   );
 };
 
 Legend.propTypes = {
+  mode: PropTypes.oneOf(['compact']),
+  onClick: PropTypes.func,
   values: PropTypes.arrayOf(PropTypes.object),
 };
--- a/aries-site/src/examples/templates/dashboards/components/Legend.js
+++ b/aries-site/src/examples/templates/dashboards/components/Legend.js
@@ -8,6 +8,7 @@ const LegendButton = ({ ...rest }) => (
     <Button hoverIndicator {...rest} />
   </Box>
 );
+
 export const Legend = ({ mode, onClick, values, ...rest }) => {
   const [active, setActive] = useState();
   const theme = useContext(ThemeContext);
--- a/aries-site/src/examples/templates/dashboards/components/Legend.js
+++ b/aries-site/src/examples/templates/dashboards/components/Legend.js
@@ -2,17 +2,23 @@ import { Fragment, useContext, useState } from 'react';
 import PropTypes from 'prop-types';
 import { Box, Button, List, Text, ThemeContext } from 'grommet';
 
+// apply rounding to Button corners by wrapping in Box
+const LegendButton = ({ ...rest }) => (
+  <Box round="xxsmall" overflow="hidden">
+    <Button hoverIndicator {...rest} />
+  </Box>
+);
 export const Legend = ({ mode, onClick, values, ...rest }) => {
   const [active, setActive] = useState();
   const theme = useContext(ThemeContext);
   const size = theme.global.edgeSize.small;
 
-  const Wrapper = onClick ? Button : Fragment;
+  const Wrapper = onClick ? LegendButton : Fragment;
 
   return (
     <List
       data={values}
-      defaultItemProps={{ pad: { vertical: 'xxsmall' } }}
+      defaultItemProps={{ pad: { vertical: !onClick ? 'xxsmall' : 'none' } }}
       {...rest}
     >
       {datum => (
@@ -24,7 +30,6 @@ export const Legend = ({ mode, onClick, values, ...rest }) => {
                   setActive(active !== datum ? datum : undefined);
                 },
                 active: active === datum,
-                hoverIndicator: true,
               }
             : {})}
         >
--- a/aries-site/src/examples/templates/dashboards/content/chartCards/CostCompact.js
+++ b/aries-site/src/examples/templates/dashboards/content/chartCards/CostCompact.js
@@ -167,7 +167,7 @@ export const CostCompact = ({ period }) => {
             alignSelf="center"
             gridArea="legend"
             mode="compact"
-            onClick={datum => setActive(active !== datum ? datum : undefined)}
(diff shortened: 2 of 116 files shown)


aries-site/src/examples/templates/dashboards/content/chartCards/CostCompact.js
```diff
@@ -0,0 +1,179 @@
...
+      xsmall: 'medium',
+      small: 'medium',
+      medium: 'small',
+      large: 'medium',
+      xlarge: 'medium',
+    },
+  };
+
+  return (
+    <ChartCard title="Cost" subtitle="Consumption Analytics">
```
britt6612, line 123: are we not adding the icons as part of the title on the right side ?

britt6612:
seems like the `gap` between the subtitle and `meter` is bigger in figma looks like it is `48px` and with the site it looks like we have `36px` between
Screen Shot 2022-02-23 at 3 44 21 PM

aries-site/src/examples/templates/dashboards/content/chartCards/CostCompact.js
taysea: At the moment we are not since we don't have any guidance/research are what those buttons would do.

aries-site/src/examples/templates/dashboards/content/chartCards/CostCompact.js
```diff
@@ -0,0 +1,179 @@
...
+                }))}
+                size="full"
+                type="circle"
+              />
+            </Box>
+            <Measure
+              alignSelf="center"
+              align="center"
+              direction="column-reverse"
+              name={{ label: { label: period, size: 'small' } }}
```
britt6612, line 157: I believe this should be `size: 'xsmall` per the designs

aries-site/src/examples/templates/dashboards/content/chartCards/CostCompact.js
```diff
@@ -0,0 +1,179 @@
...
+              }}
+            >
+              <Meter
+                values={values.map((value, index) => ({
+                  label: value.label,
+                  value: value.value,
+                  color: {
+                    color: chartColors[index % chartColors.length],
+                    opacity:
+                      !active || (active && active === value) ? '100%' : 'weak',
```
britt6612, line 146: I really like how the opacity looks nice work 👍

aries-site/src/examples/templates/dashboards/content/chartCards/CostCompact.js
britt6612: got it good to know! thanks for clarifying

aries-site/src/examples/templates/dashboards/components/Legend.js
```diff
@@ -1,30 +1,57 @@
...
+                },
+                active: active === datum,
+                hoverIndicator: true,
+              }
+            : {})}
+        >
+          <Box
+            direction="row"
+            gap="small"
+            justify="between"
```
britt6612, line 34: wondering if we need this the `justify="between`

taysea: For this example we don't, but for other cards that are using this Legend we do.

britt6612: okay that's fair I was just trying to see why we had the extra prop

aries-site/src/examples/templates/dashboards/content/chartCards/CostCompact.js
taysea: Done.

taysea:
> seems like the `gap` between the subtitle and `meter` is bigger in figma looks like it is `48px` and with the site it looks like we have `36px` between Screen Shot 2022-02-23 at 3 44 21 PM

Fixed.

britt6612 (approved):
Looks good thank you for the clean up

aries-site/src/examples/templates/dashboards/content/chartCards/CostCompact.js
```diff
@@ -0,0 +1,181 @@
...
+  Stack,
+  ThemeContext,
+} from 'grommet';
+import { parseMetricToNum } from 'grommet/utils';
+import { ChartCard, Legend, Measure } from '../../components';
+import { defaultWindow, formatCurrency, REPORT_WINDOW_MAP } from './utils';
+
+const MOCK_DATA = require('../../../../../data/mockData/consumption.json');
+
+export const CostCompact = ({ period }) => {
```
halocline, line 17: Since this is variant presentation of the CostByService, would it make sense to name them so that they appear together in the directory? Open to names, but would like them to share something in common so they are easily identifiable as being related.

```suggestion
export const CostByServiceCompact = ({ period }) => {
```

aries-site/src/examples/templates/dashboards/components/Legend.js
```diff
@@ -1,30 +1,57 @@
...
+            pad={
+              onClick
```
halocline, line 36: Should we consider adding a similar onClick conditional to the `defaultItemProps={{ pad: { vertical: 'xxsmall' } }}`? Set it to 'none' when onClick? This would keep Legend spacing consistent.

aries-site/src/examples/templates/dashboards/components/Legend.js
```diff
@@ -1,30 +1,57 @@
...
       data={values}
       defaultItemProps={{ pad: { vertical: 'xxsmall' } }}
       {...rest}
     >
       {datum => (
-        <Box direction="row" gap="small" justify="between">
-          <Box direction="row" align="center" gap="xsmall">
-            <Box background={datum.color} height={size} width={size} round />
-            <Text>{datum.label}</Text>
+        <Wrapper
```
halocline, line 19: Should we add some border-radius to soften the corners and make harmonious with the rest of our rounded corners on hover/active?

aries-site/src/examples/templates/dashboards/content/chartCards/CostCompact.js
```diff
@@ -0,0 +1,181 @@
...
+              }}
+              justify="center"
+              width="xsmall"
+            />
+          </Stack>
+          <Legend
+            alignSelf="center"
+            gridArea="legend"
+            mode="compact"
+            onClick={datum => setActive(active !== datum ? datum : undefined)}
```
halocline, line 170: To create the tie/reference between values prop.

```suggestion
            onClick={value => setActive(active !== value ? value : undefined)}
```

aries-site/src/examples/templates/dashboards/content/chartCards/CostCompact.js
```diff
@@ -0,0 +1,181 @@
...
+              name={{ label: { label: period, size: 'xsmall' } }}
+              value={{
+                value: formatCurrency(active ? active.value : totalCost),
+                size: 'xlarge',
+              }}
+              justify="center"
+              width="xsmall"
+            />
+          </Stack>
+          <Legend
```
halocline, line 166: Thinking about UX. I like the ability to click and reveal that item's value. It wasn't immediately clear that the click functioned as a toggle. I was trying to restore to "show all" by clicking away from any of the options, but that didn't restore the state.

Trying to think of ways to address. If we did, what would the appropriate click target be? The entire card the region containing the Meter + Legend?

aries-site/src/examples/templates/dashboards/content/chartCards/CostCompact.js
```diff
@@ -0,0 +1,181 @@
...
+              }}
+            >
+              <Meter
+                values={values.map((value, index) => ({
+                  label: value.label,
+                  value: value.value,
+                  color: {
+                    color: chartColors[index % chartColors.length],
+                    opacity:
+                      !active || (active && active === value) ? '100%' : 'weak',
```
halocline, line 146: I played with other opacity values. 'medium' didn't leave enough contrast, so I think 'weak' was an appropriate choice.

aries-site/src/examples/templates/dashboards/content/chartCards/CostCompact.js
taysea: I like this, will update.

aries-site/src/examples/templates/dashboards/components/Legend.js
taysea: Yes, I'll do that.

aries-site/src/examples/templates/dashboards/content/chartCards/CostCompact.js
taysea: Right, I could understand some confusion. I explore a few interactive legends and the majority followed the toggle approach which is why I followed that.

We can continue to ideate though on what might be a better pattern.

aries-site/src/examples/templates/dashboards/content/chartCards/CostCompact.js
taysea: Yes, I felt the same.

aries-site/src/examples/templates/dashboards/components/Legend.js
taysea: Yes, have adjusted.

halocline (approved):

PR #2432, opened 2022-03-07
Use new Page and PageContent components on site

taysea:
[Deploy Preview](https://deploy-preview-2432--keen-mayer-a86c8b.netlify.app/components/button)

#### What does this PR do?

Updates DS site layout to use the new Page and PageContent components. Adds necessary theme overrides to keep horizontal padding consistent with current layout.

**General UX Checks**
- [ ] Light & dark modes

**Accessibility Checks**
- [ ] Keyboard interactions

**Code Quality Checks**

### DIFF
--- a/aries-site/src/pages/index.js
+++ b/aries-site/src/pages/index.js
@@ -4,6 +4,7 @@ import {
   Card,
   Grid,
   Heading,
+  PageContent,
   Paragraph,
   ResponsiveContext,
   Stack,
@@ -33,17 +34,17 @@ const calcPad = size => {
 };
 
 // These make a box width limited to xxlarge but centered
-const widthProps = { width: { max: 'xxlarge' }, margin: 'auto' };
+const widthProps = { margin: 'auto' };
 
 const Intro = ({ children }) => {
   const size = useContext(ResponsiveContext);
   return ['xsmall', 'small'].includes(size) ? (
-    <Box>
+    <PageContent>
       <Hero height={{ max: '292px' }} margin={{ bottom: '-24px' }} />
       <Card background="none" elevation="none">
         {children}
       </Card>
-    </Box>
+    </PageContent>
   ) : (
     <Stack guidingChild="last">
       <Box
@@ -57,7 +58,7 @@ const Intro = ({ children }) => {
         <Box width="small" />
         <Hero />
       </Box>
-      <Box height={{ min: 'medium' }} justify="center" {...widthProps}>
+      <PageContent height={{ min: 'medium' }} justify="center" {...widthProps}>
         <Grid
           gap="large"
           columns={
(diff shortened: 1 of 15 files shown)


ericsoderberghp:
Can you provide a little more detail on?
> Adds necessary theme overrides to keep horizontal padding consistent with current layout.

I was hoping that the theme would work as-is.

jcfilben:
> I was hoping that the theme would work as-is.

I was expecting we would need to change the `min` width for the hpe theme (see [relevant discussion](https://github.com/grommet/grommet/pull/5960#discussion_r810356646)) as well as adding horizontal for the 'xsmall' and 'xlarge' breakpoints. However I wasn't expecting we would need to override the horizontal padding for the 'small', 'medium', and 'large' breakpoints

jcfilben:
Looks like we need extra padding under the roadmap card on the [extend page](https://deploy-preview-2432--keen-mayer-a86c8b.netlify.app/extend)

Screen Shot 2022-03-08 at 10 50 24 AM

taysea:
> However I wasn't expecting we would need to override the horizontal padding for the 'small', 'medium', and 'large' breakpoints

Good comments @jcfilben -- this is a discussion point I want to bring up with the team. If we stick with the paddings we want for app page layouts, it makes the sites layout much wider (not necessarily a bad thing, but a change). I will post something in staff channel.

aries-site/src/themes/aries.js
```diff
@@ -51,6 +51,30 @@ export const aries = deepMerge(hpe, {
...
+      },
+      medium: {
+        pad: { horizontal: 'large' },
+      },
+      large: {
+        pad: { horizontal: 'xlarge' },
+      },
+      xlarge: {
+        pad: { horizontal: 'xlarge' },
+      },
```
halocline, line 75: Are these nested too deep? They are currently a child of 'wide'. Shouldn't they be a child of 'page'?

jcfilben: This is aligned with the current theme structure of Page. Padding is defined within page[kind][size] in addition to any other box props.

aries-site/src/pages/index.js
```diff
@@ -88,7 +83,7 @@ const Index = () => {
       <Meta title={title} description={pageDetails.seoDescription} />
       <Box>
         <Intro>
-          <Box pad={{ horizontal: calcPad(size) }} width={{ max: '900px' }}>
+          <Box width={{ max: '700px' }}>
```
halocline, line 86: Can we add a comment as to how this value was arrived at?

Also, the `xxxsmall` value below looks suspect. I don't believe we have a value for that.

taysea:
@jcfilben @halocline I have been using [v2](https://github.com/grommet/hpe-design-system/pull/2433) for cleanups and the direction to use theme as is. I've incorporated all the comments here in the new PR.

Issue #2426, opened 2022-03-05
Page Layouts - update design system site to use new grommet components


ericsoderberghp:
Update site examples:
- [x] Content Layouts (1)
- [x] Dashboards (3)
- [x] DataTable Customization (1)
- [x] Tags on a Detail Page (1)

Adjusting the actual site to use page/page content (Taylor)

Make Changes in HPE theme

jcfilben:
Site examples have been updated, the site is using Page and PageContent components, and the hpe theme has been updated with the correct width for page

PR #2439, opened 2022-03-09
Update Relevant Site Examples to use Page and PageContent Components

jcfilben:
[Deploy Preview](https://deploy-preview-2439--keen-mayer-a86c8b.netlify.app/)

#### What does this PR do?
Updates the relevant design system site examples to use Page and PageContent.
Examples updated:
- Content Layouts
- Dashboards
- DataTable Customization
- Tags on a Detail Page

**General UX Checks**
- [x] Light & dark modes

**Accessibility Checks**
- [ ] Keyboard interactions

**Code Quality Checks**

#2426 

No

Backwards compatible


aries-site/src/examples/components/tag/TagResource.js
```diff
@@ -37,57 +39,59 @@ const SectionDetails = ({ data }) => (
...
-                  // eslint-disable-next-line no-alert
-                  alert(
-                    // eslint-disable-next-line max-len
-                    `Clicking on '${name}: ${value}' tag could take a user to a filtered view displaying all devices with '${name}: ${value}' applied.`,
-                  )
-                }
-              />
-            );
-          })}
+      <Box alignSelf="center" gap="large" pad="medium" flex={false}>
```
taysea, line 47:
```suggestion
      <Box alignSelf="center" gap="large" flex={false}>
```
Since PageContent is applying padding should we remove?

aries-site/src/examples/components/tag/TagResource.js
```diff
@@ -37,57 +39,59 @@ const SectionDetails = ({ data }) => (
...
-        <Box>
-          <Heading size="small" margin="none">
-            Store C-3P0
-          </Heading>
-          <Text size="large">Aruba 530 32930F</Text>
-        </Box>
-        <Button icon={<More />} />
+  <Page kind="narrow">
+    <PageContent>
+      <Box align="start" pad="medium">
```
taysea, line 44: Wondering if we need this Box at all -- can we have `alignSelf="start"` on the Button itself and remove the Box all together. If we need spacing between the button and the heading can we wrap those in a Box and add gap?

aries-site/src/examples/templates/dashboards/ThreeColumnDashboard.js
```diff
@@ -1,21 +1,30 @@
...
+import { AppContainer } from '../page-layouts/components';
 import { ContentArea } from '../page-layouts/anatomy/components';
-import { PageHeader } from './components';
 import { CostByMonth, CostByService, CostByYear, RulesAudit } from './content';
 
 export const ThreeColumnDashboard = () => (
   <AppContainer background="background-back">
     <ContentArea title="Global Header" />
-    <PageContainer>
-      <PageHeader title="Dashboard" />
```
taysea, line 16: Any reason we are removing PageHeader as opposed to adjusting PageHeader.js to align with the needs for these examples?

aries-site/src/examples/components/tag/TagResource.js
jcfilben: We do need the box otherwise the button stretches across the width of the PageContent

aries-site/src/examples/templates/dashboards/ThreeColumnDashboard.js
jcfilben: It felt like having a separate PageHeader file wasn't necessary all it would be is
```
<Header>
  <Heading>
    Page Title
  </Heading>
</Header>
```

halocline: There are a couple of other considerations:
1. We want to move towards building a browse-able library of component templates whether they are at the organism level or higher
2. For maintenance and consistency, we want to move away from one-off implementations which tend to drift over time. Instead we'd like to move towards component re-use to enforce consistency, even if they are trivial components at this time, we can build atop of them as needs and usage evolves. Heading this direction becomes the basis for which feeds the template library ☝️ .
3. There is also an educational component where we are trying to focus on presenting a particular concept in the example. Sometimes the focus on that concept is diffused if there are too many elements or moving pieces. It becomes harder to see the forrest (big picture) through the trees (extra markup, inline logic, etc.). Sometimes abstracting helps reduce the cognitive load and is a more effective educational mechanism.

aries-site/src/examples/components/tag/TagResource.js
halocline: I think this should work, right?

```
 <Button alignSelf="start" label="Devices" icon={<FormPrevious />} />
```

taysea: ^ Yes that is what I was proposing

aries-site/src/examples/templates/dashboards/ThreeColumnDashboard.js
jcfilben: Gotcha, thanks for the explanation! I will move the Header back into the PageHeader file

aries-site/src/examples/components/tag/TagResource.js
```diff
@@ -37,57 +39,57 @@ const SectionDetails = ({ data }) => (
...
-      >
-        <Box>
-          <Heading size="small" margin="none">
-            Store C-3P0
-          </Heading>
-          <Text size="large">Aruba 530 32930F</Text>
+  <Page kind="narrow">
+    <PageContent gap="small">
+      <Button alignSelf="start" label="Devices" icon={<FormPrevious />} />
+      <Box alignSelf="center" gap="large" flex={false}>
```
taysea, line 45:
```suggestion
      <Box gap="large" flex={false}>
```
Can we remove this? seems like narrow page should take care of width and centering.

halocline (approved):
Looks good, @jcfilben.

It seems like Page and PageContent are behaving as desired. Did you find anything concerning while implementing?

jcfilben:
> Did you find anything concerning while implementing?

I didn't encounter anything concerning

taysea (approved):
Nice work!

PR #2472, opened 2022-03-22
fix code links add figma button

britt6612:

#### What does this PR do?
fix code links
adds Figma button
global-banner-notifications.mdx
site

**General UX Checks**
- [ ] Light & dark modes

**Accessibility Checks**
- [ ] Keyboard interactions

**Code Quality Checks**

site
the links for the code were incorrect 
Kenny asked if we can add a link to Figma similar to how it is done in Page Layouts
none
no
backwards compatible

### DIFF
--- a/aries-site/src/pages/templates/global-banner-notifications.mdx
+++ b/aries-site/src/pages/templates/global-banner-notifications.mdx
@@ -1,4 +1,4 @@
-import { Anchor } from 'grommet';
+import { Anchor, Button } from 'grommet';
 import {  BestPracticeGroup, Example } from '../../layouts';
 import {
   BannerContentLayoutExample,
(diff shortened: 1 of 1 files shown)


aries-site/src/pages/templates/global-banner-notifications.mdx
```diff
@@ -12,6 +12,15 @@ import {
   BannerNotificationDiagram,
 } from '../../examples';
 
+<Button
+  icon={<Figma color="plain" />}
+  label="Open in Figma"
+  secondary
+  alignSelf="start"
+  href="https://www.figma.com/file/FGvOd2htjrChVVS2oz73ds/HPE-Notifications?node-id=3597%3A16611"
```
taysea, line 20: Shouldn't we add this as a `figma` prop to the first example instead? That has been our typical convention.

britt6612: I can add it here as well @KennyAtHPE would that be fine? was there any particular reason you wanted the button separate or can we add it to the example

KennyAtHPE: Hm, I'm thinking that having it at the top similar to the page layouts will be the easiest for users to see it, but if standard practice of having it on the example works then that works too

britt6612: I added it to the example how we have in our other example. I agree since this is a standard practice also see how it can be missed. I'm wondering if this was part of the user study if they were able to find the figma link for each page something to think about.

taysea (approved):

Issue #2453, opened 2022-03-15
Add Video to Accessibility Page


amandadupell:
### What this ticket does
- adds Bill's new video to the accessibility page in the design system

Issue #1816, opened 2021-08-31
Accessibility Page Improvements


amandadupell:
### Is this issue related to the design system core library or design system website?

Design System website

### Small changes
- [ ] update Developer and Designer Guidance links (current ones are pointing to incorrect page)
- [ ] add external links to Software Design and Test Resources from [Product Accessibility Office](https://hpe.sharepoint.com/sites/f5/cto/office/Accessibility/Pages/accessible-web-software-and-hardware-design.aspx) to Resources section
- [ ] add government resources regarding laws surrounding accessibility (can be found under the [Governing Standards and Regulations](https://hpe.sharepoint.com/sites/f5/cto/office/Accessibility/Pages/accessible-web-software-and-hardware-design.aspx) section) to Resources section

### Component page-level changes
- [ ] for guides that outline accessibility guidance for certain components, include those direct links to the accessibility section for that component (i.e. this guide is specific to [creating tables](https://hpe.sharepoint.com/sites/F5/CTO/Office/Accessibility/Documents/Forms/AllItems.aspx?id=%2Fsites%2FF5%2FCTO%2FOffice%2FAccessibility%2FDocuments%2FCreating%20Accessible%20Tables%20in%20software%5Ffinal%5FMarch%202020%2Epdf&parent=%2Fsites%2FF5%2FCTO%2FOffice%2FAccessibility%2FDocuments&p=true&originalPath=aHR0cHM6Ly9ocGUuc2hhcmVwb2ludC5jb20vOmI6L3MvRjUvQ1RPL09mZmljZS9BY2Nlc3NpYmlsaXR5L0VheXJ6Y0Qxc3UxR3B0STQwSXVlWE9RQkc3WFRMZ05nb05LN1Y4RVJ0ZzdkLVE_cnRpbWU9QjM5RDhLaHMyVWc) and can be included on the Table component page)

### Section changes
- [ ] move Accessibility Guides from Resources bullet point list to its own section with direct links to the guide (this section maybe to include the three most recent guides written with some indication to highlight the newest one?) 
[...]

Issue #1821, opened 2021-09-02
General Accessibility Audit


amandadupell:
# Purpose of this issue

The purpose of this issue is to identify keyboard/ARIA accessibility improvements that can be made to the Design System UI/UX.

## How were these improvements identified?

Most of the issues came from looking at the Design System pages and taking note of the experience when using a screen reader.

### Actions
- [ ] most onClick handlers should include an announcement that an action has been performed or that there was a change in UI/UX, etc.

For example, when clicking on a card, it should be announced to the user that they are being directed to a new page that card links to. If the user clicks on the Button card, the screen reader should relay something like "opening Button page", etc.

<img width="250" alt="card" src="https://user-images.githubusercontent.com/38971752/131908464-1bc01807-444d-4bd0-b4db-2e6e4fa581f6.png">

### Sections
- [ ] if there is related content under a section, the user should get some sort of indication that the content is related to that section and to the other content within that section

For example, when entering the Philosophy section, the screen reader could relay that the user is entering a section with x amount of cards. When the user accesses a card under the Philosophy section, it should be clear that the card relates to that section.

<img width="500" alt="section" src="https://user-images.githubusercontent.com/38971752/131908833-3531addb-3b74-424a-aae4-74f90abf6462.png">

### Component and Code Examples
- [ ] at the page-level for components, there should be an understanding of how the content of the page relates to the coding examples
- [ ] there should be a clear distinction between the content of the page and the examples
- [ ] there should be a clear distinction between the "playground" where the component example lives and the footer of the example where the corresponding links can be found

For example, for this checkbox, the example is directly related to creating a CheckBox with a description. When the user enters the "playground", it should be relayed to the user that this example is related to the above content. When the user exits the "playground" to access the links in the footer, it should be relayed that they have exited the example and are now interacting with links related to the example.

<img width="500" alt="code example" src="https://user-images.githubusercontent.com/38971752/131909956-5009dcd4-b77b-4454-b981-c13e787567b5.png">

PR #2477, opened 2022-03-23
Feat/access-video: adding video to accessibility page

amandadupell:
#### What does this PR do?

Adds Bill's video to Accessibility page

Accessibility.mdx

Manual local testing

**General UX Checks**
- [X] Small, medium, and large screen sizes
- [X] Cross-browsers (FireFox, Chrome, and Safari)
- [X] Light & dark modes
- [X] All hyperlinks route properly

**Accessibility Checks**
- [X] Keyboard interactions
- [X] Screen reader experience
- [X] Run WAVE accessibility plugin (Chrome)

**Code Quality Checks**
- [X] Console is free of warnings and errors

Deployment

The paragraph before the video serves as an audio description for blind or visually impaired users. It also provides context for what the video's purpose is and who Bill is.

#1816 #2453 

Yes!

Compatible


aries-site/src/pages/foundation/accessibility.mdx
```diff
@@ -4,8 +4,13 @@ import {
   VideoSection 
 } from '../../examples/foundation';
 
+import { AccessVideo } from '../../components/accessibility';
+
 At HPE, we believe in being a force for good. We solve for humanity with humanity. We use technology to make the world better. We are unconditionally inclusive. **We believe that being accessible is a requirement, not an option**.
 
+In the video below, Bill Tipton, an HPE Product Accessibility Champion, discusses his role and the importance of inclusive design and development. The video shows Bill, a white man with blondish brown hair and a mustache, in his office wearing a long-sleeved lavender-blue dress shirt with the HPE rectangular-shaped, multi-color logo against a black background as the background.
```
halocline, line 11: The second sentence seems like it belongs in an aria-describedby and `<div />` added as a child of `<video />`. Is there a reason it should be in the main content flow?

Reference: https://www.w3.org/WAI/PF/HTML/wiki/Media_Alt_Technologies#2:_Use_.40aria-describedby_for_the_longer_text_alternative

aries-site/src/components/accessibility/Video.js
```diff
@@ -0,0 +1,23 @@
+import React from 'react';
+
+import { Box } from 'grommet';
+
+export const AccessVideo = () => {
+
+  return (
+    <Box width="large" pad={{ bottom: 'large' }}>
+      <video controls>
```
halocline, line 9: Should we be using the `transcript` attribute and providing a transcription html file?

https://www.w3.org/WAI/PF/HTML/wiki/Media_Alt_Technologies#3:_Use_.40transcript_for_providing_a_link_to_a_file_with_the_full_transcription

aries-site/src/pages/foundation/accessibility.mdx
amandadupell: bill and i had discussed this a way to provide an audio description for users but i think i could incorporate it with the aria-describedby as you mentioned and see what bill's experience is with that! that way it isn't unnecessary content on the page

aries-site/src/components/accessibility/Video.js
amandadupell: yes! will be adding this today

amandadupell: just a heads up, i added the transcript file via a clickable link after the video that will open the html page. bill says this aligns with current best practice as a user will could save this page as a pdf if they wanted to digest the transcript offline.

i don't think the `transcript` attribute is implemented in HTML5 but is just an idea at the moment for future improvements to the `<video />`

aries-site/src/components/accessibility/Video.js
```diff
@@ -0,0 +1,50 @@
...
+  const announce = useContext(AnnounceContext);
+
+  return (
+    <>
+      <Box 
+        gap="small" 
+        width="large" 
+        pad={{ bottom: 'large' }}
+      >
+        <div id="video-alt" style={{ display: 'none' }}>
```
halocline, line 15: I would have expected this div to be nested in the `<video>` tag after the `<track>`. Is there a reason for having it outside?

aries-site/src/components/seo/Meta.js
```diff
@@ -20,7 +20,7 @@ export const Meta = ({
   const csp = `default-src 'self' 'unsafe-eval'; 
   style-src 'self' *.hpe.com/hfws-static/5/css/ 'unsafe-inline';
   connect-src 'self' *.githubusercontent.com/grommet/hpe-design-system/ https://www.google-analytics.com https://www.github.com/grommet/ https://eyes.applitools.com *.hpe.com/hpe/api/;
-  media-src 'self' https://d3hq6blov2iije.cloudfront.net/media/HPE+Design+System-v3.mp4;
+  media-src 'self' https://d3hq6blov2iije.cloudfront.net/media/HPE+Design+System-v3.mp4 https://d3hq6blov2iije.cloudfront.net/media/BILL-ACCESS-VID.mp4;
```
halocline, line 23: Does this achieve the same effect?

```suggestion
  media-src 'self' https://d3hq6blov2iije.cloudfront.net/media/;
```

halocline (commented):
Looking really good! The screenreader experience was quite nice. Would you be willing to demo this to the team?

A couple of small cleanup items.

aries-site/src/components/accessibility/Video.js
amandadupell: When testing this myself and Bill with the div inside the `<video />`, the information wasn't being read aloud by the screen reader. It was only relayed with this div being outside of the `<video />`.

For my most recent changes, you will see that I am using `aria-description` instead of `aria-describedby`, as Bill was still having some issues with getting the description read aloud using JAWS. They serve the same purpose, but have different levels of support/implementation. Using `aria-description` works for me on VO and I am hoping Bill has a better experience with JAWS.

aries-site/src/components/accessibility/Video.js
```diff
@@ -0,0 +1,72 @@
...
+  const [expanded, setExpanded] = useState(false);
+
+  return (
+    <>
+      <Box 
+        gap="small" 
+        width="large" 
+        pad={{ bottom: 'large' }}
+      >
+        <Box
```
halocline, line 24: Why are we using a Box here? This seems to be acting a button.

aries-site/src/components/accessibility/Video.js
```diff
@@ -0,0 +1,72 @@
...
+
+  return (
+    <>
+      <Box 
+        gap="small" 
+        width="large" 
+        pad={{ bottom: 'large' }}
+      >
+        <Box 
+          a11yTitle="Audio Desciption"
```
halocline, line 25:
```suggestion
          a11yTitle="Audio Description"
```

aries-site/src/components/accessibility/Video.js
```diff
@@ -0,0 +1,72 @@
...
+          }} 
+        >
+          Audio Description
+          {
+            expanded ? 
+              <FormUp a11yTitle="Up arrow" /> : 
+              <FormDown a11yTitle="Down arrow" />
+          }
+        </Box>
+        {expanded && desc}
```
halocline, line 42: This pattern of having the audio description visibly ahead of the video is new to me. Can you share any resources, examples, etc. which are inspiring this direction?

aries-site/src/components/accessibility/Video.js
```diff
@@ -0,0 +1,72 @@
...
+      >
+        <Box 
+          a11yTitle="Audio Desciption"
+          aria-description="Click to show audio description"
+          gap="small" 
+          direction="row" 
+          onClick={() => {
+            setExpanded(!expanded);
+            expanded ? announce("Hid audio description") : 
+              announce("Opened audio description");
```
halocline, line 32: I'm liking the use of announce.

aries-site/src/components/accessibility/Video.js
halocline: For my education, can you share any resources speaking to `aria-description` vs `aria-describedby`? I'd like to better understand why `aria-describedby` wasn't working. Is is fundamentally flawed? Was it a flaw in our implementation? It is so widely referenced and documented, that it seems like we should deeply understand this so that we are aware of pitfalls and can recognize and abate them.

aries-site/src/components/accessibility/Video.js
amandadupell: sure! after working with bill with numerous testing rounds, he had shared that the audio description through `aria-description` and `aria-describedby` was not consistent in the reading of the description. while when testing with VO i was able to hear the description at the video start and end, he was not able to get it read consistently with JAWS or NVDA.

to ensure that the description would be read aloud at some point, bill mentioned that another pattern he has seen before is to provide a way on the same page as the video for screen readers to get the description, either as hidden text on the page or a button that triggered the description to be shown.

in response to why `aria-description` and `aria-describedby` were flawed, it seems that the `<video />` doesn't have great screen reader support with related props as shown [here](https://www.powermapper.com/tests/screen-readers/media/video-aria-labelledby/) and [here](https://www.powermapper.com/tests/screen-readers/media/video-aria-label/). `aria-describedby` also seems to be [mostly used with form elements,](https://www.powermapper.com/tests/screen-readers/labelling/img-aria-describedby/) but when used with other elements like ``, the support changes drastically as shown [here](https://www.powermapper.com/tests/screen-readers/labelling/img-aria-describedby/). the same could be said for `aria-decription` since it functions the same way.

based on my testing with VO and with bill with JAWS and NVDA, it seems there is varied support around both of these aria attributes and so we took the approach of displaying the audio description in some way on the screen so users with screen readers can find it. we also included the `a11yTitle` for the audio transcript file to relay that an audio description could also be found there, which is [common practice](https://www.3playmedia.com/learn/popular-topics/audio-description/#formats).

taysea: Feels like we should have guidance on DS site of how to implement this. I don't know if we have a "Video" component page, but this kind of information would be helpful and important to share there -- even if videos are less common UI elements for our apps.

taysea: Couple of questions on the approach though:
- I'm wondering if the audio description should go beneath the video instead? It feels a bit too prominent. Also it feels more like a "Video Description" to me than an audio description, because we are describing the visuals.
- Could we potentially mirror the design approach that Ted Talks takes (showing the transcript inline on the page) -- maybe for our use we can mirror the "Show Code" approach but instead have "Read Transcript" as the collapsible label: Example on TedX approach when you click "Read Transcript" [https://www.ted.com/talks/jill_bolte_taylor_my_stroke_of_insight](https://www.ted.com/talks/jill_bolte_taylor_my_stroke_of_insight)
- The "Audio Description" is already repeated in the transcript. Is there any reason we need to have both?

amandadupell: Agreed with having guidance reflecting these Video considerations on the DS. To your other points:

- the terms "audio description" and "video description" are used interchangeably in the a11y community and usually come before the video to properly set up the context of the video. i had it as "audio description" because we are describing the video through audio for screen reader use and bill had mentioned that was the terminology he was used to. i can change to "video description" if you feel that's more appropriate!
- i had originally thought about the "show code" approach and making use of that component/pattern as it exists in the DS, but my big thing there was the accessibility of the code snippet piece as logged in [this issue](https://github.com/grommet/hpe-design-system/issues/1821). the other piece to this was the fact that transcripts should allow for a user to download the file, as pointed out by bill. i think it could be worth it to explore how we could add the inline transcript and allow for downloading the file in our Video component. i could approach this in the same way that i'm handling the expandable description if that would be better but bill noted his experience currently is accessible and up-to-standard.
- the "audio description" is included in the transcript file as it should be accessible at the page level that the video is on and included in the file itself.

let me know your thoughts!

taysea: I am very much in support of ensuring all the necessary accessibility requirements are met so those using assistive technologies can have an equal experience to those not. That being said, I want to be mindful that the assistive technology support doesn't clutter the UI or disrupt the usability of a page for users not relying on the assistive technologies, as this could create a poor experience in the opposite direction. That's where some of my design comments are coming from.

Personally, I found the "Audio Description" button to be a bit disruptive to the page flow because it comes first and draws a lot of attention to itself, which is why I'd be more in favor of it coming below the video. Good note about code snippet issue, I'm fine with exploring it implemented how you handled it as opposed to the show code approach but maybe placing underneath the video?

For the case of needing to be able to download the transcript, maybe we can have the transcript appear inline but also provide a download button for those who would want to download it? I think an icon-only button would be sufficient with an `a11yTitle="Download transcript"` ?

amandadupell:
@taysea @halocline for the latest changes, i've moved and renamed the audio description button to be below the video and now read as "video audio description". i've also changed the transcript file to be a button as well to align the UI and also help screen readers identify that an action will be triggered. 

as far as the expandable transcript, i think it would be worth it to explore the inline implementation. for the sake of getting this published, i wanted to propose opening a ticket adding this functionality, as it would require another round of user testing with Bill to ensure it's an accessible experience. this is something i can pick up as a "version 2", but i didn't want that to get in the way of publishing this "version 1" since the accessibility of this UI has full coverage and more testing may postpone the release. does this sound reasonable?

halocline:
> as far as the expandable transcript, i think it would be worth it to explore the inline implementation. for the sake of getting this published, i wanted to propose opening a ticket adding this functionality, as it would require another round of user testing with Bill to ensure it's an accessible experience. this is something i can pick up as a "version 2", but i didn't want that to get in the way of publishing this "version 1" since the accessibility of this UI has full coverage and more testing may postpone the release. does this sound reasonable?
> 

After your share-out last week and subsequent discussions, I propose the following:
- Move forward with the current direction, but apply a couple presentation tweaks which should not interfere with the keyboard nor screen reader flow. Suggested adjustments below.
- This is temporary as we focus on improving the Grommet Video component. I believe folks are gravitating towards including the description and transcript as part of the video controls.

Proposed adjustments:
- Change button label to read "Video Description." We can add "Audio description of video" as an `a11yTitle` if we think the "audio description" piece is valuable.
- Present buttons in `<Box direction="row" ... />` as opposed to column
- Style Box with `background-front`, rounded bottom corners, and appears as part of the video player as opposed to floating elements. The result should resemble how the "Example controls" appear as a row of controls related the Example container.
Screen Shot 2022-04-12 at 8 52 32 AM
- Transcript file HTML should receive the HPE Theme styling. I think we execute this as an .mdx file living under `/foundation`. My mind is rusty if we'd need adjustments to our page /data/structures and/or router...

@taysea , @amandadupell do these proposals and direction resonate? Thoughts? Adjustments?

taysea:
Agree with your recommendations @halocline. And nice work on some of these adjustments @amandadupell. Two things I'm noticing which I think would be nice to resolve before merging (but I think we're very close):
1. The accessibility transcript file appears in the search (and given its alphabetical order is one of the first items). Can we remove it from the search by chance?
Screen Shot 2022-04-12 at 9 26 57 PM

2. When I expand the video description, the "Transcript File" button moves to the bottom. @halocline is your expectation that it would remain in a row with the other button? or is this behavior fine for this pass?

Screen Shot 2022-04-12 at 9 26 11 PM

halocline:
> 1. The accessibility transcript file appears in the search (and given its alphabetical order is one of the first items). Can we remove it from the search by chance?

We can remove. I suggest we add `searchable: false` to the page object in `/structures/foundation.js`. Then modify [this line](https://github.com/grommet/hpe-design-system/blob/d5007a2358e4d1be78e8428e2cb2f82feac3db2b/aries-site/src/layouts/navigation/Search.js#L10) by inserting a filter before the map. 

> 2. When I expand the video description, the "Transcript File" button moves to the bottom. @halocline is your expectation that it would remain in a row with the other button? or is this behavior fine for this pass?

I would expect the "Transcript File" to maintain its position in the row.

halocline (approved):
🎉

taysea (approved):
Great! Thank you for all the refinements here.

Issue #2167, opened 2021-12-21
Icons - add core icons to Icons page


ericsoderberghp:
Take the outcome of the core icon research so far and put a stake in the ground for what the core icons by adding a section to the Icons page in the design system site.

- pick a direction for which icons should be "core" and which specific icons should be used for them
- for each icon, include at least a visual, a primary name, any aliases to help with searching, and some guidance on its usage
- if needed, redesign how the Icons page is structured
- request feedback from the design system community, slack channel, What's New, email newsletter

PR #2515, opened 2022-04-08
Added initial Core Icons

ericsoderberghp:
[Deploy Preview](https://deploy-preview-2515--keen-mayer-a86c8b.netlify.app/foundation/icons#core-icons)

#### What does this PR do?

Adds a Core Icons section to the Icons page, based on Chris Conetta's research. This isn't intended to cover all potential core icons, just a reasonable set to get started with.

Check the preview. I'm initially looking for feedback on the basic approach, using Cards with brief guidance within. This seemed easier to scan to me but perhaps there are other ideas to consider.

**General UX Checks**
- [x] Light & dark modes

**Accessibility Checks**
- [x] Keyboard interactions

**Code Quality Checks**
NOTE: There are some warnings about 'Meta' that seem unrelated to these changes.

NOTE: There is a bizarre eslint error that is evading triage.

Check the preview link.

#2167 

Screen Shot 2022-04-07 at 5 17 08 PM

yes

backwards compatible


taysea:
Overall I think the approach is reasonable, except I did have the thought at first that I might be clicking on the cards.

In terms of design, what if we took the approach of some of the Card variations I've been seeing (icon on top, then title, then description):
Screen Shot 2022-04-07 at 5 49 31 PM

We could also remove the elevation and do a thin border or no border as well to make them look less clickable:
Screen Shot 2022-04-07 at 5 51 02 PM

ericsoderberghp:
> In terms of design, what if we took the approach of some of the Card variations I've been seeing (icon on top, then title, then description):

I've adjusted the Card layout along these lines. Agreed that it's more consistent. I was finding that the icons seemed a little more subdued with this approach. They didn't stand out as much. But, I'm OK with this.

ericsoderberghp:
> We could also remove the elevation and do a thin border or no border as well to make them look less clickable:

Changed to have no elevation. Seems aligned better with larger examples.

aries-site/src/examples/foundation/icons/IconCoreExample.js
```diff
@@ -0,0 +1,105 @@
...
+  FormUp,
+  Notification,
+  Search,
+  Subtract,
+  Trash,
+} from 'grommet-icons';
+
+const CoreIcon = ({ Icon, name, guidance }) => (
+  <Card elevation={false}>
+    <CardHeader direction="column" align="stretch" gap="small">
```
taysea, line 20:
```suggestion
    <CardHeader direction="column" align="stretch" gap="small" pad={{ top: 'medium', horizontal: 'medium', bottom: 'small' }} >
```

Card designs call for 12px space between header and body description.

aries-site/src/examples/foundation/icons/IconCoreExample.js
```diff
@@ -0,0 +1,105 @@
...
+    <CoreIcon
+      Icon={Search}
+      name="Search"
+      guidance="Placed next to and within search fields as an action to
+        access and execute search and filtering."
+    />
+    <CoreIcon
+      Icon={Close}
+      name="Close"
+      guidance="Placed in buttons as an action to
```
taysea, line 89: Based on this description, why did we change the FileInput icon from Trash --> Close?

aries-site/src/examples/foundation/icons/IconCoreExample.js
ericsoderberghp: Does that mean that all uses of Card have to customize padding in this way? I would have expected this to be built into the theme.

aries-site/src/examples/foundation/icons/IconCoreExample.js
ericsoderberghp: The Close in FileInput is closer to "clearing an input field", which most browsers default to an 'x'. Also, there was some confusion as to whether Trash in FileInput meant that the file itself would be deleted, which isn't the case.

aries-site/src/pages/foundation/icons.mdx
```diff
@@ -78,3 +79,9 @@ The default icon color is 'text-strong' for the corresponding light/dark mode. F
 <Example code="https://raw.githubusercontent.com/grommet/hpe-design-system/master/aries-site/src/examples/foundation/icons/IconPlainExample.js">
   <IconPlainExample />
 </Example>
+
+## Core Icons
+
+The following icons are frequently used and should be used consistently to avoid confusing the user by using different icons for the same meaning.
```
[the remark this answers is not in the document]
ericsoderberghp: Updated as suggested

aries-site/src/examples/foundation/icons/IconCoreExample.js
taysea: Currently it's not built into the theme to know how to handle when CardHeader, CardBody, CardFooter are present. Would be nice if we could make some enhancements so the caller doesn't have to worry about it.

halocline: We have padding set in the [HPE theme](https://github.com/grommet/grommet-theme-hpe/blob/31e52ccc1aa211be75e63eed8d73d1f03580a42a/src/js/themes/hpe.js#L387) for body, footer, and header. Am I misunderstanding? Or are you saying the values in the theme don't align with the designs?

taysea: ^^ @halocline As Card patterns have become a little more fleshed out, the pad values that we having in the theme apply on all sides of each of the card elements (body, footer, header) which results often in too much padding on the sides where body + header meet or body + footer meet, etc.

Proposing as Card patterns become more fleshed out that we can be more intelligent (under the hood) of how the padding gets applied to card elements when body + footer, body + header etc. are in play.

halocline: Is there any consistency to the desired spacing? Can we place the responsibility on `card.container.gap` and remove vertical pad from body, footer, header?

taysea: I think that's a possibility and we could just apply horizontal spacing to the individual elements, but a bit more investigation would need to be done to confirm (and our design variants should be ironed out first). I think this would be important follow-on work to the Card exploration I had done. I filed a ticket for it: https://github.com/grommet/hpe-design-system/issues/2521

taysea (approved):
LGTM. Thanks Eric.

halocline (commented):
My mindset when searching for which icon to use begins with the "action." What functionality am I wanting to provide for my end user and thus what I want my icon to represent? For example, I start with the idea of "remove" or "delete" or "zoom out". I am not thinking "subtract" to get me to remove or delete.

This presentation lends itself to presenting one icon to many potential uses. Where my tendency is to want more of a many-to-one presentation which leads with the use and then maps to the supporting icon.

Proposed future enhancements to this experience:
- Adding a search filter
- Explore a visual layout which leads with the desired user action and maps to the associated icon.
  - I would think about this as an alphabetized list of actions which I can easily scan and find my action at its expected index and/or just use a search bar.

ericsoderberghp:
I've changed the structure of the core icons content in an attempt to address the feedback regarding typical use and aliases.

aries-site/src/examples/foundation/icons/IconCoreExample.js
```diff
@@ -0,0 +1,152 @@
...
+      <Box>
+        <Heading level="3" size="small" margin="none">
+          {name}
+        </Heading>
+        {aliases && (
+          <Box direction="row" gap="xsmall" wrap>
+            <Text>or</Text>
+            {aliases.map((a, i) => (
+              <>
+                <Text weight="bold">{a}</Text>
```
halocline, line 45: Aligning font colors of Heading and aliases. Soft opinion.

```suggestion
                <Text weight="bold" color="text-strong" >{a}</Text>
```

halocline (commented):
This addresses my desired "action-first" mindset.

aries-site/src/examples/foundation/icons/IconCoreExample.js
```diff
@@ -0,0 +1,152 @@
...
+  <Card elevation={false}>
+    <CardHeader
+      direction="column"
+      align="stretch"
+      gap="small"
+      pad={{ top: 'medium', horizontal: 'medium', bottom: 'small' }}
+    >
+      <Icon />
+      <Box>
+        <Heading level="3" size="small" margin="none">
```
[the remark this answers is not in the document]
ericsoderberghp: For a scenario like this, I think the focus of how `CoreIcon` is used is narrow enough to not need to be generalized further. Agreed on desiring lint/test tools to verify correctness.

For a more generalized scenario, I could see a few options:
- require the caller to pass `level` as a prop explicitly - still vulnerable to copy/paste
- add `level` as a prop, default to 2, but allow caller's to override - perhaps slightly friendlier but still problematic for copy/paste
- do something interesting with PageContext to capture both current Heading level as well as crossing boundaries between things like Card, `<Box as="section" />`, etc. and then flag debug warnings if the structure doesn't look right
- change Heading such that when `level` isn't specified, it defaults to the next level down, using something like a new HeadingContext, would have to be opt-in for compatibility
- slight modification to above but allow something like `level="peer"`, `level="down"`, and `level="up"`, still a bit cumbersome as arrays would need to check for the first item being "down" but all others being "peer"
- ???

halocline (approved):

aries-site/src/examples/foundation/icons/IconCoreExample.js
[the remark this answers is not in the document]
halocline: @taysea ☝️ FYI, some thoughts for consideration...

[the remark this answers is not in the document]
taysea: Nice, these are great ideas. Helpful things to ponder on as we are moving forward with Card.

taysea (approved):

PR #2527, opened 2022-04-13
Explore PageHeader implementation

### LINKED ISSUE #2517: Page Header - explore how to implement in code

taysea:
[Deploy Preview](https://deploy-preview-2527--keen-mayer-a86c8b.netlify.app/templates/dashboards)

#### What does this PR do?

This PR is a first pass exploration of how PageHeader might be implemented. On first pass, I feel the layout is potentially too opinionated to add as a Grommet component and it might be better to just provide a starter template from the Design System.

The PageHeader will:
1. Control the layout, placement, and spacing of the contextual navigation, title, subtitle, and actions
2. Determine the responsive layout of page-level actions on larger screens (row of buttons) to smaller screens (collapsed into Menu)
3. Determine the sizing of the page title and subtitle

```
// Sample Implementation

<PageHeader
   title="Manage Devices"
   subtitle={`Onboard and manage all devices in your inventory. 
      This could also be slightly longer.`}
   actions={[
      { label: 'Other action', secondary: true },
      { label: 'Add Device', primary: true },
   ]}
   contextualNav={
      <Link href="#" passHref>
         <Button label="Settings" icon={<FormPrevious />} kind="up" />
      </Link>
   }
/>
```

```
// API
actions: arrayOf(ButtonProps)
contextualNav: JSX.Element
title: string,
subtitle: string | JSX.Element
```

**actions - arrayOf(ButtonProps)**

Actions are page-level actions that appear at the far-right edge of the PageHeader and are aligned vertically with the title. It takes an array of Button prop objects. 

PROS:
PageHeader controls the responsive behavior:
- By default, actions are displayed in a row
- On small/xsmall screens, buttons are condensed into a Menu

CONS:
- This approach is that this wouldn't allow "non-Button" elements in the page-level actions, such as a toggle. Do we expect to support these kinds of elements?

**contextualNav - JSX Element** (Up Button or BreadCrumbs)

**Up Button**

1. Should be a JSX Element to make it easiest for caller to integrate their router, without us having to make too many assumptions about how a given router works.
2. Should add `kind="up` to the theme to control the styling.
   - It would be nice to also be able to specify an icon for a given Button kind within the theme.

PROS:
- Easy to integrate router
- Leverage theme for kind styling

CONS:
- People could forget to use `kind="up"`

**Breadcrumbs**
While the "Up Button" is more common, I have also explored a BreadCrumb component (I think this component could potentially be a valuable Grommet component IF we are going to support the idea of breadcrumbs in design system). In the code, I've explored 2 approaches:
- Approach 1: BreadCrumbs that takes `items` (array of anchor props) and `router` (a function that returns the router wrapper)
- Approach 2: BreadCrumbs acts as a wrapper that will place a separator between each of its children (this is more similar to how Material UI works and more similar to my proposal for "Up Button" which makes it easy for caller to wrap their Anchor in their router).

aries-site/src/examples/templates/PageHeader/PageHeader.js

**General UX Checks**
- [ ] Light & dark modes

**Accessibility Checks**
- [ ] Keyboard interactions

**Code Quality Checks**

Closes #2517


halocline:
> **actions - arrayOf(ButtonProps)**
> 
> Actions are page-level actions that appear at the far-right edge of the PageHeader and are aligned vertically with the title. It takes an array of Button prop objects.
> 
> PROS: PageHeader controls the responsive behavior:
> 
> * By default, actions are displayed in a row
> * On small/xsmall screens, buttons are condensed into a Menu
> 
> CONS:
> 
> * This approach is that this wouldn't allow "non-Button" elements in the page-level actions, such as a toggle. Do we expect to support these kinds of elements?
> 

I can see a Menu as a common item folks might need/want in addition to buttons. I am not sure about a toggle.

It would be good to have a list of potential page actions to weigh this against. I didn't see a list of action types provided in the use cases. Here's a stab:
- apply/save
- new/create, edit/update, delete
- upload/download
- share
- favorite/bookmark
- subscribe
- start trial
- enable (I suppose this could be a candidate for toggle)


Regardless, from an interface it seems like we should lean towards flexibility. Perhaps `PropTypes.arrayOf(PropTypes.oneOfType([...])`?

taysea:
> I can see a Menu as a common item folks might need/want in addition to buttons. I am not sure about a toggle.
> 
> It would be good to have a list of potential page actions to weigh this against. I didn't see a list of action types provided in the use cases. Here's a stab:
> 
> * apply/save
> * new/create, edit/update, delete
> * upload/download
> * share
> * favorite/bookmark
> * subscribe
> * start trial
> * enable (I suppose this could be a candidate for toggle)
>

Thanks for adding potential list of actions. Agree it would be helpful to have that brainstormed from the design side. I thought I remembered seeing a toggle in some GreenLake design but I'd have to dig it up.

> Regardless, from an interface it seems like we should lean towards flexibility. Perhaps `PropTypes.arrayOf(PropTypes.oneOfType([...])`?

Agree on flexibility -- in that case we let the implementer handle the responsive behavior and just give guidance that it should collapse into a Menu w a certain icon/label (if that's the behavior we decide on?)

aries-site/src/examples/templates/PageHeader/PageHeader.js
```diff
@@ -1,24 +1,67 @@
...
+  actions,
+  contextualNav,
+  ...rest
+}) => {
+  const size = useContext(ResponsiveContext);
+
+  return (
+    <Header direction="column" align="stretch" gap="small" {...rest}>
+      <Box align="start">{contextualNav}</Box>
+      <Box align="start" direction="row" gap="small" justify="between">
```
halocline, line 26: Do we have any designs/guidance for what happens when the Heading is longer and collides with the actions?

Are there max-widths to be defined at various breakpoints?

aries-site/src/examples/templates/PageHeader/PageHeader.js
```diff
@@ -1,24 +1,67 @@
...
+            <Paragraph margin="none" size="large">
+              {subtitle}
+            </Paragraph>
+          ) : (
+            <>{subtitle}</>
+          )}
+        </Box>
+        <Box direction="row" gap="small">
+          {!['xsmall', 'small'].includes(size) ? (
+            actions.map(action => (
```
halocline, line 41: What happens if I pass in an array of 5+ action items? Is there a limit? Should those in excess of the limit be put under an "overflow menu"?

aries-site/src/examples/templates/PageHeader/PageHeader.js
taysea: Great question -- I posed this question to Vicky as something she should explore from the design side. I do think there should be a min-width for the text content (taking precedence over the page-level actions).

aries-site/src/examples/templates/PageHeader/PageHeader.js
taysea: Another great design question which I posed to Vicky yesterday for her to consider from the design side. Yes, there should be a limit of how many are showed flat on the screen.

From the dev side, I was exploring an approach locally (but didn't push up) where the PageHeader is implemented as a Grid. The page-level actions are "smart" about their gridArea dimensions and when they are wider than their allowed space, they collapse into a menu.

vavalos5: Yes, indeed great question @halocline. I haven't gotten to this section of the PageHeader. However, I imagine we'd give guidance that IF there are more than 4 action items or when on a smaller resolution such as mobile, the action items will then live in a menu item or and "overflow menu" (this is still tbd). But again, have not gotten to this.

taysea: Just planning some seeds for when you get there, doesn't have to be answered now. If there are (for example) 4 action items, do:
1. All the action items collapse into a menu?
2. 3 of the action items collapse into but the primary action stays visible next to the Menu.
3. Some other thing?

halocline:
> **contextualNav - JSX Element** (Up Button or BreadCrumbs)
> 

Does this mean that an Up Button and Breadcrumbs cannot exist at the same time? They are mutually exclusive? I guess that would make sense as they are both ascending in the information hierarchy.

> **Up Button**
> 
> 1. Should be a JSX Element to make it easiest for caller to integrate their router, without us having to make too many assumptions about how a given router works.

Is there an option to consider just providing an `onClick` and let the caller provide a function with their routing behavior?

ericsoderberghp:
A few API thoughts ...
- I like the simple API surface of `<PageHeader title="..." actions={<Button />} />`. Always good to ensure that the simplest case works cleanly.
- I wonder if having `actions` should just allow passing any element and not attempt to embed anything within PageHeader to handle things when there are combinations of things, such as toggles, etc. So, the caller can pass a Menu, a CheckBox toggle, a Button, or a Box with a couple of things.
- I wonder if we should just call it `context` and let the caller put whatever they want in there would be sufficient, allowing them to tie into whatever routing they've got via components, rather than event handlers.

taysea:
> > **contextualNav - JSX Element** (Up Button or BreadCrumbs)
> 
> Does this mean that an Up Button and Breadcrumbs cannot exist at the same time? They are mutually exclusive? I guess that would make sense as they are both ascending in the information hierarchy.
> 

Yes, my thought would be it's one or the other.

> > **Up Button**
> > 
> > 1. Should be a JSX Element to make it easiest for caller to integrate their router, without us having to make too many assumptions about how a given router works.
> 
> Is there an option to consider just providing an `onClick` and let the caller provide a function with their routing behavior?

Definitely -- that aligns with one of the breadcrumbs approaches I have tried.

taysea:
> * I wonder if having `actions` should just allow passing any element and not attempt to embed anything within PageHeader to handle things when there are combinations of things, such as toggles, etc. So, the caller can pass a Menu, a CheckBox toggle, a Button, or a Box with a couple of things.

That would be fine with me. Then we would just want to provide examples of how we want the responsive behavior to happen.

> * I wonder if we should just call it `context` and let the caller put whatever they want in there would be sufficient, allowing them to tie into whatever routing they've got via components, rather than event handlers.

You are referencing `contextualNav` --> `context` ? That could work too, I'd like to queue that up as a discussion point on the shareout.

taysea:
Closing in favor of grommet PR: https://github.com/grommet/grommet/pull/6071

PR #2540, opened 2022-04-19
Add eslint and prettier config locally, add eslint-plugin-grommet

taysea:

#### What does this PR do?

Because we are not actively maintaining (or have easy access to maintain) HPE project scripts, Eric and I have discussed doing the eslint and prettier configs locally. Allowing us to also add in the eslint-plugin-grommet, which I have also added here and made the necessary lint fixes.

This will be beneficial to allow us to use our examples as a testbed for the grommet lint plugin. One note that already came up was with the DataTable ariadescribedBy rule is a little too strict (requiring the header to be in the same file). This needs to be rethought because this won't always (or often) be the case.

.eslintrc.js

Linter ran pre-commit and issues were resolved.

**General UX Checks**
- [ ] Light & dark modes

**Accessibility Checks**
- [ ] Keyboard interactions

**Code Quality Checks**


aries-site/src/examples/components/spinner/ContentSpinnerExample.js
```diff
@@ -86,7 +86,13 @@ export const ContentSpinnerExample = () => {
       >
         <Tab title="Customer Details">
           <Box align="center" justify="center" height={{ min: 'small' }}>
-            <Spinner size="medium" />
+            <Spinner
+              size="medium"
+              message={{
```
britt6612, line 91: good catch we didn't have `message` when we say in our guidance to provide this

halocline (approved):

PR #2683, opened 2022-06-22
change deeplink from anchor to button and copy to clipboard

JCoder121:
[Deploy Preview](https://deploy-preview-2683--keen-mayer-a86c8b.netlify.app/components)

#### What does this PR do?
Allows for a user to click the deeplink subsection buttons in the design system's pages and copies the URL + anchor tag to the clipboard. 
Also provides a "copy to clipboard" hint upon hovering over the link button. 
Fix bug in SubsectionHeader.js where deeplink copy button icon would not appear when user was navigating the site with keyboard only.

aries-site/src/layouts/content/Subsection.js
aries-site/src/layouts/content/SubsectionHeader.js

Visual/manual testing on local machine. 

**General UX Checks**
- [x] Light & dark modes

**Accessibility Checks**
- [x] Keyboard interactions

**Code Quality Checks**

Check that all deeplink button clicks successfully copy the URL to clipboard, test if navigating the site by keyboard only is also able to copy the URL. Test that hovering over all deeplink buttons cause a tip to pop up for that button. Note: when running WAVE accessibility on Chrome locally I received errors unrelated to this PR issue.

Previously, the deeplink button only redirected the user to the specific anchor tag without copying that link to clipboard.

### DIFF
--- a/aries-site/src/layouts/content/Subsection.js
+++ b/aries-site/src/layouts/content/Subsection.js
@@ -99,14 +99,15 @@ export const Subsection = ({
             </Box>
             {level > 1 && (
               <Button
+                tip='Copy link to clipboard'
                 a11yTitle={`Jump to section titled ${name} 
                             and copy link to clipboard`}
                 icon={<LinkIcon color={over ? 'text-xweak' : 'transparent'} />}
                 onClick={() => {
                   window.location.href = `#${id}`;
                   navigator.clipboard.writeText(window.location.href);
-                }}
-              />
+              }}
+            />
             )}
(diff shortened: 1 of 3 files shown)


taysea:
> Just curious, would it be better functionality if after the user presses the deeplink button to copy link, should that button's tip change text to say something like "Copied!" instead of continuing to show "Copy to cliboard"?

I think that's a great idea, maybe once it's clicked we change the text to "Copied" and then the entire tip could disappear after some reasonable delay like 1.5-2sec?

PR #2729, opened 2022-07-13
column selector move next to search

### LINKED ISSUE #2717: Data collection toolbar - Update filtering / datatable examples so that the "column selector" control is placed alongside the search and filter as opposed to inline with the "action" buttons
There may be an issue for this already, but couldn't find.

This is related to having a data collection toolbar composed with:
- View/presentation controls (presentation only, does not manipulate underlying data)
- Action controls (change data and/or apply actions against)
- Results Summary
- Results section

britt6612:
[Deploy Preview](https://deploy-preview-2729--keen-mayer-a86c8b.netlify.app/templates/filtering#filtering-with-selectable-results)

#### What does this PR do?
This PR  Updates the filtering / datatable examples so that the "column selector" control is placed alongside the search and filter as opposed to inline with the "action" buttons
filterControls.js
site

**General UX Checks**
- [x] Light & dark modes

**Accessibility Checks**
- [ ] Keyboard interactions

**Code Quality Checks**

site
This is related to having a data collection toolbar composed with:

View/presentation controls (presentation only, does not manipulate underlying data)
Action controls (change data and/or apply actions against)
Results Summary
Results section

closes #2717 
yes
backwards compatible

### DIFF
--- a/aries-site/src/examples/templates/FilterControls/FilterControls.js
+++ b/aries-site/src/examples/templates/FilterControls/FilterControls.js
@@ -1,11 +1,13 @@
 import { useContext, useEffect } from 'react';
 import PropTypes from 'prop-types';
-import { Box, ResponsiveContext } from 'grommet';
+import { Box, Button, ResponsiveContext } from 'grommet';
+import { Splits } from 'grommet-icons';
 
 import { Filters, ResultsSummary, SearchFilter, useFilters } from '.';
 
 export const FilterControls = ({
   actions,
+  configure,
   data,
   filters,
   layerProps,
@@ -62,6 +64,7 @@ export const FilterControls = ({
             <SearchFilter placeholder={searchFilter.placeholder} />
           )}
           <Filters />
+          {configure}
         </Box>
         {actions}
       </Box>
@@ -72,6 +75,7 @@ export const FilterControls = ({
 
 FilterControls.propTypes = {
   actions: PropTypes.element,
+  configure: PropTypes.element,
   data: PropTypes.array,
   filters: PropTypes.arrayOf(
     PropTypes.shape({
--- a/aries-site/src/examples/templates/FilterControls/FilterControls.js
+++ b/aries-site/src/examples/templates/FilterControls/FilterControls.js
@@ -1,7 +1,6 @@
 import { useContext, useEffect } from 'react';
 import PropTypes from 'prop-types';
-import { Box, Button, ResponsiveContext } from 'grommet';
-import { Splits } from 'grommet-icons';
+import { Box, ResponsiveContext } from 'grommet';
 
 import { Filters, ResultsSummary, SearchFilter, useFilters } from '.';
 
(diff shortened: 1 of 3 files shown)


taysea (approved):
This looks reasonable to me!

aries-site/src/examples/templates/FilterControls/FilterControls.js
```diff
@@ -62,6 +64,7 @@ export const FilterControls = ({
             <SearchFilter placeholder={searchFilter.placeholder} />
           )}
           <Filters />
+          {configure}
```
halocline, line 67: What do you all think about the control order here? Ian called out how the buttons get spaced when filters are applied: 
Screen Shot 2022-07-15 at 10 36 17 AM

I am conflicted, as search & filter are more closely related in function, but the "clear filters" button certainly throws off the visual...

@britt6612 @taysea @bahriaditi @MeganBeierle

halocline (commented):
Looks good. I'd like to hear a little bit of discussion on the button ordering.

aries-site/src/examples/templates/FilterControls/FilterControls.js
britt6612: agree I wonder if we can move the `clear filters` to the last option. Visually this doesn't flow

ericsoderberghp: I think the 80% case is not to have the column order/select control. So that's a relief.
For the case where it is there, I tend to think no options are ideal, as filtering is likely to be used way more than column selection and is more associated with searching, so it makes sense for both proximity and frequency to have it next to search. I do think "Clear Filters" should be next to the filter control. If I had to choose, I'd probably go with what we have in the screenshot above instead of swapping the filter and column buttons. But, I that is a weak preference.

taysea: Oh good catch @halocline -- I think I've seen the "Configure" button appearing between the search + filter in some incoming contributions. I like that approach because then we have a smoother visual flow.

ericsoderberghp: I'm OK with configure before filter

MeganBeierle: Good call @bahriaditi 

The example you provided with column sorting on the right looks so much better. Do we know how often customers choose to configure columns? Just curious.

britt6612: > We should probably at some point reassess placement of 'configure' - whether it belongs on the right or the left. I see that there is an [example on our site](https://design-system.hpe.design/templates/filtering?q=filt#filtering-with-selectable-results) which has it on the right.

@bahriaditi that is what this PR is doing is changing how we currently have it. which is not having the selector by the action buttons

taysea (approved):

aries-site/src/examples/templates/FilterControls/FilterControls.js
bahriaditi: Well, this is embarrassing 🤷🏼‍♀️ - Friday brain

halocline (approved):

PR #2134, opened 2021-12-10
POC for In-page Navigation

taysea:
[Deploy Preview](https://deploy-preview-2134--keen-mayer-a86c8b.netlify.app/components/anchor)

#### What does this PR do?

Exploring how we might leverage the search content work to dynamically create in-page navigation. A common feedback from teams is "It is hard to know what is on a page because it is so long". This creates a sticky side nav that allows people to scan page content via headings and jump to them directly.

All feedback welcomed.

#### What testing has been done on this PR?

In addition to the feature you are implementing, have you checked the following:

**General UX Checks**
- [ ] Small, medium, and large screen sizes
- [ ] Cross-browsers (FireFox, Chrome, and Safari)
- [ ] Light & dark modes
- [ ] All hyperlinks route properly

**Accessibility Checks**
- [ ] Keyboard interactions
- [ ] Screen reader experience
- [ ] Run WAVE accessibility plugin (Chrome)

**Code Quality Checks**
- [ ] Console is free of warnings and errors
- [ ] Passes E2E commit checks
- [ ] Visual snapshots are reasonable

PR #2753, opened 2022-07-25
2733 Design Site In-Page Table of Contents

### LINKED ISSUE #2734: Add an in-page table of contents to increase visibility of page content
To improve the ability to discover page content and navigate to it, an in-page "jump to section" navigation should be implemented.

Design file: https://www.figma.com/file/qbbYGi73ncFAoLQ0IM1Y7x/DS-Guide-Template?node-id=3875%3A20038

- [x] The "jump to section" should remain sticky on scroll so it's always in view
- [x] The current section should be marked as active and update as the user scrolls
- [x] The headings should be parsed from the markdown and pulled in dynamically to ease maintenance as more guidance is written

JCoder121:
[Deploy Preview](https://deploy-preview-2753--keen-mayer-a86c8b.netlify.app/components/card)

#### What does this PR do?
Adds a table of contents to component pages that allows the user to quickly navigate to all sections headers on the page. The ToC is "sticky" and will not disappear from view when the user scrolls through the page. It also highlights the current section the user is on for added visibility. Uses regex to capture and parse all markdown pages. On certain pages with no headings (ex: Box), the table of contents is not shown. Final design to indicate active section header in the ToC TBD.

aries-site/src/layouts/content/InPageNavigation.js
aries-site/src/scripts/site-contents.mjs

Manual testing on all component pages to ensure all headings are correctly found by regex and displayed on table of contents.

**General UX Checks**
- [x] Light & dark modes

**Accessibility Checks**
- [x] Keyboard interactions

**Code Quality Checks**

Click heading links on table of contents, ensure that all headings show up on each component page.

Existing table of contents branch for reference at #2134, this branch builds off of it and implements further details that conform to DS template guide. Currently many component pages have a lot of information that is not easily accessible and requires a lot of scrolling in a single page, and the table of contents is the first step towards making standardized and accessible documentation. There was a formatting issue with some of the markdown pages (with CardGrid component) that caused the content to be crushed into each other, but this had to do with how src/components/cards/CardGrid.js was filling the container and has been fixed.

Closes #2734.

Most updated designs:

image

image

image

Yes - Add table of contents to the Design System content pages.

Backwards compatible.


aries-site/src/layouts/content/InPageNavigation.js
```diff
@@ -0,0 +1,106 @@
...
+        position: 'sticky',
+        marginTop: '50px',
+        flexShrink: 0,
+        top: '30px',
+      }}
+      width="small"
+      flex={false}
+    >
+      <Text color="text-strong" size="xsmall" weight="bold">
+        CONTENTS
```
taysea, line 64:
```suggestion
        Jump to section
```
Double check with the design for sizing and spacing with other surrounding content.

aries-site/src/layouts/main/Layout.js
```diff
@@ -53,58 +58,79 @@ export const Layout = ({
...
+                </PageContent>
+                <MainContentWrapper>
+                  <Main overflow="visible">
+                    <SkipLinkTarget id="main" />
+                    <Box direction={layout !== 'plain' ? 'row' : 'column'}>
+                      {layout !== 'plain' ? (
+                        <>
+                          <Box
+                            width={
+                              size !== 'small'
```
taysea, line 92:
```suggestion
                              !size.includes['xsmall', 'small']
```

aries-site/src/layouts/content/InPageNavigation.js
```diff
@@ -0,0 +1,106 @@
...
+      </Text>
+
+      <Nav gap="none">
+        {headings.map((heading, index) => {
+          const levelRegexp = new RegExp(/^(#)+/);
+          const [level] = heading[0].match(levelRegexp);
+          const headingIsActive = activeId === nameToSlug(heading[1]);
```
taysea, line 71:
```suggestion
          const active = activeId === nameToSlug(heading[1]);
```
small suggestion to simplify variable name -- will need to adjust below also

taysea, line 71: also suggesting we abstract "heading[1]" to its own variable since we use it in multiple places.

maybe heading = heading[1] with a comment of why we need the [1]

aries-site/src/layouts/content/InPageNavigation.js
```diff
@@ -0,0 +1,106 @@
...
+                alignSelf="start"
+                style={{ textAlign: 'start', borderRadius: '3px' }}
+                size="small"
+                kind="option"
+                active={headingIsActive}
+                label={
+                  <Text
+                    color="text-strong"
+                    size="small"
+                    weight={headingIsActive ? 'bold' : 'normal'}
```
taysea, line 89: Wondering if we shouldn't toggle the font weight and just use the default font weight always. When I scroll, sometimes the boldness causes the label to wrap which is a bit of a jumpy experience:

https://user-images.githubusercontent.com/12522275/180891542-35455109-cea1-4de4-8ef6-6660fca5be5c.mov

aries-site/src/layouts/main/Layout.js
```diff
@@ -53,58 +58,79 @@ export const Layout = ({
...
+                canonicalUrl={`https://design-system.hpe.design${router.route}`}
+              />
+              <>
+                <SkipLinks id="skip-links">
+                  <SkipLink id="main" label="Main Content" />
+                </SkipLinks>
+                <PageContent>
+                  <Header fill="horizontal" alignSelf="center" />
+                </PageContent>
+                <MainContentWrapper>
```
taysea, line 84: This page is rendering a bit weird -- not sure exactly what's driving it but putting the comment here: https://deploy-preview-2753--keen-mayer-a86c8b.netlify.app/templates/content-layouts

Screen Shot 2022-07-25 at 4 30 30 PM

taysea: notice how content is being crushed / overlapped.

taysea: This one as well: https://deploy-preview-2753--keen-mayer-a86c8b.netlify.app/templates/navigation

aries-site/src/layouts/content/InPageNavigation.js
```diff
@@ -0,0 +1,115 @@
...
+import Link from 'next/link';
+import PropTypes from 'prop-types';
+import { Box, Button, Nav, Text } from 'grommet';
+import { siteContents } from '../../data/search/contentForSearch';
+import { nameToSlug } from '../../utils';
+
+const useActiveHeadingId = (headings, options) => {
+  const [activeHeadingId, setActiveHeadingId] = useState();
+  const observer = useRef();
+  useEffect(() => {
```
taysea, line 11: Can we add a comment for future about what this `useEffect` does?

aries-site/src/layouts/content/InPageNavigation.js
```diff
@@ -0,0 +1,115 @@
...
+  const topMargin = '50';
+  const bottomMargin = '-95';
+  const activeId = useActiveHeadingId(headings, {
+    rootMargin: `${topMargin}% 0% ${bottomMargin}% 0%`,
+  });
+
+  return headings.length > 0 ? (
+    <Box
+      gap="xsmall"
+      pad={{ left: 'xxsmall' }} // account for focus indicator space
```
taysea, line 51: can we confirm if this is needed or not?

aries-site/src/layouts/content/InPageNavigation.js
```diff
@@ -0,0 +1,115 @@
...
+        flexShrink: 0,
+        top: '30px',
+      }}
+      width="small"
+      flex={false}
+    >
+      <Text
+        color="text-strong"
+        weight="bold"
+        margin={{
```
taysea, line 66: Generally we avoid `margin` and apply spacing using `gap` on the container Box -- can we try that?

aries-site/src/layouts/content/InPageNavigation.js
```diff
@@ -0,0 +1,115 @@
...
+          vertical: 'xsmall',
+        }}
+      >
+        Jump to section
+      </Text>
+
+      <Nav gap="none">
```
taysea, line 73: We may need to add an aria-labelledby here and an `id` on the "Jump to section" text so the screenreader can have some info on what kind of navigation this is.

taysea: Or maybe just adding `a11yTitle="Jump to section"`

aries-site/src/layouts/main/Layout.js
```diff
@@ -53,58 +58,85 @@ export const Layout = ({
...
+                  <Main overflow="visible">
+                    <SkipLinkTarget id="main" />
+                    <Box direction={layout !== 'plain' ? 'row' : 'column'}>
+                      {layout !== 'plain' ? (
+                        <>
+                          <Box
+                            width={
```
taysea, line 91: Maybe we could abstract all of this to something like:

```
const contentWrapperStyles = !['xsmall', 'small'].includes(size) ? { width: ..., pad: ..., margin: ... } : undefined;
```

then spread the `contentWrapperStyles` here

taysea: also wondering if we might be able to just do an xlarge pad as opposed to using margin too

aries-site/src/layouts/main/Layout.js
```diff
@@ -53,58 +58,85 @@ export const Layout = ({
...
+                            {relatedContent.length > 0 && (
+                              <RelatedContent
+                                relatedContent={relatedContent}
+                                title={title}
+                              />
+                            )}
+                            <FeedbackSection />
+                          </Box>
+                          <Box pad="small">
+                            {!['xsmall', 'small'].includes(size) ? (
```
taysea, line 124: We should move the logic outside of the Box on line 123 so that we don't render an empty Box to the DOM.

aries-site/src/layouts/main/Layout.js
JCoder121: I removed the ResponsiveContext.Consumer and instead declared a size variable above, so that I could implement it this way.
image
image

jcfilben:
The tab order of the page is a little confusing. When I tab through the page the jump to section links are at the very end of the page, I was expecting they would be at the beginning of the page. Also when using a screenreader it isn't super clear that these links are in page navigation.

jcfilben:
Sometimes when I scroll quickly to the top of the page the table of contents doesn't accurately reflect the section I am on. (Tested on the [Card page](https://deploy-preview-2753--keen-mayer-a86c8b.netlify.app/components/card#what-makes-up-a-card))

JCoder121:
> Sometimes when I scroll quickly to the top of the page the table of contents doesn't accurately reflect the section I am on. (Tested on the [Card page](https://deploy-preview-2753--keen-mayer-a86c8b.netlify.app/components/card#what-makes-up-a-card))

I think this issue along with the issue of scrolling through the page quickly leading to incorrect sections being highlighted has to do with how the logic is determining which section is active, I'll look into it.

JCoder121:
> > Sometimes when I scroll quickly to the top of the page the table of contents doesn't accurately reflect the section I am on. (Tested on the [Card page](https://deploy-preview-2753--keen-mayer-a86c8b.netlify.app/components/card#what-makes-up-a-card))
> 
> I think this issue along with the issue of incorrect sections being highlighted has to do with how the logic is determining which section is active, I'll look into it.

I discussed this with Taylor and for now I'll add logic to show no active heading if a user scrolls all the way to the top really fast,

aries-site/src/layouts/content/InPageNavigation.js
```diff
@@ -0,0 +1,126 @@
...
+        overflowY: 'auto',
+        position: 'sticky',
+        marginTop: '50px',
+        flexShrink: 0,
+        top: '30px',
+      }}
+      width="small"
+      flex={false}
+    >
+      <Box pad="small" a11yTitle="Table of Contents Heading">
```
taysea, line 71: This a11yTitle shouldn't be on this Box. The screenreader will naturally read the text. I think what we want is an `a11yTitle` on the `nav` that reads "Jump to section"

taysea: If we do that, then the screenreader reads the following: 
Screen Shot 2022-08-02 at 11 00 39 AM

JCoder121: I also added a Jump to section {headingTitle} in the boxes below. It's probably a little redundant, but this way the screen reader actually indicates that it is a jump to section link. Is this ok?

aries-site/src/layouts/content/InPageNavigation.js
```diff
@@ -0,0 +1,126 @@
...
+  });
+
+  return headings.length > 0 ? (
+    <Box
+      pad={{ horizontal: 'xsmall' }} // pad for keyboard focus
+      style={{
+        height: 'calc(100vh - 50px)',
+        overflowY: 'auto',
+        position: 'sticky',
+        marginTop: '50px',
```
taysea, line 64: I think that "Jump to section" should be vertically aligned with page title to start.

Screen Shot 2022-08-02 at 11 02 35 AM

aries-site/src/layouts/content/InPageNavigation.js
```diff
@@ -0,0 +1,126 @@
...
+        flexShrink: 0,
+        top: '30px',
+      }}
+      width="small"
+      flex={false}
+    >
+      <Box pad="small" a11yTitle="Table of Contents Heading">
+        <Text color="text-strong" weight="bold">
+          Jump to section
+        </Text>
```
taysea, line 74: Noticed that on pages with lots of headings, "Jump to section" is getting squished. Wondering if we need a `flex={false}` or something on the Box surrounding this Text.
Screen Shot 2022-08-02 at 10 11 00 AM

aries-site/src/layouts/content/InPageNavigation.js
```diff
@@ -0,0 +1,126 @@
...
+            >
+              <Link key={index} href={`#${nameToSlug(headingTitle)}`} passHref>
+                <Button
+                  style={{ textAlign: 'start' }}
+                  size="small"
+                  kind="option"
+                  label={
+                    <Box
+                      pad={{ left: subsectionPad }}
+                      margin={active ? undefined : { left: '2px' }}
```
taysea, line 107: Nice! Small comment -- instead of hard-coding the pixels, can you grab directly from the theme for borderSize.medium (or edgeSize? i forget exactly which theme object it is).

We'll need to do something above like:

```
import { ThemeContext } from 'styled-components';
...
...
```

aries-site/src/layouts/content/InPageNavigation.js
taysea: I don't think we need to add to each button necessarily if we have it once on the overall nav. Here's the MDN best practices and example: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/navigation_role#associated_wai-aria_roles_states_and_properties

aries-site/src/layouts/main/Layout.js
```diff
@@ -77,24 +83,38 @@ export const Layout = ({
...
-                      <FeedbackSection />
+                      {!['xsmall', 'small'].includes(size) ? (
+                        <Box pad={{ left: 'large' }}>
+                          <InPageNavigation title={title} />
+                        </Box>
+                      ) : undefined}
+                      <Box
+                        width={
+                          !['xsmall', 'small'].includes(size)
+                            ? 'calc(100% - 192px)'
```
taysea, line 98: Can we add a small comment here that `192px` is equivalent to `small` t-shirt size?

aries-site/src/layouts/content/InPageNavigation.js
```diff
@@ -61,20 +63,20 @@ export const InPageNavigation = ({ title }) => {
         height: 'calc(100vh - 50px)',
         overflowY: 'auto',
         position: 'sticky',
-        marginTop: '50px',
```
taysea: I noticed we change this value, should we also change the "height" calculation a few lines above? is that `100vh - 50px` related to this 50px?

aries-site/src/layouts/content/InPageNavigation.js
```diff
@@ -61,20 +63,20 @@ export const InPageNavigation = ({ title }) => {
...
     >
-      <Box pad="small" a11yTitle="Table of Contents Heading">
+      <Box pad="small" flex={false}>
         <Text color="text-strong" weight="bold">
           Jump to section
         </Text>
       </Box>
 
-      <Nav gap="none">
+      <Nav gap="none" a11yTitle="Table of Contents Heading">
```
taysea, line 79: Recommending we just call this "Table of contents"

taysea: or "Jump to section" to align with what's visually shown on the page

aries-site/src/layouts/content/InPageNavigation.js
```diff
@@ -104,7 +103,12 @@ export const InPageNavigation = ({ title }) => {
                   label={
                     <Box
                       pad={{ left: subsectionPad }}
-                      margin={active ? undefined : { left: '2px' }}
+                      // margin={active ? undefined : { left: '2px' }}
```
taysea: Remove this commented out line now

aries-site/src/layouts/content/InPageNavigation.js
JCoder121: Not related, the `height` calculation refers to the size of the box that contains the ToC, and I was playing around with it but I liked it where it was at before. We can discuss further on this.

aries-site/src/layouts/content/InPageNavigation.js
```diff
@@ -0,0 +1,132 @@
...
+        position: 'sticky',
+        marginTop: '123px', // align "Jump to section" with page title at start
+        flexShrink: 0,
+        top: '30px',
+      }}
+      width="small"
+      flex={false}
+    >
+      <Box
+        pad="small"
```
taysea, line 74:
```suggestion
        pad={{ horizontal: 'small', bottom: 'small' }}
```

aries-site/src/layouts/content/InPageNavigation.js
```diff
@@ -0,0 +1,132 @@
...
+          };
+
+          let subsectionPad = 'xsmall';
+          if (level.length > 3) subsectionPad = 'large';
+          else if (level.length === 3) subsectionPad = 'medium';
+          return (
+            <Box border={borderLeft}>
+              <Link key={index} href={`#${nameToSlug(headingTitle)}`} passHref>
+                <Button
+                  style={{ textAlign: 'start' }}
```
taysea, line 104: We try to avoid doing inline `style` wherever possible and relying purely on Grommet props. Is this having an effect on the layout or can we remove?

If it is having an effect on layout, can we apply `textAlign="start"` on the Box at 107?

aries-site/src/layouts/content/InPageNavigation.js
```diff
@@ -0,0 +1,132 @@
...
+                  label={
+                    <Box
+                      pad={{ left: subsectionPad }}
+                      margin={
+                        active
+                          ? undefined
+                          : { left: theme.global.borderSize.small }
+                      }
+                    >
+                      <Text color="text-strong" size="small" weight="normal">
```
taysea, line 115: Do we need `weight="normal"` or is that the default?

aries-site/src/layouts/content/InPageNavigation.js
JCoder121: I just used the textAlign prop within the Text component.

aries-site/src/layouts/content/InPageNavigation.js
JCoder121: `weight = "normal"` is needed, since otherwise the text becomes bolded similar to that of the "Jump to section" title

taysea (approved):
Nice work!

jcfilben (commented):
Coming along nicely!
One thing I noticed on the [all components page](https://deploy-preview-2753--keen-mayer-a86c8b.netlify.app/components/all-components#themecontext) is the table of contents acts a bit jumpy when I am near the top of the page (let me know if you want me to include a video of this) but basically the table of contents jumps between 'Layout/Header" or "Layout/Main" while I am at the top on the page and not on either of these sections.

Also the table of contents on the all components page is long enough that not all of it is visible on my screen. I would be nice when I am scrolling down through the page if the table of contents adjusted so the section I am viewing is visible within the table of contents.

It might also be nice if there was a way to skip over the table of contents if you are a keyboard only user instead of having to tab all the way through it. For example maybe you can tab to the table of contents container and then choose to interact with it on choose to tab past it.

JCoder121:
> 

I'll fix the jumpiness on the All Components page, that should be straightforward.
For a dynamically adjusting table of contents, I'll look into making the section visible like you mentioned.
Also, I'll touch base with Taylor on the keyboard interactions, since I've never interacted with a component that had the ability to be "skipped over"

jcfilben (approved):

jcfilben:
Approving, additional work can be handled as separate PRs

PR #2829, opened 2022-08-16
Use HPE's useIntersection hook in Table of Contents

### LINKED ISSUE #2798: Leverage "useIntersection" hook for Jump to Section
Leverage HPE project scripts useIntersection hook in place of one-off implementation for consistency across projects.

https://github.com/HewlettPackard/hpe-js-project-scripts/tree/master/packages/react-hooks#useintersection

JCoder121:
[Deploy Preview](https://deploy-preview-2829--keen-mayer-a86c8b.netlify.app/components/card/)

#### What does this PR do?
Uses HPE's useIntersection hook rather than a one-off hook for Table of Contents for standardization.

aries-site/src/layouts/content/InPageNavigation.js
aries-site/src/layouts/content/SubsectionHeader.js

Visual testing on local build. Active heading is not 100% accurate, still need to define the correct observation window. Additionally, pages like Templates/Content Layouts freeze (cannot scroll), this issue still needs to be fixed.

**General UX Checks**
- [ ] Light & dark modes

**Accessibility Checks**
- [ ] Keyboard interactions

**Code Quality Checks**

Ensure all pages can scroll properly and highlight the correct active section header.

Previously, in InPageNavigation, the active heading for the Table of Contents was found using a one-time hook implementation. This was replaced with HPE's existing useIntersection hook which also used Context.

Closes #2798 

image
In the screenshot above, the highlighted section should be "What Makes Up A Card?" but the observation window needs to be fixed.

Unsure - consult with Taylor.

Backwards compatible.

### DIFF
--- a/aries-site/src/layouts/content/SubsectionHeader.js
+++ b/aries-site/src/layouts/content/SubsectionHeader.js
@@ -1,4 +1,4 @@
-import React, { useState } from 'react';
+import React, { useState, useContext, useEffect } from 'react';
 import PropTypes from 'prop-types';
 import { Box, Button, Heading } from 'grommet';
 import { Link as LinkIcon } from 'grommet-icons';
@@ -6,11 +6,43 @@ import { Link as LinkIcon } from 'grommet-icons';
 import { HighlightPhrase } from '../../components';
 import { nameToSlug } from '../../utils';
 
+import { ThemeContext } from 'styled-components';
+
+import { useIntersection } from '@hpe/react-hooks';
+import { ActiveHeadingContext } from './ActiveHeadingContext';
+
 export const SubsectionHeader = ({ children, level }) => {
   const [over, setOver] = useState(false);
   const id = nameToSlug(children);
+  const { setActiveHeading } = useContext(ActiveHeadingContext);
+
+  const theme = useContext(ThemeContext);
+  let { large, medium } = theme.global.edgeSize;
+  large = parseInt(large.replace('px', ''), 10); // 48
+  medium = parseInt(medium.replace('px', ''), 10); // 24
+
+  // top and bottom margin values to calculate intersection window
+  const topMargin = large;
+  const bottomMargin = `-${2 * large}`;
+
+  const options = {
+    rootMargin: `${topMargin}% 0% ${bottomMargin}% 0%`,
+    //threshold: 1.0,
+  };
+
+  const [headingToWatch, entry] = useIntersection({ options });
+  const headingIsVisible = entry.isIntersecting;
+
+  // update activeHeading, passed up to Table of Contents
+  useEffect(() => {
+    if (headingIsVisible) {
+      setActiveHeading(id);
+    }
+  });
+
   return (
     <Box
+      ref={headingToWatch}
       direction="row"
       align="center"
       gap="small"
@@ -29,7 +61,7 @@ export const SubsectionHeader = ({ children, level }) => {
         <HighlightPhrase size="inherit">{children}</HighlightPhrase>
       </Heading>
       <Button
-        tip='Copy link to clipboard'
+        tip="Copy link to clipboard"
         a11yTitle={`Jump to section titled ${children} 
                     and copy link to clipboard`}
         icon={<LinkIcon color={over ? 'text-xweak' : 'transparent'} />}
--- a/aries-site/src/layouts/main/Layout.js
+++ b/aries-site/src/layouts/main/Layout.js
@@ -1,7 +1,9 @@
-import React, { Fragment, useEffect, useContext } from 'react';
+import React, { Fragment, useEffect, useContext, useState } from 'react';
 import { useRouter } from 'next/router';
 import PropTypes from 'prop-types';
 import { initialize, pageview } from 'react-ga';
+import { ActiveHeadingContext } from '../content/ActiveHeadingContext';
+
 import {
   Box,
   Main,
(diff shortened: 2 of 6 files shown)


aries-site/src/layouts/content/SubsectionHeader.js
```diff
@@ -1,16 +1,46 @@
...
+  const options = {
+    rootMargin: `${topMargin}% 0% ${bottomMargin}% 0%`,
+    // threshold: 1.0,
+  };
+
+  const [headingToWatch, entry] = useIntersection({ options });
+  const headingIsVisible = entry.isIntersecting;
+
+  // update activeHeading, passed up to Table of Contents
+  useEffect(() => {
```
MikeKingdom, line 35: Shouldn't this have dependencies so it's only calling setActiveHeading if id or headingIsVisible change?

aries-site/src/layouts/content/SubsectionHeader.js
```diff
@@ -1,16 +1,46 @@
...
+  // top and bottom margin values to calculate intersection window
+  const topMargin = large;
+  const bottomMargin = `-${2 * large}`;
+
+  const options = {
+    rootMargin: `${topMargin}% 0% ${bottomMargin}% 0%`,
+    // threshold: 1.0,
+  };
+
+  const [headingToWatch, entry] = useIntersection({ options });
```
MikeKingdom, line 31: This isn't quite making sense to me. Isn't the first thing returned the `setNode` function so you can set the node state in the hook? I would think you'd need to call this function with the ref.current in a useEffect rather than pass it like a reference on line 43. Or am I missing something?

aries-site/src/layouts/main/Layout.js
```diff
@@ -71,6 +70,22 @@ export const Layout = ({
...
   ].filter(link => link !== undefined);
 
+  const [activeHeading, setActiveHeading] = useState();
+  const activeHeadingContextValue = React.useMemo(() => {
+    const nextValue = { activeHeading, setActiveHeading };
+    return nextValue;
+  }, [activeHeading]);
+
+  // useEffect finds which page heading is on screen as the user scrolls
+  useEffect(() => {
```
MikeKingdom, line 80: I would definitely think this needs dependencies rather than always replacing the window.onscroll after each render.  I'd also think you'd use `window.addEventListener` instead so you're not affecting any other listeners? Something like:
```jsx
useEffect(() => {
  const onScroll = () => {
    if (window.pageYOffset === 0) {
      setActiveHeading(null);
    }
  };
  window.addEventListener('scroll', onScroll);
  return () => {
    window.removeEventListener('scroll, onScroll);
  };
}, [setActiveHeading]);
```

MikeKingdom (commented):
If I'm understanding what this is doing, another way to do this is to let the browser handle it via an anchor. You can define an anchor on the top of the section which you want to be able to scroll to:
```html
<a name=”somename”></a>
```
Then links from the table of contents to get to it would be hashtags like this:
```html
<a href=”#somename”>Go to the section</a>
```

aries-site/src/layouts/content/SubsectionHeader.js
JCoder121: I initially had the wrong dependencies which caused the Table of Contents scroll bar not to activate, but with id and headingIsVisible it works!

JCoder121:
> If I'm understanding what this is doing, another way to do this is to let the browser handle it via an anchor. You can define an anchor on the top of the section which you want to be able to scroll to:
> 
> ```
> <a name=”somename”></a>
> ```
> 
> Then links from the table of contents to get to it would be hashtags like this:
> 
> ```
> <a href=”#somename”>Go to the section</a>
> ```

I tried using <a href></a> method but it causes the keyboard navigation to focus twice on the same section (once for the button, once for the link), so I think <Link> might be preferable here as it only causes the keyboard to focus once per section header.

aries-site/src/layouts/content/SubsectionHeader.js
JCoder121: I went off of the example [here](https://github.com/HewlettPackard/hpe-js-project-scripts/tree/master/packages/react-hooks#useintersection) which is HPE's useIntersection documentation, but I'm not sure if it is outdated/not best practices. In the actual code, node and entry are returned, but does the current implementation accept the ref correctly?
image

MikeKingdom:
> I tried using method but it causes the keyboard navigation to focus twice on the same section (once for the button, once for the link), so I think might be preferable here as it only causes the keyboard to focus once per section header.

I would think this would be avoidable with use of `tabindex="-1"` in the appropriate spots. It just seems like you're having to do a very complicated solution to a type of navigation that is rather well supported by browsers

JCoder121:
> > I tried using method but it causes the keyboard navigation to focus twice on the same section (once for the button, once for the link), so I think might be preferable here as it only causes the keyboard to focus once per section header.
> 
> I would think this would be avoidable with use of `tabindex="-1"` in the appropriate spots. It just seems like you're having to do a very complicated solution to a type of navigation that is rather well supported by browsers

Got it, thanks for the feedback! I'll try that.

aries-site/src/layouts/content/SubsectionHeader.js
MikeKingdom: Very true. It's a little subtle but it seems like rather than actually using a ref it's relying on React to call the thing passed in as the `ref` like a setter function. e.g. https://flow.org/en/docs/react/refs/

JCoder121: Sorry I wasn't able to follow-up yesterday, but is it ok to keep it as is or should I change it to use ref.current in a useEffect?

MikeKingdom: No problem if it's working fine. Feel free to leave it as is.

taysea:
@JCoder121 I'm thinking we can close this PR for now since the current implementation is functioning properly. If you agree, feel free to close.

JCoder121:
Closing for now as DS functionality works without useIntersection hook.

PR #2852, opened 2022-08-24
Some initial analytics logging

### LINKED ISSUE #2819: Make use of onAnalytics in Design System Site
Instrument Design System Site with onAnalytics
- [x] Update grommet dependency
- [x] Add `onAnalytics` to any `<Grommet>` tags (simple console.log for now)
- [ ] Add data/id attributes to any interesting forms or action buttons

See #2818

MikeKingdom:
Add an onAnalytics callback to Grommet to log some basic analytics info to the console.

The next step will be to connect this to Amplify or some other analytics system like Adobe or Google Analytics

[Deploy Preview](https://deploy-preview-2852--keen-mayer-a86c8b.netlify.app/)

closes #2819 

ThemeMode.js

**General UX Checks**
- [ ] Light & dark modes

**Accessibility Checks**
- [ ] Keyboard interactions

**Code Quality Checks**

No

### DIFF
--- a/aries-site/src/layouts/main/ThemeMode.js
+++ b/aries-site/src/layouts/main/ThemeMode.js
@@ -2,7 +2,7 @@ import React from 'react';
 import { Box, Grommet } from 'grommet';
 import PropTypes from 'prop-types';
 import { aries } from '../../themes/aries';
-import { useDarkMode } from '../../utils';
+import { analytics, useDarkMode } from '../../utils';
 
 export const ThemeMode = ({ children, ...rest }) => {
   const { value } = useDarkMode(aries.defaultMode === 'dark');
@@ -25,6 +25,7 @@ export const ThemeMode = ({ children, ...rest }) => {
       theme={aries}
       themeMode={themeMode}
       full="min"
+      onAnalytics={analytics}
       {...rest}
     >
       {children}
--- a/aries-site/src/utils/analytics.js
+++ b/aries-site/src/utils/analytics.js
@@ -1,0 +1,5 @@
+// (C) Copyright 2022 Hewlett Packard Enterprise Development LP.
+
+export const analytics = (data) => {
+  console.log('\x1B[1;44m ANALYTICS \x1B[0m', data);
+};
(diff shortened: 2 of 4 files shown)


halocline:
Implementation looks fine. Some things I am noticing:

- Content cards clicks are not picked up by onAnalytics. Our content cards are built with `<Card as="a" .../>`. Wondering how we might capture these events... Support some native HTML elements? Rethink how we are approaching clickable cards --> `<Button href=""><Card .../></Button>` instead?
- skipTo links layerOpen, layerClose events
- The importance of meaningful ids or other unique identifiers.
- Support for List `onClickItem`
- Means to present meaningful ancestor when the target is generic div, svg, text, etc.

I'll do some more investigating.

MikeKingdom:
Thanks @halocline. I hadn't checked to see what that layer is really doing but it seemed odd that we did mount it. For the icon button targets and such I didn't find a clean way yet to get the best element ref given how the Button is constructed. 

In general we want nav or action types of events so I'm less sure about the List on click items and such. Note that you have to be extra careful not to actually send personal identifiable info to the real analytics engine (say a server hostname or IP or other selections etc.) With a list onClick it seems a bit too easy to accidentally send an overly identifiable key. Note that URLs have that danger too.

halocline:
> In general we want nav or action types of events so I'm less sure about the List on click items and such. 

What about navigation occurring when I select an item from a list? For example, I have a collection of resources presented in a List and I want to navigate to the resource detail when I click on the particular list item. On the DS site, in the search experience, I would like to know which search result was selected and navigated to. Without that, we really only have pageView to and from, but no context as to what triggered that navigation.

MikeKingdom:
> > In general we want nav or action types of events so I'm less sure about the List on click items and such.
> 
> What about navigation occurring when I select an item from a list? For example, I have a collection of resources presented in a List and I want to navigate to the resource detail when I click on the particular list item. On the DS site, in the search experience, I would like to know which search result was selected and navigated to. Without that, we really only have pageView to and from, but no context as to what triggered that navigation.

Isn't that typically an `<Anchor>` within the list item? Navigating from a row click would seem very odd. I know that's very true with DataTables (`<Anchor>` in a column cell).

halocline (approved):

PR #2864, opened 2022-08-30
Feedback submit option

britt6612:
This is another approach @taysea brought up to how we can handle the submit `buttons` for the user. In the other PR I have a prop `footerActions` in which the user can decide how the button for the submit are placed and how they want to use them.

This PR offers that we have a submit button in the component render by default. This way the user can pass if they want a `cancel` button or only a submit button. now the order not sure if they would be able to change that? someone may want `submit` on the right where someone else may want it on the left. This also brings up if `onSubmit` is required to send the data to an API then will the submit be required?

### DIFF
--- a/aries-site/src/components/feedback/Feedback.js
+++ b/aries-site/src/components/feedback/Feedback.js
@@ -19,8 +19,7 @@ export const Feedback = ({
   value,
 }) => {
   const theme = useContext(ThemeContext);
-
-  console.log(messages);
+  
   let content = (
     <Box {...theme?.feedback?.container}>
       <FeedbackHeader title={title}>
--- a/aries-site/src/components/feedback/Feedback.js
+++ b/aries-site/src/components/feedback/Feedback.js
@@ -6,8 +6,8 @@ import { FormClose } from 'grommet-icons';
 
 export const Feedback = ({
   children,
-  footerActions,
   layerProps,
+  messages,
   modal,
   onChange,
   onClose,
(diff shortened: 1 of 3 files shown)


aries-site/src/components/feedback/Feedback.js
```diff
@@ -6,8 +6,8 @@ import { FormClose } from 'grommet-icons';
 
 export const Feedback = ({
   children,
-  footerActions,
   layerProps,
+  messages,
```
taysea, line 10: This aligns with what I was thinking -- prop name TBD

aries-site/src/components/feedback/Feedback.js
```diff
@@ -46,19 +43,24 @@ export const Feedback = ({
       >
         <Box {...theme?.feedback?.body}>
           <>{children}</>
-          <Box {...theme?.feedback?.footer}>{footerActions}</Box>
+          <Box {...theme?.feedback?.footer}>
+            {/* how would we decide which order? offer reverse prop? */}
+            {/* what about passing in other button props? primary icon? */}
```
taysea, line 48: Maybe the buttons can be styled by the theme? the order can be handled by theme.feedback.footer.

It feels like for the first pass it's okay to be a bit prescriptive.

halocline (approved):

aries-site/src/components/feedback/Feedback.js
britt6612: okay cool but you like this direction of having a prop for the Buttons?

halocline: It seems reasonable, but undecided. Since we are prototyping, we can try it out and see how it feels.

britt6612: yeah of course Im going to fix merge conflicts then merge it into the main branch

PR #3005, opened 2022-10-25
Update guidance regarding optional vs. required field indicators

vavalos5:
[Deploy Preview](https://deploy-preview-3005--keen-mayer-a86c8b.netlify.app/templates/forms)

#### What does this PR do?
This PR updates our guidance on required vs optional fields 
forms.mdx
site

**General UX Checks**
- [ ] Light & dark modes

**Accessibility Checks**
- [ ] Keyboard interactions

**Code Quality Checks**

site
Should You Mark the Optional Fields, Too?
DS channel question
no
 backwards compatible


taysea (commented):
We probably need to also remove the text under "Do's" that says "If a field is optional, write "Optional" next to the form field label."

https://design-system.hpe.design/templates/forms?q=option#dos

britt6612:
> We probably need to also remove the text under "Do's" that says "If a field is optional, write "Optional" next to the form field label."
> 
> https://design-system.hpe.design/templates/forms?q=option#dos

yes, we were a little confused there seems to be guidance everywhere Ill go back and remove that.

aries-site/src/pages/templates/forms.mdx
```diff
@@ -215,6 +215,8 @@ Users must actively check a box next to a statement that shows they have agreed
 
 For standard forms, use an asterisk(\*) to indicate required field/controls.
 
+**Please note that at this time we do not have use cases that would be required to be labeld as optional. If you feel you have a use case that would need to be labeled as optional, please contact us to discuss further.**
```
MikeKingdom, line 218:
```suggestion
**Please note that at this time we do not have use cases that would be required to be labeled as optional. If you feel you have a use case that would need to be labeled as optional, please contact us to discuss further.**
```

MikeKingdom (commented):
Someone already mentioned the do's/don'ts section.

In the don'ts, some explanation of not having more than two optional fields might be in order.

In the Exceptions section it says to `Ensure to mark the field as optional.`

There are also a few placeholders that have `(optional)` in them (Mostly Address2 fields). Do we feel that is needed still?

KennyAtHPE:
Will update our Figma components/templates based on this change

britt6612:
> Someone already mentioned the do's/don'ts section.
> 
> In the don'ts, some explanation of not having more than two optional fields might be in order.
> 
> In the Exceptions section it says to `Ensure to mark the field as optional.`
> 
> There are also a few placeholders that have `(optional)` in them (Mostly Address2 fields). Do we feel that is needed still?

I believe I took out all the references to `optional` in the page

MikeKingdom (approved):

britt6612:
> Will update our Figma components/templates based on this change

thank you @KennyAtHPE

britt6612 (approved):

aries-site/src/pages/templates/forms.mdx
```diff
@@ -46,9 +46,8 @@ Complicated processes where the user might need guidance at various stages, or w
 
 - Ask only for the required information. Keep the fields to a minimum to allow the user to finish their task as quickly as possible.
 - List only the most common field requirements below the relevant field.
-- Use asterisks for required fields when the form has more than two required fields and/or has optional fields.
```
taysea, line 49: I feel like this should stay. For example, what if a for had 1 required field and 1 optional field -- wouldn't we want to mark the required field to differentiate it?

aries-site/src/pages/templates/forms.mdx
```diff
@@ -80,7 +79,6 @@ Complicated processes where the user might need guidance at various stages, or w
 
 **Format and fields**
 
-- Have more than two optional fields.
```
taysea, line 83: Shouldn't this guidance still apply? Aren't we trying to limit the amount of optional fields included in forms?

aries-site/src/pages/templates/forms.mdx
vavalos5: I agree with this. I think we should keep this section. We could just remove, " and/or has optional fields."

aries-site/src/pages/templates/forms.mdx
vavalos5: I agree. While we don't have solid guidance of when "optional" should be used, so we're sticking to  asterisks being mandatory and the rest without asterisks being "optional".

taysea: Agree but that's not what this section is talking about. This section is saying "Don't have more than two optional fields." and has nothing to do with marking as required or optional. I feel like that guidance is valid to stay.

aries-site/src/pages/templates/forms.mdx
taysea: I am saying we should not remove "and/or has optional fields" because of the example I listed in my comment.

aries-site/src/pages/templates/forms.mdx
britt6612: put back

MikeKingdom: > Agree but that's not what this section is talking about. This section is saying "Don't have more than two optional fields." and has nothing to do with marking as required or optional. I feel like that guidance is valid to stay.

Agreed. Might help to have a hint of why though.

aries-site/src/pages/templates/forms.mdx
```diff
@@ -215,6 +214,8 @@ Users must actively check a box next to a statement that shows they have agreed
 
 For standard forms, use an asterisk(\*) to indicate required field/controls.
 
+**Please note that at this time we do not have use cases that would be required to be labeled as optional. If you feel you have a use case that would need to be labeled as optional, please contact us to discuss further.**
```
taysea, line 217:
```suggestion
**Please note that at this time we do not have use cases that require labeling a form field as optional. If you feel you have a use case that would need to be labeled as optional, please contact us to discuss further.**
```

taysea (commented):
small wording suggestion

taysea (approved):

PR #3026, opened 2022-11-15
Select Component for Datatable Pagination example

### LINKED ISSUE #2948: Pagination – Add example of Select control for # of items per page
Request from Design System council for a pattern of how to show a control for the # of item per page. Feels like something to be added to "Pagination" page or "DataTable"/"Lists".

Deliverables:
- [ ] Example added on Design System site
- [ ] Guidance added of when to use, what default step should be, what options for steps should be (to connect with designer here)

JCoder121:

#### What does this PR do?
Provides a select component in the datatable pagination example, allowing users to choose how many results to display at a time instead of a hardcoded step size

aries-site/src/examples/components/pagination/PaginationServerTableExample.js

Manual and visual testing

**General UX Checks**
- [ ] Light & dark modes

**Accessibility Checks**
- [ ] Keyboard interactions

**Code Quality Checks**

Ensure that the user can increase and decrease step size and the table dynamically resizes to accommdate

Currently the user can increase the step size and the table resizes, but shrinking table size shows blank data and the table does not resize. However, the backend API fetch call for the data seems to be grabbing the correct number of rows.

Closes #2948.


aries-site/src/examples/components/pagination/PaginationServerTableExample.js
```diff
@@ -0,0 +1,172 @@
...
+          pad={{ vetical: 'small' }}
+          align="center"
+          direction="row-responsive"
+          gap="small"
+        >
+          <Text>Select</Text>
+          <Select
+            id="select"
+            name="select"
+            placeholder="10"
```
MikeKingdom, line 126: These aren't needed. You're not in a form and the `value` is always set so you'll never see the placeholder

aries-site/src/examples/components/pagination/PaginationServerTableExample.js
```diff
@@ -0,0 +1,172 @@
...
+
+  return (
+    <Box pad="small">
+      <Heading id="server-side-pagination-heading" level={3}>
+        Launches
+      </Heading>
+
+      {numberItems > limit && (
+        <Box
+          pad={{ vetical: 'small' }}
```
MikeKingdom, line 117: misspelled so you actually have no pad.

aries-site/src/examples/components/pagination/PaginationServerTableExample.js
```diff
@@ -0,0 +1,172 @@
...
+          align="center"
+          direction="row-responsive"
+          gap="small"
+        >
+          <Text>Select</Text>
```
MikeKingdom, line 122: Would a better word be `Show` ?

aries-site/src/examples/components/pagination/PaginationServerTableExample.js
```diff
@@ -0,0 +1,172 @@
...
+        </Box>
+      )}
+      <Box>
+        <DataTable
+          aria-describedby="server-side-pagination-heading"
+          columns={columns}
+          data={data}
+          sort={{ ...sort, external: true }}
+          replace
+          step={limit}
```
MikeKingdom, line 141: Remove the `replace` and the `step`. You don't need either since you're not doing infinitescroll and you're handling the step and pagination yourself. I think this fixes the issue you saw with selecting a smaller step in the select and the table going empty. Not sure why though. Sounds like a potential DataTable bug

JCoder121: It turns out that the issue was with the `replace` prop! It makes sense to remove the `step` prop as well, since it is being handled below in the Pagination component. Thanks for your feedback!

aries-site/src/examples/components/pagination/PaginationServerTableExample.js
JCoder121: Should I put the Select component in a Form/FormField component? I'm getting a WAVE check "missing form label" error, and I've noticed that the Select components on the guidance page [here](https://design-system.hpe.design/components/select) are wrapped in a Form component.

MikeKingdom: No reason to have it within a Form let alone a FormField. FormField provides label and borders plus some validation and such when used in a Form, none of which is needed here. And you never need to do a form submit or any of the other logic related to a form so I think WAVE check is overly opinionated.

aries-site/src/examples/components/pagination/PaginationServerTableExample.js
```diff
@@ -53,13 +62,16 @@ const columns = [
...
+  const size = useContext(ResponsiveContext);
   const [sort, setSort] = useState({ property: 'name', direction: 'asc' });
   const [data, setData] = useState([]);
   const [numberItems, setNumberItems] = useState(0);
   const [page, setPage] = useState(1);
-  const limit = 10;
+  const [limit, setLimit] = useState(10);
   const pageResultStart = (page - 1) * limit + 1;
   const pageResultEnd = Math.min(page * limit, numberItems);
+  const options = [5, 10, 25, 50];
```
taysea, line 73: We can pull this `options` variable outside of the PaginationServerTableExample since it's not reliant on anything in here. (just placing on the line before, like "columns" is fine.

aries-site/src/examples/components/pagination/PaginationServerTableExample.js
```diff
@@ -53,13 +62,16 @@ const columns = [
...
+  const smallScreenSize = ['xsmall', 'small'].includes(size);
```
taysea, line 74:
```suggestion
  const mobile = ['xsmall', 'small'].includes(size);
```

aries-site/src/examples/components/pagination/PaginationServerTableExample.js
```diff
@@ -104,36 +116,51 @@ export const PaginationServerTableExample = () => {
...
-            </Text>
-            <Pagination
-              step={limit}
-              numberItems={numberItems}
-              page={page}
-              onChange={opts => setPage(opts.page)}
-              direction="row"
-              flex={false}
-            />
+            {!smallScreenSize && (
```
taysea, line 138: I'm wondering if the ordering of this should be altered so that the results summary ("Showing X of Y") is on the far left, then we have the controls that affect the view grouped. So something like:

```
<Box direction="row" align="center" gap="small">
      <Text>Showing X of Y</Text>
      <Box direction ="row" gap="small" align="center">
            <ResultsPerPageSelect />
            <Pagination />
      </Box>
</Box>

File: aries-site/src/examples/components/pagination/PaginationServerTableExample.js [hunk 8]
```diff
@@ -104,35 +116,50 @@ export const PaginationServerTableExample = () => {
...
+            align="center"
+            wrap
           >
-            <Text>
-              Showing {pageResultStart}-{pageResultEnd} of {numberItems}
-            </Text>
+            {!mobile && (
+              <Box direction="row-responsive" gap="medium" align="center">
+                <Text>
+                  Showing {pageResultStart}-{pageResultEnd} of {numberItems}
```
taysea, line 143: I was expecting that the "Show X of Y" text would be on its own and the "Rows per page" would be grouped in a Box with the Pagination control

aries-site/src/examples/components/pagination/PaginationServerTableExample.js
```diff
@@ -104,35 +116,50 @@ export const PaginationServerTableExample = () => {
...
           <Box
             direction="row-responsive"
-            fill="horizontal"
             border="top"
-            justify="end"
-            pad={{ vertical: 'xsmall' }}
+            gap="medium"
+            pad={{ vertical: 'small' }}
+            align="center"
+            wrap
```
taysea, line 138:
```suggestion
            justify="between"
            wrap
```
This will make sure any extra space is put between the "Show X of Y" and the group containing the select + pagination.

taysea:
Screen Shot 2022-12-14 at 12 22 10 PM

Noticing some behavior when I collapse the screen where the select is getting squished

JCoder121:
Progress as of 1/13/22:
- Finishing up defining the success criteria for this Pagination Datatable example, to be finalized today (main points below)
- Incorporate table header (search bar, filtering actions, batch actions button) in accordance to existing examples
- Clearly differentiate between pagination + row controls and table results information

Remaining work:
- Finalize success criteria based on feedback
- Code all components in the example adhering to success criteria
- Test layout and functionality for any issues

JCoder121:
I will touch base with Taylor later this afternoon (1/17/23) to discuss future steps of this ticket. I think that the Items per page Select component has potential to be included in the Toolbar/Filtering section of the Data component, and if that is the direction that we choose to move forward in, this ticket description can change to include refactoring the current code to use the Data component and adding a Select component inside Data's filtering actions.

JCoder121:
1/20/23:
Keep the row select component at the bottom of the datatable, inline with the pagination component. Remove the search/filtering header, to be added to the example in a future refactoring ticket. Continue designing and measuring options based on success criteria.

taysea:
Closing this PR since we've added `stepOptions` to pagination directly.

PR #3077, opened 2023-01-06
start of skeleton page

### LINKED ISSUE #1600: Skeleton - Implement first pass of site page
[Current Guidance
](https://www.figma.com/file/dQG3xZ6m4AUt7Lwrdp1gVB/Skeleton-Guidance?node-id=216%3A1663&t=m7Kpg99SjOyveF02-1) Related Ticket #3088 

## Deliverables

- [x] Add to grommet-site
- [ ] Create a first pass of the Skeleton page on the DS site. Sections to include:
     - [ ] Main example
     - [ ] When to use/when not to use
     - [ ] Anatomy

britt6612:
[Deploy Preview](https://deploy-preview-3077--keen-mayer-a86c8b.netlify.app/components/skeleton)
[Figma link](https://www.figma.com/file/dQG3xZ6m4AUt7Lwrdp1gVB/Skeleton-Guidance?node-id=300%3A9435&t=xB9b5v2e14DyTAU3-0)

#### What does this PR do?
Skeleton Page 
skeleton.mdx
locally

**General UX Checks**
- [ ] Light & dark modes

**Accessibility Checks**
- [ ] Keyboard interactions

**Code Quality Checks**

 Create a first pass of the Skeleton page on the DS site. Sections to include:
- [ ]  Main example
- [ ]  When to use/when not to use
- [ ]  Anatomy

closes #1600 
yes
backwards compatible


aries-site/src/examples/components/skeleton/SimpleSkeleton.js
```diff
@@ -0,0 +1,10 @@
+import React from 'react';
+import { Box, Text, Skeleton } from 'grommet';
+
+export const SkeletonExample = () => (
```
MikeKingdom, line 4: You kind of hint at it in the Anatomy table, but when you do examples, keep in mind that we're trying to avoid folks using the `<Skeleton/>` component directly. Instead we're hoping they're dashboard cards or other components can just be skeleton-ized by simply using the 'skeleton' prop to 'Box' or 'Card' and let the contained 'Text', 'Heading', etc to render themselves as the gray box.

aries-site/src/data/structures/components.js
```diff
@@ -1191,6 +1192,30 @@ export const components = [
       'spooling',
     ],
   },
+  {
+    name: 'Skeleton Loading',
```
taysea, line 1196:
```suggestion
    name: 'Skeleton',
```
Should this be called "Skeleton" to align with the component name?

aries-site/src/data/structures/components.js
```diff
@@ -1191,6 +1192,30 @@ export const components = [
...
+    category: 'Visualizations',
+    description:
+      'Skeleton is used to provide a sense of the structure of the page and the ability to anticipate where certain pieces and types of content are loading. Loading patterns are used to communicate that data is loading which can improve a user experience.',
```
taysea, line 1199:
```suggestion
      'Skeletons tell the user that data is loading and provide a general sense of the page elements and structure. Their subtle animation conveys a sense of forward progress which can improve the user experience.',
```

aries-site/src/examples/components/skeleton/DashBoardSkeletonExample.js
```diff
@@ -0,0 +1,20 @@
...
+
+export const DashBoardSkeletonExample = () => {
+  const [loading, setLoading] = useState(true);
+  useEffect(() => {
+    setTimeout(() => setLoading(!loading), 3000);
+  }, [loading]);
+
+  return (
+    <>
+      <Box width="large" skeleton={loading ? skeleton : undefined}>
```
taysea, line 15: I noticed we don't have any animation on this. Is that going to be handled in a separate PR/next step?

taysea, line 15: Should we use `Page` and `PageContent`? Does `skeleton` work on PageContent? (i'm thinking it should be PageContent is a Box.

aries-site/src/examples/components/card/ActivitiesNavigationalCards.js
```diff
@@ -37,6 +37,7 @@ export const ActivitiesNavigationalCards = ({ heading = true }) => {
               />
             }
             level={3}
+            skeleton={skeleton}
```
taysea, line 40: I know you didn't touch it in this code, but can we sentence case "Activities and tasks" and use "and" instead of &?

aries-site/src/pages/components/skeleton-loading.mdx
```diff
@@ -0,0 +1,33 @@
...
+  componentName="Skeleton"
+  code="https://raw.githubusercontent.com/grommet/hpe-design-system/master/aries-site/src/examples/components/skeleton/DashBoardSkeletonExample.js"
+  docs="https://v2.grommet.io/skeleton?theme=hpe#props"
+>
+  <DashBoardSkeletonExample />
+</Example>
+
+### When to use
+Skeletons are used when:
+- Loading elements are known ahead of time but are not immediately available.
```
taysea, line 15: When is an example of when we don't know what kind of elements we'd be rendering? It feels like generally, we know what elements will be rendered but we are waiting on the data to populate those elements

aries-site/src/pages/components/skeleton-loading.mdx
```diff
@@ -0,0 +1,33 @@
...
+- All of the content on a page is loaded in at the same time.
+- Users are loading high-traffic pages, such as a user dashboard.
```
taysea, line 17: Wouldn't skeletons apply to any page? not just high-traffic ones

aries-site/src/pages/components/skeleton-loading.mdx
```diff
@@ -0,0 +1,33 @@
...
+<Example
+  componentName="Skeleton"
+  code="https://raw.githubusercontent.com/grommet/hpe-design-system/master/aries-site/src/examples/components/skeleton/DashBoardSkeletonExample.js"
+  docs="https://v2.grommet.io/skeleton?theme=hpe#props"
+>
+  <DashBoardSkeletonExample />
+</Example>
+
+### When to use
+Skeletons are used when:
```
taysea, line 14: Not sure where this would go (probably elsewhere in the gudiance) we should mention that data should be rendered as soon as it's fetched. Using Instagram as an example, there may be multiple fetches happening (1 for stories and another for the main feed). If the stories response comes back before the main feed one, we should render the stories right away and just leave the main feed as a skeleton).

With how grommet skeleton is built, this is easy to implement by just controlling what container you flag the `skeleton` prop on. Might be a nice opportunity to highlight the ease of implementing with Grommet.

aries-site/src/data/structures/components.js
britt6612: good call

aries-site/src/examples/components/card/ActivitiesNavigationalCards.js
britt6612: yeah I wasnt sure if this was the best to touch this file or copy over code and change it in another file

aries-site/src/examples/components/skeleton/DashBoardSkeletonExample.js
britt6612: yeah I left that out while Kenny is looking more into what we want for guidance we can have it as a next step

aries-site/src/pages/components/skeleton-loading.mdx
britt6612: @KennyAtHPE tagging for eyes

aries-site/src/pages/components/skeleton-loading.mdx
britt6612: agree this was some feedback we got from Eric I think were going to include just not sure where but noted ✅

aries-site/src/pages/components/skeleton-loading.mdx
britt6612: 🤔 hmm maybe we can reword @KennyAtHPE

aries-site/src/examples/components/skeleton/DashBoardSkeletonExample.js
taysea: Perfect, makes sense.

aries-site/src/examples/components/card/ActivitiesNavigationalCards.js
taysea: I think we can just touch this file.

aries-site/src/pages/components/skeleton-loading.mdx
taysea: Cool sounds good.

aries-site/src/pages/components/skeleton-loading.mdx
```diff
@@ -0,0 +1,33 @@
+import { Notification } from 'grommet';
```
taysea, line 1: We'll need to update this file name from "skeleton-loading" to "skeleton" for the routing to work

britt6612: done

aries-site/src/examples/components/skeleton/DashBoardSkeletonExample.js
britt6612: changed it to `page` and `pageContent` it jumps a little but let me know what you think

aries-site/src/pages/components/skeleton.mdx
```diff
@@ -0,0 +1,37 @@
...
+  <SkeletonAnatomy />
+</Example>
+
+
+Skeletons come in three different shapes - circular, rounded, and rectangular. The shape, size, and elevation of a Skeleton are all dependent on the UI elements that are loading. 
+
+Some common components that may be skeletonized are:
+- Text, paragraph, heading
+- Avatar, image
+- Button, footer
```
taysea, line 36: I'd group "Footer" with the layout elements in the last bullet point

aries-site/src/pages/components/skeleton.mdx
```diff
@@ -0,0 +1,37 @@
...
+- Card, table, pageheader
```
taysea, line 37: Is this list supposed to be the components that currently allow `skeleton` in Grommet? or just general? If specific, it might be nice to call this out in an implementation note and/or link to the grommet docs for skeleton

taysea: Something like "Grommet has added `skeleton` property in order to make it easy to switch between a skeleton state and data-rich state for a variety of components. The components that currently support `skeleton` are:"

britt6612: good point Ill make that clear

aries-site/src/examples/components/skeleton/SkeletonAnatomy.js
```diff
@@ -0,0 +1,41 @@
+import React from 'react';
+import { Box, Button, Card, CardBody, Paragraph } from 'grommet';
+
+export const SkeletonAnatomy = ({ ...rest }) => (
+  <Card id="container" flex={false} width="large" {...rest}>
+    <CardBody align="start" gap="medium">
+      <Box align="center" gap="small" direction="row">
+        <Box
```
MikeKingdom, line 8: Why not use Text in a `<Box skeleton />` or at least `<Skeleton>` for these? More likely to just work if theme changes.

aries-site/src/examples/templates/Card/Card.js
```diff
@@ -14,6 +14,18 @@ import {
 } from 'grommet';
 import { adjustPad } from './utils';
 
+//  demo purposes to make a skeleton icon for loading
+const IconSkeleton = () => (
+  <Box
```
MikeKingdom, line 19: really this should instead either directly use `<Skeleton/>` or a `<Box background="whatever"/>` rather than specifying the skeleton prop (I know I'm trying to avoid specific uses of Skeleton). You should be able to avoid the conditional around where it's used if it's something the icon can be inside anyway.

aries-site/src/examples/templates/Card/Card.js
```diff
@@ -25,13 +37,17 @@ const DesignSystemCard = ({
...
+  skeleton,
   subtitle,
   pretitle,
   title,
   ...rest
 }) => {
   const theme = useContext(ThemeContext);
 
+  // demo purposes for using skeleton
+  const skeletonAlign = skeleton ? 'none' : 'start';
```
MikeKingdom, line 49: would ne nice to figure out how to avoid having to specify this and still get the right 100% behavior

aries-site/src/examples/templates/Card/Card.js
```diff
@@ -45,14 +61,18 @@ const DesignSystemCard = ({
...
               />
             )}
-            {icon && <Box pad={{ bottom: 'small' }}>{icon}</Box>}
+            {icon && skeleton ? (
+              <IconSkeleton />
+            ) : (
+              <Box pad={{ bottom: 'small' }}>{icon}</Box>
+            )}
             {pretitle && <Text size="small">{pretitle}</Text>}
             {title && typeof title === 'string' ? (
```
MikeKingdom, line 77: in the loading situation wouldn't the title be undefined? We still want the heading there in this situation so it can be a skeleton.

aries-site/src/data/structures/components.js
```diff
@@ -1191,6 +1192,30 @@ export const components = [
...
+        background: 'background-front',
+      },
+    seoDescription:
+      'Skeletons act as placeholder previews for UI elements so that users are provided with a structure of a loading page. This helps users to anticipate where and what types of content are loading.',
+    relatedContent: ['Box', 'DataTable', 'Spinner'],
+    sections: [
+      'When to use Spinner',
+      'Spinner within a list',
+      'Spinner with announcement on screen reader',
+      'Spinner loading content',
```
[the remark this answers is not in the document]
britt6612: no updated

aries-site/src/examples/templates/Card/Card.js
britt6612: agree I think I would have to go back and look at grommet code

aries-site/src/examples/templates/Card/Card.js
britt6612: Im not sure Im following,.. the `title` is still present in the skeleton render

aries-site/src/examples/components/skeleton/SkeletonAnatomy.js
britt6612: I wanted the `rounding` and Im not able to pass `round` to the parent `box` of `text`
so I changed to `skeleton`

aries-site/src/examples/templates/Card/Card.js
MikeKingdom: My point is I think it's probably working now because a title is always passed but in a real world situation you wouldn't have the title yet during loading but you still want a `<Heading>` component to provide the structure for the skeletonized view. Ie. this should still render a `<Heading>` with an empty title in a real world example

britt6612: aw okay thank you sorry I wasnt sure what you meant

aries-site/src/examples/templates/Card/Card.js
britt6612: changed this let me know what you think with the `icon`

aries-site/src/examples/components/skeleton/DashBoardSkeletonExample.js
```diff
@@ -0,0 +1,99 @@
...
+    setTimeout(() => setLoading(!loading), 3000);
+  }, [loading]);
+
+  return (
+    // passing skeleton to Page will skeletize all of the children that are
+    // text, button, box, paragraph, heading
+    <Page skeleton={loading ? skeleton : undefined} height="large">
+      <PageContent>
+        <Grid columns={columns[breakpoint]} gap="medium">
+          {activities.map((activity, index) => (
```
MikeKingdom, line 46: I still think this needs to do something different during the 'loading' phase. It shouldn't be just iterating over the same activities. That would never happen in a real situation.

I'd do something like:
```jsx
const placeholders = [{title: 'placeholder' }, {title: 'placeholder'}, {title: 'placeholder'}];
const items = loading ? placeholders : activities;
```
then here on 46 so `items.map` rather than `activities.map`. You'll have to adjust the code to handle minimal content in the activity or put enough stuff in the placholder.

MikeKingdom (commented):
Better. Can we make the grid cell height force the cards to be the same height?

aries-site/src/examples/components/skeleton/DashBoardSkeletonExample.js
```diff
@@ -0,0 +1,99 @@
...
+            <Box>
```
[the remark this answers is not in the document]
britt6612: good eye no dont need I will take it out

aries-site/src/examples/components/skeleton/DashBoardSkeletonExample.js
```diff
@@ -0,0 +1,99 @@
...
+                        width="24px" // size of icon
+                        height="24px"
+                        background="background"
+                        flex="grow"
+                        margin={{ bottom: 'small' }}
+                      />
+                    ) : (
+                      <Box pad={{ bottom: 'small' }}>{item?.icon}</Box>
+                    )}
+                    <Heading
```
taysea, line 72: Screen Shot 2023-01-23 at 10 46 18 AM

it's a little strange how the padding on the right side of the card isn't there. I'd expect the heading to be a bit narrower so the padding can be there

aries-site/src/pages/components/skeleton.mdx
```diff
@@ -0,0 +1,42 @@
...
+
+Some common components that may be skeletonized are:
+- text, paragraph, heading
+- avatar, image
+- button,
+- card, table, pageheader, footer, header
+
+Grommet has added `skeleton` property in order to make it easy to switch between a skeleton state and data-rich state for a variety of components.
+The components that currently support skeleton are:
+- box, button, heading, paragraph, text
```
taysea, line 42: Should we capitalize these since we are referring directly to the Grommet components?

aries-site/src/examples/components/skeleton/DashBoardSkeletonExample.js
MikeKingdom: That's because the Heading skeleton has a width fixed by the theme. So as the cards shrink too much they start to not have space for the heading skeleton.

aries-site/src/examples/components/skeleton/DashBoardSkeletonExample.js
```diff
@@ -0,0 +1,99 @@
...
+
+  return (
+    // passing skeleton to Grid will skeletize all of the children that are
+    // text, button, box, paragraph, heading
+    <Page height="large">
+      <PageContent>
+        <Box skeleton={loading ? skeleton : undefined}>
+          <Grid columns={columns[breakpoint]} gap="medium">
+            {items.map((item, index) => {
+              console.log(item);
```
MikeKingdom, line 51: should probably get rid of this now

aries-site/src/examples/components/skeleton/DashBoardSkeletonExample.js
taysea: Gotcha, wondering if we should adjust the width in theme to be something like:

```
width: { max: insert width, width: 100% }
```

I think that would allow us to set a max width and make sure the skeleton fills available space

aries-site/src/examples/components/skeleton/DashBoardSkeletonExample.js
```diff
@@ -0,0 +1,99 @@
...
+              return (
+                <Card key={index}>
```
MikeKingdom, line 53: I'm guessing this needs a flex="grow" to keep it from shrinking below the size needed for the heading (and maybe true for the CardHeader too.)

MikeKingdom (commented):
Looking good and getting close

aries-site/src/examples/components/skeleton/DashBoardSkeletonExample.js
MikeKingdom: Heading with width 100% doesn't look good as a skeleton. It might be worth setting a better min/max in pixels though

aries-site/src/pages/components/skeleton.mdx
britt6612: yes I can

aries-site/src/examples/components/skeleton/DashBoardSkeletonExample.js
britt6612: I can play around with this in Grommet and see for a min/max.

aries-site/src/pages/components/skeleton.mdx
britt6612: done

aries-site/src/examples/components/skeleton/DashBoardSkeletonExample.js
britt6612: done

aries-site/src/examples/components/skeleton/DashBoardSkeletonExample.js
```diff
@@ -0,0 +1,92 @@
...
+
+  return (
+    // passing skeleton to Grid will skeletize all of the children that are
+    // text, button, box, paragraph, heading
+    <Page height="large">
+      <PageContent>
+        <Box skeleton={loading ? skeleton : undefined}>
+          <Grid columns={columns[breakpoint]} gap="medium">
+            {items.map((item, index) => (
+              <Card flex="grow" key={index}>
```
MikeKingdom, line 51: `flex="grow"` isn't sufficient here. Does it also need to be on `CardHeader` (ie. the thing containing the `Heading`)?

image

MikeKingdom (approved):
Looks better. Thanks!

britt6612:
> Looks better. Thanks!

cool any thoughts on if this `min and max` should be in grommet or should we put in hpe theme?

ericsoderberghp:
> 

I lean towards grommet

taysea:
Screen Shot 2023-01-26 at 8 21 38 AM

Given there are only 3 cards, do we need all this extra white space at the bottom of the example? feels like we don't need the example to be so tall.

MikeKingdom:
> Screen Shot 2023-01-26 at 8 21 38 AM
> 
> Given there are only 3 cards, do we need all this extra white space at the bottom of the example? feels like we don't need the example to be so tall.

I think the height is fixed to something that works when you shrink the width such that cards wrap to other rows. We were also trying to avoid it changing height as it flips between skeletonized and not skeletonized (since it bounces the text you're trying to read.) Thoughts on a good compromise? It looks like height=large seems a bit big but height=medium is too small for when cards wrap.

britt6612:
> > Screen Shot 2023-01-26 at 8 21 38 AM
> > Given there are only 3 cards, do we need all this extra white space at the bottom of the example? feels like we don't need the example to be so tall.
> 
> I think the height is fixed to something that works when you shrink the width such that cards wrap to other rows. We were also trying to avoid it changing height as it flips between skeletonized and not skeletonized (since it bounces the text you're trying to read.) Thoughts on a good compromise? It looks like height=large seems a bit big but height=medium is too small for when cards wrap.

you are correct I can try `medium` but yes for this the page was bouncy when the content got loaded in which was a little distracting when trying to read the guidance.

aries-site/src/examples/components/skeleton/DashBoardSkeletonExample.js
```diff
@@ -0,0 +1,92 @@
...
+    setTimeout(() => setLoading(!loading), 3000);
+  }, [loading]);
+
+  const placeholders = [{}, {}, {}];
+  const items = loading ? placeholders : activities;
+
+  return (
+    // passing skeleton to Grid will skeletize all of the children that are
+    // text, button, box, paragraph, heading
+    <Page height="large">
```
taysea, line 46:
```suggestion
    <Page>
```
Instead of setting the height here, we can set it on the "Example" container and do a custom value (since it won't be exposed)

aries-site/src/pages/components/skeleton.mdx
```diff
@@ -0,0 +1,37 @@
+import { Notification } from 'grommet';
+import { Example } from '../../layouts';
+import { DashBoardSkeletonExample, SkeletonAnatomy } from '../../examples';
+
+<Example
+  componentName="Skeleton"
+  code="https://raw.githubusercontent.com/grommet/hpe-design-system/master/aries-site/src/examples/components/skeleton/DashBoardSkeletonExample.js"
+  docs="https://v2.grommet.io/skeleton?theme=hpe#props"
+  width="large"
```
taysea, line 9:
```suggestion
template
height="500px"
```
^^ this should allow us to set a reasonable height that won't jump. See if this works.

taysea:
> > > Given there are only 3 cards, do we need all this extra white space at the bottom of the example? feels like we don't need the example to be so tall.
> > 
> > 
> > I think the height is fixed to something that works when you shrink the width such that cards wrap to other rows. We were also trying to avoid it changing height as it flips between skeletonized and not skeletonized (since it bounces the text you're trying to read.) Thoughts on a good compromise? It looks like height=large seems a bit big but height=medium is too small for when cards wrap.
> 
> you are correct I can try `medium` but yes for this the page was bouncy when the content got loaded in which was a little distracting when trying to read the guidance.

I've dropped a few comments about how we might handle this in the "Example" height and do a custom value. This cleans things up so we don't have a fixed height in the example code itself (since we wouldn't want people copy and pasting that into their app anyway).

britt6612:
> > > > Given there are only 3 cards, do we need all this extra white space at the bottom of the example? feels like we don't need the example to be so tall.
> > > 
> > > 
> > > I think the height is fixed to something that works when you shrink the width such that cards wrap to other rows. We were also trying to avoid it changing height as it flips between skeletonized and not skeletonized (since it bounces the text you're trying to read.) Thoughts on a good compromise? It looks like height=large seems a bit big but height=medium is too small for when cards wrap.
> > 
> > 
> > you are correct I can try `medium` but yes for this the page was bouncy when the content got loaded in which was a little distracting when trying to read the guidance.
> 
> I've dropped a few comments about how we might handle this in the "Example" height and do a custom value. This cleans things up so we don't have a fixed height in the example code itself (since we wouldn't want people copy and pasting that into their app anyway).

sounds good Ill check out the changes you suggested

MikeKingdom:
> > I've dropped a few comments about how we might handle this in the "Example" height and do a custom value. This cleans things up so we don't have a fixed height in the example code itself (since we wouldn't want people copy and pasting that into their app anyway).
> 
> sounds good Ill check out the changes you suggested

Thanks Taylor. I learned a long time ago assume people will cut and paste everything without learning much about it so good to think of that ahead of time :)

taysea (approved):
LGTM, thanks for all of iterations!

PR #3122, opened 2023-01-27
Add backdrop-filter to layer overlay and align DataTable pinned

taysea:

#### What does this PR do?

Adds blur to layer overlay for "modal" layers.

modal={true}
Screen Shot 2023-01-27 at 12 59 25 PM

modal={false)
Screen Shot 2023-01-27 at 12 59 14 PM

**General UX Checks**
- [ ] Light & dark modes

**Accessibility Checks**
- [ ] Keyboard interactions

**Code Quality Checks**

### DIFF
--- a/aries-site/src/themes/aries.js
+++ b/aries-site/src/themes/aries.js
@@ -240,6 +240,8 @@ export const aries = deepMerge(hpe, {
     overlay: {
       background: '#0000001F',
     },
+    // temp CSS selector to target Layer overlay
+    extend: '> div { backdrop-filter: blur(12px); }',
   },
   calendar: {
     // using level as a means of styling doesn't seem like the best...
--- a/aries-site/src/themes/aries.js
+++ b/aries-site/src/themes/aries.js
@@ -205,6 +205,17 @@ export const aries = deepMerge(hpe, {
       weight: 400,
       color: 'text-strong',
     },
+    pinned: {
+      header: {
+        extend: 'backdrop-filter: blur(12px);',
+      },
+      body: {
+        extend: 'backdrop-filter: blur(12px);',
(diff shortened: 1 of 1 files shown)


aries-site/src/themes/aries.js
```diff
@@ -240,6 +240,8 @@ export const aries = deepMerge(hpe, {
     overlay: {
       background: '#0000001F',
     },
+    // temp CSS selector to target Layer overlay
+    extend: '> div { backdrop-filter: blur(12px); }',
```
halocline, line 244: Should we align this blur with the same that we have on `datatable.pinned`? What is the visual difference between 8px vs 12px? While we are looking at blur and with the Firefox support, are we able to clean up the opacity in `datatable.pinned` as well?

taysea: Here is 12px vs 8px. Personally, I prefer the extra blurring on 12px because it softens it to the point where my eyes aren't trying to make out the content at all. 8px feels a little distracting to me.

12px
Screen Shot 2023-01-27 at 12 59 25 PM

8px
Screen Shot 2023-01-27 at 1 10 25 PM

Agree that will Firefox support of backdrop-filter now, we can cleanup datatable.pinned. I'll look at that.

taysea: On DataTable, the difference between at 8px and 12px blur isn't really noticeable so I'd pitch that whatever we decide for layer we update DataTable.pinned to follow.

halocline: right on. Let's go with 12px for now. We can refine when we get into true support for all of these changes.

halocline (approved):

PR #3191, opened 2023-03-07
Use heading for Card title instead of text

### LINKED ISSUE #3141: Brand Refresh - Update Card examples to use correct semantic heading
Some [older card examples](https://design-system.hpe.design/templates/filtering?q=filterin) implement card titles as plain text with weight="bold". These should be updated to use a Heading with correct semantic level.

## Acceptance Criteria

- [ ] no Cards use Text for titles

taysea:
[Deploy Preview](https://deploy-preview-3191--keen-mayer-a86c8b.netlify.app/)

#### What does this PR do?

Updates instances of Card in examples and site layout to use proper semantic heading level.

Also removed unused files as I came across them.

**General UX Checks**
- [ ] Light & dark modes

**Accessibility Checks**
- [ ] Keyboard interactions

**Code Quality Checks**

Closes #3141


aries-core/src/js/components/core/Identifier/Identifier.js
```diff
@@ -1,29 +1,44 @@
...
+  size,
+  ...rest
+}) => (
   <Box align="center" {...rest}>
     {children}
     <Box>
-      <Text size={size} weight="bold" color="text-strong">
-        {title}
-      </Text>
+      {level ? (
```
halocline, line 17: I like this thought.

aries-site/src/components/home/Highlights.js
```diff
@@ -44,16 +43,11 @@ const HighlightsLayout = () => {
...
               </Box>
-              <Box fill>
-                <Text
-                  weight="bold"
-                  size="large"
-                  margin={{ top: 'small' }}
-                  color="text-strong"
-                >
+              <Box fill gap="xsmall">
+                <Heading level={3} margin={{ top: 'small', bottom: 'none' }}>
```
halocline, line 47: Do we need both gap and margin specified? Can we eliminate one?

aries-site/src/examples/components/namevaluelist/Examples/NameValueListSummaryExample.js
```diff
@@ -13,9 +14,11 @@ export const NameValueListSummaryExample = () => (
   <Card width="medium">
     <CardHeader gap="none" align="start" direction="column">
       <Text>Catalog Lifecyle Manager</Text>
-      <Text size="xxlarge" color="text-strong" weight="bold">
+      {/* adjust level according to the correct 
+      semantic level for your layout */}
```
halocline, line 18:
```suggestion
      semantic heading level for your layout */}
```

aries-site/src/examples/components/skeleton/DashBoardSkeletonExample.js
```diff
@@ -67,6 +68,8 @@ export const DashBoardSkeletonExample = () => {
                   ) : (
                     <Box pad={{ bottom: 'small' }}>{item?.icon}</Box>
                   )}
+                  {/* adjust level according to correct semantic level
```
halocline, line 71:
```suggestion
                  {/* adjust level according to correct semantic heading level 
```

aries-site/src/examples/templates/filtering/FilteringCards/FilteringCards.js
```diff
@@ -78,30 +76,26 @@ const Users = () => {
...
+            title={datum.name}
+            subtitle={datum.location}
+            actions={
               <Box>
                 <Text size="small">Role</Text>
                 <Text color="text-strong">{datum.role}</Text>
               </Box>
-            </CardBody>
-          </Card>
+            }
```
halocline, line 96: This seems like we need to revisit how our Card is built up. "actions" doesn't seem like the right name/place for this type of content. Would treating this content as `children` make more sense? Children could then be inserted as child of CardBody?

halocline (commented):
Minor comments, otherwise looking very nice.

aries-site/src/components/home/Highlights.js
taysea: Good call, I'll look to remove `margin` if possible.

aries-site/src/examples/templates/filtering/FilteringCards/FilteringCards.js
taysea: Could call, I'll take a look at that. I was feeling the same.

JCoder121 (approved):
Overall looks good! One final question I had related to the Card component specifically: in /templates/Card/Card.js `DesignSystemCard` is exported as `Card`, used with `import { Card } from '../../templates/Card/Card';`. I was wondering if this might cause naming confusion or vagueness between grommet's existing `Card` and other DS-specific Card implementations that build off of grommet's `Card` that accept different or additional properties. However, I realize this naming scheme might be a site-wide nuance, which I know I would require further close inspection. Just an idea 😄

halocline:
> Overall looks good! One final question I had related to the Card component specifically: in /templates/Card/Card.js `DesignSystemCard` is exported as `Card`, used with `import { Card } from '../../templates/Card/Card';`. I was wondering if this might cause naming confusion or vagueness between grommet's existing `Card` and other DS-specific Card implementations that build off of grommet's `Card` that accept different or additional properties. However, I realize this naming scheme might be a site-wide nuance, which I know I would require further close inspection. Just an idea 😄

I agree this is a topic we should discuss. Card is an area we have quite of work to do regarding approach and guidance. 

@JCoder121 good awareness and perspective. Thank you for questioning and raising the issue.

halocline (approved):

PR #3371, opened 2023-05-25
Add Data to Learn page and render cards by category

taysea:
[Deploy Preview](https://deploy-preview-3371--keen-mayer-a86c8b.netlify.app/learn)

#### What does this PR do?

Adds "Data" to Learn page and renders cards in separate categories. This will help with organization/findability of content moving forward. At the moment, I have just included a search. As we build out more content we can discuss additional filters that might be desired.

I opted to name the section that's comprised of tutorials as "Getting started" because it felt like the nuanced difference between "Tutorial" vs "How-to" guide might not be clear to an outside audience (received this feedback from Sherry as well). Open to feedback though.

**General UX Checks**
- [ ] Light & dark modes

**Accessibility Checks**
- [ ] Keyboard interactions

**Code Quality Checks**

### DIFF
--- a/aries-site/src/pages/learn/index.js
+++ b/aries-site/src/pages/learn/index.js
@@ -1,9 +1,18 @@
 import React from 'react';
-import { Box, Heading, PageContent, Paragraph } from 'grommet';
-
-import { CardGrid, Meta } from '../../components';
+import {
+  Box,
+  Data,
+  DataContext,
+  DataSearch,
+  DataSummary,
+  Heading,
+  Grid,
+  PageContent,
+  Paragraph,
+} from 'grommet';
+import { Meta, ContentCard } from '../../components';
 import { Layout } from '../../layouts';
-import { getCards, getPageDetails } from '../../utils';
+import { getCards, getPageDetails, nameToPath } from '../../utils';
 
 const title = 'Learn';
 const pageDetails = getPageDetails(title);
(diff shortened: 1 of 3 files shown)


aries-site/src/pages/learn/index.js
```diff
@@ -16,12 +25,73 @@ const Learn = () => (
...
+
+            return (
+              <Box gap="large" pad={{ top: 'medium' }}>
+                {results.map((type, index) =>
+                  type.data?.length ? (
+                    <Box gap="medium" key={index}>
+                      <Heading level={2} margin="none">
+                        {type.heading}
+                      </Heading>
+                      <Grid columns="medium" gap="medium">
```
halocline, line 76: Is [Cards](https://v2.grommet.io/cards) an option here? I haven't played with it much, but curious what your experience with it might be.

aries-site/src/pages/learn/index.js
```diff
@@ -16,12 +25,73 @@ const Learn = () => (
...
     />
-    <PageContent gap="large">
+    <PageContent>
       <Box pad={{ vertical: 'medium' }} justify="center" width="large">
         <Heading margin="none">{title}</Heading>
         <Paragraph size="large">{pageDetails.description}</Paragraph>
       </Box>
-      <CardGrid cards={cards} pad={{ bottom: 'large' }} headingLevel={2} />
+      <Data data={cards} pad={{ bottom: 'large' }}>
+        <DataSearch width={{ max: 'medium', width: '100%' }} />
```
halocline, line 34: Wondering how we might communicate the categories earlier in the page flow? As a reader, I might be interested in knowing these before I start browsing.

`<DataFilter />` might be one option. Another might be exploration of a [toggle-button](https://mui.com/material-ui/react-toggle-button/) which would also solve for the GLCP use case where they are using cards to filter.

Doesn't need to be in this work. Just sharing thoughts.

halocline (approved):
A couple comments to consider, but don't need to get in the way of shipping this as is.

aries-site/src/pages/learn/index.js
taysea: Cards has the DataContext built into it, but since we wanted to split up the results across a few different grids and have a heading for each section, managing it ourselves makes more sense in this case.

aries-site/src/pages/learn/index.js
taysea: Totally agree. Given the size of typography and cards, you can't get a quick sense of what's available. I'll do a separate  design exploration on how to surface that and handle in a different PR once we have an idea of direction

jcfilben (approved):
Looks good!

PR #3574, opened 2023-10-26
Update header.mdx

SeamusLeonardHPE:
Updated link from old component to refactored component in main library file

#### What does this PR do?
Updated link to correct figma file

**General UX Checks**


taysea:
@SeamusLeonardHPE This link goes to Global Header which is a different component than the generic Grommet header. I'm fine with updating for now since the other Header file has been deprecated, just noting because it feels like "Header" page could probably use some updates/revisiting so there's no confusion between global header, app header

PR #3575, opened 2023-10-26
Update index.mdx

SeamusLeonardHPE:
Updated text to clarify rules on closing center layer when it is "actionable"

Check copy to make sure it the rules are more clearly expressed, and that bold has been applied to correct sentence,

### DIFF
--- a/aries-site/src/pages/components/layer/index.mdx
+++ b/aries-site/src/pages/components/layer/index.mdx
@@ -100,11 +100,10 @@ Learn more about the specifics of [informational center layers](/components/laye
 ### Actionable
 
 Actionable layers require the user to either accept or apply the action prompt in the layer or cancel the action.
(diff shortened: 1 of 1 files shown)


aries-site/src/pages/components/layer/index.mdx
```diff
@@ -100,11 +100,10 @@ Learn more about the specifics of [informational center layers](/components/laye
 ### Actionable
 
 Actionable layers require the user to either accept or apply the action prompt in the layer or cancel the action.
```
taysea, line 102:
```suggestion
Actionable layers require the user to either accept or apply the action prompt in the layer or cancel the action.

```

taysea (approved):

taysea:
Removed my accept because actionable layers do have a "X" button to allow user to close, except for the center layer. We have a "Note" under that list item specifically for the center layer use case. Wondering if this change is necessary.

PR #3667, opened 2024-01-30
Update search.mdx

oliverHPE:
Have updated several sections of the search guidance to provide more accurate info.

Information updated includes:

Placeholder text
Icon position
Clear seach icon
Accessibility

The placeholder text on the 'don't' guidance needs to include: "Search first name, last name, email, role, location"and if possible is long enough to show truncation happening.

For the clear search icon, we have two examples we can show. One from Chrome and one from Safari. They're in the Figma file.

Thanks

### DIFF
--- a/aries-site/src/pages/components/search.mdx
+++ b/aries-site/src/pages/components/search.mdx
@@ -70,7 +70,7 @@ All search inputs should specify `type="search"`.
 
 If the search input is within a form then the placeholder should be used instead of a label. If the search input is outside of a form use an`aria-label` in addition to a placeholder.
 
-For additional detail, see [ARIA form hints](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/forms/Basic_form_hints) and ['aria-labelledby'](https://www.w3.org/WAI/tutorials/forms/labels/#using-aria-labelledby).
+For additional detail, see [ARIA form hints](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/forms/Basic_form_hints).
 
 ## Icons
 
--- a/aries-site/src/pages/components/search.mdx
+++ b/aries-site/src/pages/components/search.mdx
@@ -136,7 +136,7 @@ Use simple search when its input field is desired to blend with its surrounding 
 
 ### Icon position
 
-The search icon may be placed either to the right or left of the input field. Icon position is controlled via the [reverse](https://v2.grommet.io/textinput?theme=hpe#reverse) property.
+Icon position is controlled via the [reverse](https://v2.grommet.io/textinput?theme=hpe#reverse) property.
 
 <Example
   code="https://raw.githubusercontent.com/grommet/hpe-design-system/master/aries-site/src/examples/components/search/SearchIconPositionExample.js"
--- a/aries-site/src/pages/components/search.mdx
+++ b/aries-site/src/pages/components/search.mdx
@@ -27,14 +27,14 @@ Many search best practices extend "behind the scenes," such as how query strings
 
 ### Placeholder
 
-Placeholder text should only contain the text "Search".
+Placeholder text should ideally be brief and to the point.
 
 <BestPracticeGroup>
   <Example
     height="small"
     bestPractice={{
       type: 'do',
-      message: 'Search placeholder only contains the text "Search".',
+      message: 'Make the placeholder text concise.',
     }}
   >
     <SearchPlaceholder />
@@ -43,7 +43,7 @@ Placeholder text should only contain the text "Search".
     height="small"
     bestPractice={{
       type: 'dont',
-      message: `Don’t use specific attributes or descriptors in the search placeholder.`,
+      message: `Avoid lengthly placeholder text and avoid truncation.`,
     }}
   >
     <SearchPlaceholder bestPractice={false} />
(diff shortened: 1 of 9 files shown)


aries-site/src/pages/components/search.mdx
```diff
@@ -64,15 +64,26 @@ Placeholder text should only contain the text "Search".
 
 ### Accessibility
 
-Search inputs should be accessible. Ensure users have the ability to return results using their keyboard as well as clicking a button.
+Search inputs should be accessible. Ensure users have the ability to return results using their keyboard or by clicking a button.
 
-Inputs should always be accompanied by labels for accessibility. An icon may serve as a label if:
+All search inputs should specify `type="search"`
```
taysea, line 69:
```suggestion
All search inputs should specify `type="search"`.
```

aries-site/src/pages/components/search.mdx
```diff
@@ -64,15 +64,26 @@ Placeholder text should only contain the text "Search".
...
+Search inputs should be accessible. Ensure users have the ability to return results using their keyboard or by clicking a button.
 
-Inputs should always be accompanied by labels for accessibility. An icon may serve as a label if:
+All search inputs should specify `type="search"`.
 
-1. The icon meaning is well understood.
-2. Has an 'aria-labelledby' attribute.
+If the search input is within a form then the placeholder should be used instead of a label. If the search input is outside of a form use an`aria-label` in addition to a placeholder.
 
 For additional detail, see [ARIA form hints](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/forms/Basic_form_hints) and ['aria-labelledby'](https://www.w3.org/WAI/tutorials/forms/labels/#using-aria-labelledby).
```
jcfilben, line 73:
```suggestion
For additional detail, see [ARIA form hints](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/forms/Basic_form_hints).
```
We should remove the link to aria-labelledby

aries-site/src/pages/components/search.mdx
```diff
@@ -64,15 +64,26 @@ Placeholder text should only contain the text "Search".
...
 
+## Icons
+
+### 'Search' icon placement
+
+The search icon may be placed either to the right or left of the input field. Icon position is controlled via the [reverse](https://v2.grommet.io/textinput?theme=hpe#reverse) property
```
jcfilben, line 79:
```suggestion
The search icon may be placed either to the right or left of the input field. Icon position is controlled via the [reverse](https://v2.grommet.io/textinput?theme=hpe#reverse) property.
```

aries-site/src/pages/components/search.mdx
```diff
@@ -64,15 +64,26 @@ Placeholder text should only contain the text "Search".
...
 
-Search inputs should be accessible. Ensure users have the ability to return results using their keyboard as well as clicking a button.
+Search inputs should be accessible. Ensure users have the ability to return results using their keyboard or by clicking a button.
 
-Inputs should always be accompanied by labels for accessibility. An icon may serve as a label if:
+All search inputs should specify `type="search"`.
 
-1. The icon meaning is well understood.
-2. Has an 'aria-labelledby' attribute.
+If the search input is within a form then the placeholder should be used instead of a label. If the search input is outside of a form use an`aria-label` in addition to a placeholder.
```
taysea, line 71: Can you link to resource of where it's appropriate for placeholder text to serve as replacement for label? This feels off from my understanding. If anything, it feels like we should always suggest an aria-label unless a visual label in a form is provided.

jcfilben: Link to a Figma conversation about this https://www.figma.com/file/KamqhSB2V8TutRsGMAuGgi?node-id=3:1011&mode=design#683671697. 

I think my message might have accidentally been misread, a placeholder can **not** be used instead of a label.

oliverHPE: Thanks for picking up on that Taylor. I must have mis read the comment.

oliverHPE:
> Thanks, Oliver! I've made [this PR](https://github.com/grommet/hpe-design-system/pull/3671/files) with some suggestions to wording/adjusting examples.
> 
> The one thing I wasn't clear on was the rationale for when you would have the "Search" icon at the far right of the input. The rationale for having it before the placeholder text was clear, but feels like we have a gap for the other position. Can you add some content for that?

Okay I see, thanks Taylor. I didn't want to specify whether to use the icon on the left or right as there could be reasons to use either or. Also we don't seem to have a default preference. The point I was trying to get across is that the icon should be seen before the label is seen. Whether it's on the left or right, the icon should be seen first to help with scanning the page. If the icon is seen second, then it's more decoration and secondary support for the label rather than a primary way of locating the search input. I need to adjust the wording to make that more clear if you didn't pick that up. I think the word 'before' here could be the problem, as it seems like it is perhaps suggesting the icon 'before the label' means should be to the left of the label (in left to right reading languages). Saying the icon should be seen first is probably more clear and isn't recommending left or right.

aries-site/src/pages/components/search.mdx
```diff
@@ -64,14 +64,25 @@ Placeholder text should only contain the text "Search".
...
```
oliverHPE, line 71:
```suggestion
Use an `aria-label` in addition to a placeholder unless a visual label is provided.
```

taysea (approved):
LGTM, thanks Oliver for writing the content.

jcfilben (commented):
We talk about icon positioning in both the Search Icon/Placement section and the Variants/Icon position.

I think we should just have this information in one of these sections so we don't duplicate content.

Other than that everything else looks good!

taysea:
> We talk about icon positioning in both the Search Icon/Placement section and the Variants/Icon position.
> 
> I think we should just have this information in one of these sections so we don't duplicate content.
> 
I incorporated this feedback! @jcfilben

jcfilben (approved):
Looks good!

PR #3726, opened 2024-02-15
Update skeleton.mdx

julauxen:
[Deploy Preview](https://deploy-preview--#3726--keen-mayer-a86c8b.netlify.app/)

#### What does this PR do?
Update skeleton guidance.

At guidance.

**General UX Checks**
- [ ] Light & dark modes

**Accessibility Checks**
- [ ] Keyboard interactions

**Code Quality Checks**

### DIFF
--- a/aries-site/src/pages/components/skeleton.mdx
+++ b/aries-site/src/pages/components/skeleton.mdx
@@ -42,4 +42,3 @@ The components that currently support skeleton are:
 - Text, Paragraph, Heading
 - Button
 - Box, Card, Pageheader, Footer, Header
-- DataSearch
--- a/aries-site/src/pages/components/skeleton.mdx
+++ b/aries-site/src/pages/components/skeleton.mdx
@@ -16,7 +16,7 @@ import { DashBoardSkeletonExample, SkeletonAnatomy } from '../../examples';
 </Example>
 
 ### Guidance
-Page skeletons are used in scenarios where the content is being fetched asynchronously and may take some time to load. A skeleton's anatomy is based on the UI element it is rendering and should be responsive.
+Page skeletons are used in scenarios where the content is being fetched asynchronously and may take some time to load. A skeleton's anatomy is based on the UI element it is rendering and should be responsive. If the loading process takes more than a few seconds to load pair the skeleton with a notification to provide the user with better visibility of a system’s status.
 
 ### When to use
 - The interface is content-rich, with numerous elements such as images, text, and other media that take time to load and are known.
@@ -27,7 +27,6 @@ Page skeletons are used in scenarios where the content is being fetched asynchro
 - The data or elements being loaded are dynamic. Use a [Spinner] (/components/spinner) instead.
 - The user is awaiting feedback for in-context actions such as fetching search results. Use a [Spinner] (/components/spinner) instead.
 - The user is awaiting feedback on a submission process, like submitting a form. Use a [Busy Button] (/components/button#busy-button) instead. 
-- The loading process takes more than a few seconds to load. In that case, pair it with a notification to provide the user with better visibility of a system’s status.
 
 ### Anatomy
 
--- a/aries-site/src/pages/components/skeleton.mdx
+++ b/aries-site/src/pages/components/skeleton.mdx
@@ -15,16 +15,19 @@ import { DashBoardSkeletonExample, SkeletonAnatomy } from '../../examples';
   <DashBoardSkeletonExample />
 </Example>
 
+### Guidance
+Page skeletons are used in scenarios where the content is being fetched asynchronously and may take some time to load. A skeleton's anatomy is based on the UI element it is rendering and should be responsive.
+
 ### When to use
-Skeletons are used on all pages to provide a visual layout of the content being loaded.  They help users perceive a loading period as being shorter in duration.
+- The interface is content-rich, with numerous elements such as images, text, and other media that take time to load and are known.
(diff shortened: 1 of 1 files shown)


aries-site/src/pages/components/skeleton.mdx
```diff
@@ -15,16 +15,19 @@ import { DashBoardSkeletonExample, SkeletonAnatomy } from '../../examples';
   <DashBoardSkeletonExample />
 </Example>
 
+### Guidance
```
taysea, line 18: Wondering if we need this section, feels like it is captured by the introduction already:

Screen Shot 2024-02-15 at 3 48 03 PM

aries-site/src/pages/components/skeleton.mdx
```diff
@@ -15,16 +15,19 @@ import { DashBoardSkeletonExample, SkeletonAnatomy } from '../../examples';
...
+- The expected response time is greater than ~300ms.
 
 ### When not to use
-Skeletons should not be used if:
-- The data or elements being loaded are unknown.
-- The user is awaiting feedback for in-context actions such as submitting a form or fetching search results.
-  - The Spinner component may be more appropriate.
-- The loading process takes more than a few minutes to load.
-  - Pair with a message or notification to provide the user with better visibility of a system’s status.
+- The data or elements being loaded are dynamic. Use a [Spinner] (/components/spinner) instead.
```
taysea, line 27: I believe skeleton would also apply when the data is dynamic, so I'd remove this section. I'd think we would always know what ultimately would render in a section of a page, because if the data was there, we'd render it.

aries-site/src/pages/components/skeleton.mdx
```diff
@@ -15,16 +15,19 @@ import { DashBoardSkeletonExample, SkeletonAnatomy } from '../../examples';
...
+- The user is awaiting feedback for in-context actions such as fetching search results. Use a [Spinner] (/components/spinner) instead.
+- The user is awaiting feedback on a submission process, like submitting a form. Use a [Busy Button] (/components/button#busy-button) instead. 
+- The loading process takes more than a few seconds to load. In that case, pair it with a notification to provide the user with better visibility of a system’s status.
```
taysea, line 30: I'm questioning why this is under "When not to use" -- it feels like we're really saying "Use skeleton, and also use a notification". Can we pull this bullet to elsewhere on the guidance?

aries-site/src/pages/components/skeleton.mdx
```diff
@@ -39,3 +42,4 @@ The components that currently support skeleton are:
 - Text, Paragraph, Heading
 - Button
 - Box, Card, Pageheader, Footer, Header
+- DataSearch
```
taysea, line 45: DataSearch component doesn't support skeleton out of the box with grommet at the moment, so let's remove this for now.
```suggestion
```

aries-site/src/pages/components/skeleton.mdx
julauxen: This is a discussion I was having with @KennyAtHPE... He mentioned that the main difference between skeleton and spinner is that skeleton is only for known content and spinner is for unknown. We changed the word to dynamic because unknow was hard to understand. Do you have a better way to put it? or are you saying that this distinction should not be made?

aries-site/src/pages/components/skeleton.mdx
julauxen: Agreed, we can put the in guidance part.

taysea: Given my comment above about removing the "guidance" section, how would you feel about adding this in the "When to use" instead?

aries-site/src/pages/components/skeleton.mdx
taysea: Just for some context on how the pages have been built over time, we haven't quite ironed out the sections that we always want across pages. That being said, Spinner is a bit of an older page so I wouldn't rely on it as the "gold standard" for page structure. Some newer pages are Layer, Card, etc.

I think we realized somewhere along the way that the entire docs site is "guidelines"/"guidance" so having a section called that wasn't super meaningful. I'd still opt for removing this section and just altering the description at the top of the page if we need wording changes.

taysea: I think as a team we should probably revisit what our "gold standard" is for page structure

aries-site/src/pages/components/skeleton.mdx
taysea: I'm more suggesting the distinction should not be made. Since skeleton is aiming to give a scaffold of the page as data is being fetched, I don't think there's really a scenario when elements are unknown. Happy to discuss this further next week on a call if helpful!

aries-site/src/pages/components/skeleton.mdx
julauxen: I am not sure, because this is more of a generic guideline... I don't think it fits right in the "when not to use" or in the "when to use" section. That's why I was a unsure about removing the general guidance section, even though I understand that it is redundant. If we do remove it, where do we put the more generic info?

aries-site/src/pages/components/skeleton.mdx
julauxen: I will schedule some time for us

aries-site/src/pages/components/skeleton.mdx
julauxen: Yes, agreed, Oliver and I are starting to look at that.

taysea:
Closing since skeleton guidance updates were published by https://github.com/grommet/hpe-design-system/pull/3742

Issue #3840, opened 2024-04-12
Draft PR for Selector component


taysea:
Create draft PR for "Selector" component.

[Code prototype](https://codepen.io/Lucas-Speyer/pen/abxERwJ?editors=1100)

Component building checklist: [Grommet Component building checklist.docx](https://hpe.sharepoint.com/:w:/r/teams/hpe-design/Shared%20Documents/Grommet/Grommet%20Dev/Grommet%20Component%20building%20checklist.docx?d=w4dc48049d19148f98de88946fed54955&csf=1&web=1&e=wf9gcY)

### General

**What is the purpose of this component?**

This component aims to streamline the process of making single and multiple selections in scenarios where options should all be exposed on the screen and may contain rich content. Unlike "Select" which seeks to condense the amount of screen real estate, this component is for cases where the controls are intended to have higher visual prominence.

The component enforces:
- Keyboard interactions
- hover/focus states
- selection indication
- container styles

The component does not tightly enforce content, allowing it to be flexible to a variety of use cases.

**Design link?**

Designs: https://www.figma.com/file/5uOrsL2qk8XwwH8C1ZpDb6/branch/QNzyc61prL0zNDyDPHLGz1/HPE-Design-System-Components?type=design&node-id=27691-1164&mode=design&t=bBqv3mOWEbQG0nz4-0

Use cases: https://www.figma.com/file/gEl7V4t4Dd1ZGOZy6EswmQ/Selector-component-%233729?type=design&node-id=1-532&mode=design&t=fzMFObky4yCxaYRy-0

**What are common use cases that Storybook examples should represent? What’s the 80% case?** (Connect with Lucas)

Summary:
- "Quick filtering"
- "Choice" -- making a choice amongst options

Examples:
- **Data experiences:** Users should be able to use the Selector component to efficiently filter and select items in data tables, improving the overall user experience in managing large sets of data. (quick filtering)
- **Onboarding, Wizard:** The Selector component should provide a user-friendly and rapid selection process during onboarding and/or configuration stages, making it easier for users to navigate and choose preferences.
- **Inbox:** Enhance case management usability by allowing users to efficiently filter through numerous cases/items, selecting multiple options at once. (quick filtering)
- **Checkout flows**
- ~**Support and Chat Interactions:** Implement the Selector component to enhance the selection process during support and chat interactions, reducing the time required for users to make choices.~ Removed this use case based on user testing.

### Component naming

**Is there a native HTML element namespace to use?** 

[...]

PR #3865, opened 2024-05-09
Selector template

taysea:
[Deploy Preview](https://deploy-preview-3865--keen-mayer-a86c8b.netlify.app/templates/selector)

#### What does this PR do?

- [x] What should keyboard behavior be? Tab or arrows.
   - Tabs, this component behaves most like a checkboxgroup, so align to native behavior.
- [x] Selector `title` , `description`, `icon` config -- here or on SelectorHeader?
   - On Selector is fine for template, people can expand in their repo if they find a need or we can expand in future.
- [x] If you click on quick filter after having filtered/searched by something, should it clear those other filters/search out?
   - Yes, quick filters essentially act as an exposed saved view.
- [x] Do we need to supportive cases where there's interactive content inside?
   - Not at this time. Let's let use cases drive the conversation before we build to account for it.

2 outstanding decisions we need to make. 

- [x] Wrapping within the Selector. Currently `wrap` is used to allow the titlte/description to wrap under the icon when space is limited. That said, sometimes some Selectors wrap before others in the same group just based on how long their title is. Are we okay with this? Is that misalignment awkward? Should the wrapping be breakpoint driven (that feels a little awkward)? Other approaches?
    - After looking into this, `wrap` isn't the best solution because it wasn't allowing the title to wrap at all. Removing `wrap`. I'm wondering at this point if we need to support the icon on top of title variant at this point?
- [x] Right now the selectors are content driven, so in the grid, one row might grow taller than another just based on the content. Is this okay? If we wanted all rows consistent we'd need to set a fixed row height, but that feels dangerous/might cause content to clip.

Screen image to reflect the above issues

Screen Shot 2024-05-21 at 1 42 01 PM

**General UX Checks**
- [ ] Light & dark modes

**Accessibility Checks**
- [ ] Keyboard interactions

**Code Quality Checks**

Closes https://github.com/grommet/hpe-design-system/issues/3840, https://github.com/grommet/hpe-design-system/issues/3867

#### Notifications

[New]Selector

[]

[]


aries-site/src/examples/components/selector/SupportSelector.js
```diff
@@ -0,0 +1,135 @@
...
+    title: 'Containers as a Service',
+  },
+];
+
+const devices = [
+  {
+    value: 'aruba',
+    title: 'Aruba AP-635 WiFi 6E',
+    description: <Text size="xlarge">18</Text>,
+    icon: <Iteration height="medium" />,
```
jcfilben, line 47:
```suggestion
    icon: <Iteration aria-hidden height="medium" />,
```
Since the icons are decorative lets go ahead and mark them with `aria-hidden` to give a better screenreader experience.

aries-site/src/examples/components/selector/SupportSelector.js
jcfilben, line 1: It feels like this example is right on the line of the maximum number of items we would want to display in this type of format. Like if we have any more than 8 maybe there is a better, more space conscious, component that should be used instead. I think if we publish this page it may be helpful to have a little more guidance on when to use.
Screen Shot 2024-05-20 at 3 21 08 PM

Looks like we need to make some adjustments on the smaller screen size. I'm wondering if it makes sense to use a select/selectmultiple when we switch down to mobile view and there are more than ~3 options.
Screen Shot 2024-05-20 at 3 22 49 PM

jcfilben, line 1: The text wrapping is sometimes a little awkward. I wonder how it would look if the second card's text started on the line right next to the icon (like the first card) but then just took up 2 lines?

Not a huge deal but I'm confused what the '18' is supposed to mean in these examples.
Screen Shot 2024-05-20 at 3 27 17 PM

aries-site/src/examples/components/selector/QuickFilter.js
jcfilben, line 1: This isn't necessarily an issue but I just want to flag it to make sure we are aware of it before we move forward. The green outline is a little similar to our focus indicator. I don't think this is an accessibility issue, but just want to make sure from a styling perspective we are okay with this.

Screen Shot 2024-05-20 at 3 30 27 PM

Screen Shot 2024-05-20 at 3 32 55 PM

aries-core/src/js/components/core/Selector/SelectorGroup.js
```diff
@@ -0,0 +1,131 @@
...
+      setSelectedValue,
+      handleToggle,
+    ],
+  );
+
+  return (
+    <ThemeContext.Extend
+      // theme overrides to remove excess padding
+      value={{
+        checkBox: { pad: 'none' },
```
jcfilben, line 85: It looks like CheckBox has a pad prop, does it work to just set that to 'none'?

aries-site/src/examples/components/selector/QuickFilter.js
jcfilben, line 1: When I use voice over and I read all the elements on the page (ctrl + option + a) as opposed to tabbing through I am getting a weird behavior where the failed launches radio button is read out as '1 of 19' and successful launches is '2 of 19'. It seems like there is an issue in the way the screen reader is interpreting the groups. I would expect it to announce as '1 of 2' or '2 or 2'

aries-site/src/data/structures/components.js
```diff
@@ -1392,6 +1393,46 @@ export const components = [
...
+      'SelectMultiple',
+    ],
+    notification: {
+      message:
+        'Selector is available as a code template that can be copy and pasted into your repository, but it is not published in Grommet.',
+      status: 'info',
+      actions: [
+        {
+          label: 'See source code',
+          icon: <ShareRounded />,
```
jcfilben, line 1425:
```suggestion
          icon: <ShareRounded aria-hidden />,
```

aries-site/src/examples/components/selector/QuickFilter.js
britt6612: This is also how our Secondary Button is do I think its fine but valid call out

aries-site/src/examples/components/selector/SupportSelector.js
britt6612: Definitely should have some sort of guidance for sure based on the number of selectors. 

>> It feels like this example is right on the line of the maximum number of items we would want to display in this type of format. Like if we have any more than 8 maybe there is a better, more space conscious, component that should be used instead. I think if we publish this page it may be helpful to have a little more guidance on when to use.

For now we just moved this to be single column for small and xsmall break points. We can discuss if we want to enhance in future to do a select/selectmultiple

britt6612:
2 outstanding decisions we need to make. 

- [ ] Wrapping within the Selector. Should we add this so that there is columns at certain points vs row or leave as is for now. 
- [ ] Right now the selectors are content driven is this okay? Should we have them all look consistent across? 

Screen image to reflect the above issues

Screen Shot 2024-05-21 at 1 42 01 PM

aries-site/src/examples/components/selector/QuickFilter.js
britt6612: So I heard this as well its actually counting all of the RadioButtons on the whole page not just in that group. Im not sure if this is expected or why its counting all of them.

aries-site/src/examples/components/selector/SupportSelector.js
taysea: Good callout, that's one of our remaining questions about how to handle wrapping in this instance.

Re: "18", it's supposed to represent how many "Aruba AP-..." there are but I can see it's confusing since the data repeats itself. @britt6612 can you update to be unique/more realistic?

aries-site/src/examples/components/selector/QuickFilter.js
taysea: Can someone add a screen recording? I can't replicate this.

aries-site/src/examples/components/selector/SupportSelector.js
britt6612: yes I can!

britt6612: updated

aries-site/src/examples/components/selector/QuickFilter.js
taysea: This is resolved with `aria-hidden="true"` since the radiobutton control is purely decorative.

taysea:
We're getting WAVE violations since the radiobutton/checkbox doesn't have a label. In our case, we really just want the radiobutton / checkbox as a visual but it's not actually performing any form logic. Thoughts on how we might resolve? @jcfilben @britt6612 
 
Screen Shot 2024-05-22 at 2 44 58 PM

jcfilben:
> We're getting WAVE violations since the radiobutton/checkbox doesn't have a label. In our case, we really just want the radiobutton / checkbox as a visual but it's not actually performing any form logic. Thoughts on how we might resolve? Do we need to worry about this since it's not actually interactive? Could we associate it with the title? @jcfilben @britt6612

Instead of using the CheckBox/RadioButton component can we just have an icon?

aries-site/src/examples/components/selector/SupportSelector.js
jcfilben: The small and extra small breakpoints look good now, but it looks like there is still some responsive issues.

Screen Shot 2024-05-22 at 4 14 12 PM

aries-site/src/examples/components/selector/SupportSelector.js
```diff
@@ -0,0 +1,101 @@
...
+            <Heading margin="none">What can we help you with?</Heading>
+            <Text>Choose a product to continue.</Text>
+          </Box>
+          <TextInput
+            aria-label="search"
+            icon={<SearchIcon />}
+            placeholder="Search all assets"
+            type="search"
+            width="medium"
+          />
```
jcfilben, line 73: Can we remove the search here since it is not functional?

taysea: Is it valuable to have if someone were to copy + paste this template?

taysea:
> > We're getting WAVE violations since the radiobutton/checkbox doesn't have a label. In our case, we really just want the radiobutton / checkbox as a visual but it's not actually performing any form logic. Thoughts on how we might resolve? Do we need to worry about this since it's not actually interactive? Could we associate it with the title? @jcfilben @britt6612
> 
> When I run DevTools it is flagging the RadioButton/CheckBox as nested interactive elements. So that is another issue in addition to the label problem.
> 

Hmm .. even with the tabIndex={-1}? I thought that was a viable pattern for this kind of thing.

> I feel like this case isn't necessarily an accessibility issue, but it also doesn't align with the WCAG rules so kinda a grey area.
> 
> Instead of using the CheckBox/RadioButton component can we just have an icon? This might just create a cleaner experience in general. Even with `aria-hidden` set on the RadioButton/CheckBox we still do technically have two nested click targets. We also have this focus behavior coming through on RadioButton if you click directly on it.
> 

With regards to use of an icon, that would have to be a design discussion we bring up with design team. Lucas had two design versions of this component (one with just an icon, one with the RadioButton/Checkbox) and had settled on the RadioButton/CheckBox I believe as a result of some user testing, so we should better understand if the icon path might still be an option.


> Screen Shot 2024-05-22 at 4 20 35 PM

Also wasn't expected this to be able to be active element with tabIndex, but realizing that doesn't necessarily block "click". That said, we do a similar pattern on DS site for our preview cards which is why I thought it'd work.

britt6612:
> > > We're getting WAVE violations since the radiobutton/checkbox doesn't have a label. In our case, we really just want the radiobutton / checkbox as a visual but it's not actually performing any form logic. Thoughts on how we might resolve? Do we need to worry about this since it's not actually interactive? Could we associate it with the title? @jcfilben @britt6612
> > 
> > 
> > When I run DevTools it is flagging the RadioButton/CheckBox as nested interactive elements. So that is another issue in addition to the label problem.
> 
> Hmm .. even with the tabIndex={-1}? I thought that was a viable pattern for this kind of thing.
> 
> > I feel like this case isn't necessarily an accessibility issue, but it also doesn't align with the WCAG rules so kinda a grey area.
> > Instead of using the CheckBox/RadioButton component can we just have an icon? This might just create a cleaner experience in general. Even with `aria-hidden` set on the RadioButton/CheckBox we still do technically have two nested click targets. We also have this focus behavior coming through on RadioButton if you click directly on it.
> 
> With regards to use of an icon, that would have to be a design discussion we bring up with design team. Lucas had two design versions of this component (one with just an icon, one with the RadioButton/Checkbox) and had settled on the RadioButton/CheckBox I believe as a result of some user testing, so we should better understand if the icon path might still be an option.
> 
> > Screen Shot 2024-05-22 at 4 20 35 PM
> 
> Also wasn't expected this to be able to be active element with tabIndex, but realizing that doesn't necessarily block "click". That said, we do a similar pattern on DS site for our preview cards which is why I thought it'd work.

I almost wonder if we can just make an icon look like our radiobutton and checkbox then use that if that is what is wanted in the design.

jcfilben (commented):
Can we round the Checkbox icon container so that it matches the hpe theme styling for the CheckBox component?

britt6612:
> Can we round the Checkbox icon container so that it matches the hpe theme styling for the CheckBox component?

done

aries-core/src/js/components/core/Selector/SelectorGroup.js
```diff
@@ -0,0 +1,142 @@
...
+          hover: {
+            background: 'none',
+          },
+          container: {
+            extend: '',
+          },
+          extend: '',
+        },
+      }}
+    >
```
jcfilben, line 103: Can we remove this theme override since we are using icons now?

britt6612: good call out yes!

jcfilben:
> Wrapping within the Selector. Currently wrap is used to allow the titlte/description to wrap under the icon when space is limited. That said, sometimes some Selectors wrap before others in the same group just based on how long their title is. Are we okay with this? Is that misalignment awkward? Should the wrapping be breakpoint driven (that feels a little awkward)? Other approaches?
> After looking into this, wrap isn't the best solution because it wasn't allowing the title to wrap at all. Removing wrap. I'm wondering at this point if we need to support the icon on top of title variant at this point?
> 

I don't feel like we need to support the icon on top variant

> Right now the selectors are content driven, so in the grid, one row might grow taller than another just based on the content. Is this okay? If we wanted all rows consistent we'd need to set a fixed row height, but that feels dangerous/might cause content to clip.

I think it is fine if one row is taller than another row

britt6612:
Just an update on the changes that Matt added:

SelectorGroup: 
```
  children,
  defaultValue,
  layout = 'grid',
  multiple,
  onSelect,
  value,
```

The new change here is the `layout` this is a new added prop and it takes either `grid` `fit`. When the `grid` is used then the `columns` are set to `count: 'fit', size: ['small', 'flex']` while the `fit` is wrapped in a Box with `direction='row'`

Selector:

The changes here are the following condition was added 
```
      {(indicator !== false || title || description || icon) && (
          <SelectorHeader
            indicator={indicator}
            title={title}
            icon={icon}
            description={description}
            selected={selected}
          />
        )}
```

SelectorHeader:

Here the new addition was you can pass an object to SelectorIndicator for example 
`indicator={{ round: 'full', size: 'medium' }}`
so when `multiple` if someone wanted their indicator to be round that can happen. 

Matt left a comment if there should always be a `checkmark` which him Jessica and I did discuss however we need to run by Lucas in my opinion to make sure this is okay on the design side as well.

jcfilben:
> Here the new addition was you can pass an object to SelectorIndicator for example
indicator={{ round: 'full', size: 'medium' }}
so when multiple if someone wanted their indicator to be round that can happen.

I'm confused why we added this. I was thinking having the square vs circle was included so that people could have a visual indicator of whether the selector is a single select or multiple. Is there a use case we are envisioning for passing an object to `indicator` to override this behavior?

aries-core/src/js/components/core/Selector/SelectorHeader.js
```diff
@@ -0,0 +1,82 @@
...
+      height={theme.global.edgeSize[size]}
+      width={theme.global.edgeSize[size]}
+      justify="center"
+      align="center"
+      background={selected ? 'brand' : undefined}
+      flex={false}
+      {...rest}
+    >
+      {selected && (
+        // need to talk with Lucas
```
jcfilben, line 29: I'm not opposed to this, I like how it looks visually, but it seems like we lose a little of the styling similarities between radio button and single selection. This change might not make it is as clear for the user that the circle is a single select.

I agree it would be good to check with Lucas

aries-site/src/pages/components/selector.mdx
```diff
@@ -0,0 +1,78 @@
...
+import { ShareRounded } from 'grommet-icons';
+import { Example } from '../../layouts';
+import {
+  SimpleSelector,
+  QuickFilter,
+  SupportSelector,
+  TEMP,
+} from '../../examples';
+
+{/* Let's move this under templates instead of components */}
```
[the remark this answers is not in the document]
britt6612: oh shoot need to do this thanks

aries-core/src/js/components/core/Selector/SelectorHeader.js
britt6612: yes Im planning tomorrow to discuss

britt6612:
> > Here the new addition was you can pass an object to SelectorIndicator for example
> > indicator={{ round: 'full', size: 'medium' }}
> > so when multiple if someone wanted their indicator to be round that can happen.
> 
> I'm confused why we added this. I was thinking having the square vs circle was included so that people could have a visual indicator of whether the selector is a single select or multiple. Is there a use case we are envisioning for passing an object to `indicator` to override this behavior?

good question, I also don't have clarity when reviewing I figure it was just Matts way of making things more flexible. I wonder if he saw another use case with clicking multiple but then had a round indicator.

aries-site/src/pages/components/selector.mdx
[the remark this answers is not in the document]
britt6612: actually @taysea didnt we say we wanted this under components? Thats why we were adding the notification?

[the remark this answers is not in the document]
taysea: Matt decided it would be better to have it in Templates for now. Let's go ahead with moving it there for now.

taysea:
> > > Here the new addition was you can pass an object to SelectorIndicator for example
> > > indicator={{ round: 'full', size: 'medium' }}
> > > so when multiple if someone wanted their indicator to be round that can happen.
> > 
> > 
> > I'm confused why we added this. I was thinking having the square vs circle was included so that people could have a visual indicator of whether the selector is a single select or multiple. Is there a use case we are envisioning for passing an object to `indicator` to override this behavior?
> 
> good question, I also don't have clarity when reviewing I figure it was just Matts way of making things more flexible. I wonder if he saw another use case with clicking multiple but then had a round indicator.

@jcfilben @britt6612 I believe this was to allow something a bit more flexible 

> > > Here the new addition was you can pass an object to SelectorIndicator for example
> > > indicator={{ round: 'full', size: 'medium' }}
> > > so when multiple if someone wanted their indicator to be round that can happen.
> > 
> > 
> > I'm confused why we added this. I was thinking having the square vs circle was included so that people could have a visual indicator of whether the selector is a single select or multiple. Is there a use case we are envisioning for passing an object to `indicator` to override this behavior?
> 
> good question, I also don't have clarity when reviewing I figure it was just Matts way of making things more flexible. I wonder if he saw another use case with clicking multiple but then had a round indicator.

@jcfilben @britt6612 I believe this was to give a bit more flexibility to support something like Gina's "choice chips" design if desired.

aries-site/src/pages/components/selector.mdx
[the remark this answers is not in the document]
britt6612: okay was there any reasoning behind that or?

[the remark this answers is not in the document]
taysea: His reasoning was to keep "components" limited to things that truly have a published Grommet component. If we lean into that, we should move "Search" to templates as well (currently under components)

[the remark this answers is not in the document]
britt6612: ☝️ thank you that was my next question was should we move "Search"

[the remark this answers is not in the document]
britt6612: moved to templates

aries-site/src/examples/cardPreviews/selector.js
jcfilben, line 1: Can we add a little gap between the selector elements on the Card preview? They just look a little squished
Screen Shot 2024-06-18 at 12 21 48 PM

jcfilben (commented):
Just one more comment, otherwise looks good

aries-site/src/examples/cardPreviews/selector.js
britt6612: good catch

britt6612: fixed in last commit

jcfilben (approved):
Looks good!

aries-site/src/data/structures/templates/templates.js
```diff
@@ -222,6 +223,29 @@ export const templates = [
       'sorting criteria',
     ],
   },
+  {
+    name: 'Selector',
+    available: true,
+    cardOrder: 4,
+    description: `A selection control that allows users to choose one or
+    more related options to access information, filter, make selections
+    or take actions.`,
```
julauxen, line 232: I would not say "take actions" because we had a lot of confusion in terms of that functionality. The selector is not a button, therefore not a place to take action.

aries-site/src/pages/templates/selector.mdx
```diff
@@ -0,0 +1,93 @@
...
+  width="100%"
+>
+  <SimpleSelector />
+</Example>
+
+## Use cases
+
+### Quick Filtering
+
+Enables users to filter data based on high-level visual options to narrow down datasets within tables and list. Making it quickly recognizable, based on specific criteria, enhancing efficiency and usability.
```
julauxen, line 28: "Enables users to filter data based on high-level visual options to narrow down datasets within tables and list."

All the rest is not really adding clarity to the use case.

aries-site/src/pages/templates/selector.mdx
```diff
@@ -0,0 +1,93 @@
...
+  ]}
+  github="https://github.com/grommet/hpe-design-system/tree/master/aries-core/src/js/components/core/Selector"
+  width="100%"
+>
+  <QuickFilter />
+</Example>
+
+### Quick option selection
+
+Helps users quickly navigate through choices, ensuring they find the relevant options with ease.
```
julauxen, line 45: "Enables users to quickly navigate through choices, ensuring they find the relevant options with ease."

Just changed a bit to keep the format of text of previous use case.

aries-site/src/pages/templates/selector.mdx
```diff
@@ -0,0 +1,93 @@
...
+  code={[
+    'https://raw.githubusercontent.com/grommet/hpe-design-system/master/aries-site/src/examples/components/selector/SupportSelector.js',
+  ]}
+  github="https://raw.githubusercontent.com/grommet/hpe-design-system/master/aries-site/src/examples/components/selector/SupportSelector.js"
+>
+  <SupportSelector />
+</Example>
+
+## Anatomy
+
```
julauxen, line 60: There's no imagine for anatomy? It's hard to understand without a visual

aries-site/src/pages/templates/selector.mdx
```diff
@@ -0,0 +1,93 @@
...
+
+| Label | Region                  | Purpose                                                      | Required | Notes                                                   |
+| :---: | --------------------    | ------------------------------------------------------------ | :------: | ------------------------------------------------------- |
+| **1** | **Title**               | Heading providing context to the options available.          | Optional | Titles should use a Heading of the correct semantic heading level based on where the selector is in the document. |
```
julauxen, line 64: Is the title optional?

I ask because it's weird to me reading the guidance that everything in a selector is optional.

aries-site/src/pages/templates/selector.mdx
```diff
@@ -0,0 +1,93 @@
...
+| **2** | **Description**         | Brief explanation or additional information about the choices.| Optional |    --                                                  |
+| **3** | **Counter of Items**    | Displays number of items available for selection.            | Optional | -- |
+| **4** | **Icon**                | Visual aid to represent the category or type of selection.   | Optional | a11yTitle required. |
+| **5** | **Selector Indicator**  | Visual cue (such as a checkmark or highlight) to denote selected items. | Optional | aria-label required. |
+
+## Types
+
```
julauxen, line 71: I would suggest we use a different visual representation for "Types", like a table with a visual and an explanation, so the user can easily understand the difference not only on functionality but on type of component.

Also, if we are delivering all the examples on the TEMP kitchen sink, we might need to amplify this section, since we have different visual types for single select and multi select.

aries-site/src/pages/templates/selector.mdx
vavalos5: Totally get it. We have an image in Figma and something that's still a bit more TBD. Imagery will come later.

aries-site/src/pages/templates/selector.mdx
vavalos5: @britt6612 I'll let you input here

aries-site/src/pages/templates/selector.mdx
vavalos5: I'm not sure I'm following. For types we don't include imagery but rather indicate the "types" of selectors there are.

For kitchen sink, these are all TBD and will be clarified soon. :)

britt6612: Kitchen sink is just for us devs to test will not be published

aries-site/src/pages/templates/selector.mdx
britt6612: anatomy to come..

britt6612: have to code it up

aries-site/src/pages/templates/selector.mdx
britt6612: So essentially the way Selector is built in code is that a person does not necessarily have to provide a title..

Screen Shot 2024-07-02 at 11 31 14 AM


for example above the user is passing a text within `children` which has a different placement for the title. We built selector in code to be very flexible so we are not making anything `required` because the user can really pass a variety of things into the Selector

aries-site/src/pages/templates/selector.mdx
britt6612: @julauxen Im also not sure I follow your visual with "types"

aries-site/src/data/structures/templates/templates.js
britt6612: @vavalos5 lets work on re-wording

aries-site/src/pages/templates/selector.mdx
britt6612, line 28:
```suggestion
Enables users to filter data based on high-level visual options to narrow down datasets within tables and lists.
```

aries-site/src/pages/templates/selector.mdx
```diff
@@ -0,0 +1,93 @@
...
+  ]}
+  github="https://github.com/grommet/hpe-design-system/tree/master/aries-core/src/js/components/core/Selector"
+  width="100%"
+>
+  <QuickFilter />
+</Example>
+
+### Quick option selection
+
+Helps users quickly navigate through choices, ensuring they find the relevant options with ease.
```
britt6612, line 45:
```suggestion
Enables users to quickly navigate through choices, ensuring they find the relevant options with ease.
```

aries-site/src/pages/templates/selector.mdx
```diff
@@ -0,0 +1,93 @@
...
+  ]}
+  github="https://github.com/grommet/hpe-design-system/tree/master/aries-core/src/js/components/core/Selector"
+  width="100%"
+>
+  <SimpleSelector />
+</Example>
+
+## Use cases
+
+### Quick Filtering
```
taysea, line 26:
```suggestion
### Quick filtering
```
Sentence case

aries-site/src/pages/templates/selector.mdx
```diff
@@ -0,0 +1,93 @@
...
+
+Enables users to filter data based on high-level visual options to narrow down datasets within tables and lists.
```
taysea, line 28: It feels like this should capture:
- Provides an efficient way to filter data to a specific view that will be useful and relevant to the workflow.

Maybe not the best wording, but given its high visual prominence, they should be useful and relevant.

It feels like the current guidance is redefining "Selector" in some ways (high prominence visual options) / we can clean up wording since "to narrow down datasets" is redundant with "filter data".

aries-site/src/pages/templates/selector.mdx
```diff
@@ -0,0 +1,93 @@
...
+  ]}
+  github="https://github.com/grommet/hpe-design-system/tree/master/aries-core/src/js/components/core/Selector"
+  width="100%"
+>
+  <QuickFilter />
+</Example>
+
+### Quick option selection
+
+Enables users to quickly navigate through choices, ensuring they find the relevant options with ease.
```
taysea, line 45: Since this is "use cases" which should be a bit specific. Can we add something that says this is typically found in support and onboarding flows?

aries-site/src/pages/templates/selector.mdx
```diff
@@ -0,0 +1,93 @@
...
+| **2** | **Description**         | Brief explanation or additional information about the choices.| Optional |    --                                                  |
+| **3** | **Counter of Items**    | Displays number of items available for selection.            | Optional | -- |
+| **4** | **Icon**                | Visual aid to represent the category or type of selection.   | Optional | a11yTitle required. |
+| **5** | **Selector Indicator**  | Visual cue (such as a checkmark or highlight) to denote selected items. | Optional | aria-label required. |
+
+## Types
+
+### Single select
+
+Users can select only one option at a time. Ideal for scenarios where only one choice is allowed.
```
taysea, line 74: Let's reuse similar wording from ToggleGroup:

With single selection, only one option can be selected at a time. This is the default behavior.

aries-site/src/pages/templates/selector.mdx
```diff
@@ -0,0 +1,93 @@
...
+
+### Multiple select
+
+Users can select multiple options simultaneously. Useful for filtering or multi-choice settings.
```
taysea, line 78:
```suggestion
With multiple selection, multiple options can be selected at the same time.
```

aries-site/src/pages/templates/selector.mdx
britt6612: okay I re-wording lmk what you think

aries-site/src/pages/templates/selector.mdx
britt6612: done

aries-site/src/pages/templates/selector.mdx
britt6612: done

aries-site/src/pages/templates/selector.mdx
```diff
@@ -0,0 +1,98 @@
...
+| :----: | ----------------------- | -------------------------------------------------------------------------------------------- | :------: | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
+| **1**  | **Container**           | Provides distinct boundary representing the option as a distinct choice.                     | Required | --                                                                                                                                                                       |
+| **2**  | **Selection Indicator** | Provides additional visual prominence and distinction between selected and unselected state. | Optional | For single it will be rounded for multiple sqaured off                                                                                                                   |
```
vavalos5, line 68: for the notes, spell check: multiple squared off

Issue #4070, opened 2024-07-30
Add `primary` to dataCell


taysea:
Missing tokens for `primary` column of dataCells

PR #4074, opened 2024-07-31
Add primary datacell tokens

jcfilben:

#### What does this PR do?
Adds tokens for primary datacell

**General UX Checks**
- [ ] Light & dark modes

**Accessibility Checks**
- [ ] Keyboard interactions

**Code Quality Checks**

Closes https://github.com/grommet/hpe-design-system/issues/4070

### DIFF
--- a/design-tokens/tokens/component.default.json
+++ b/design-tokens/tokens/component.default.json
@@ -6187,7 +6187,7 @@
     },
     "fontWeight": {
       "$type": "number",
-      "$value": "{base.fontWeight.medium}",
+      "$value": "{base.fontWeight.regular}",
       "$description": "",
       "$extensions": {
         "com.figma": {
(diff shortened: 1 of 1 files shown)


design-tokens/tokens/component.default.json
```diff
@@ -5858,6 +5858,291 @@
     }
   },
   "dataCell": {
+    "primary": {
+      "paddingX": {
+        "$type": "number",
+        "$value": "{base.static.spacing.small}",
```
taysea, line 5864: To make it easier for things to stay in sync in future, can we have these all reference `{dataCell.paddingX}` or whatever the property is instead?

Then just for the ones that deviate (like fontWeight) have it be its own value?

jcfilben: Yeah I'll make that change. Also I think the `dataCell.fontWeight` and `dataCell.enabled.fontColor` are wrong. I wasn't sure if I should touch those in this PR though since they don't fall under the scope of `dataCell.primary`

taysea: Thanks for flagging, feel free to fix them thank you!

jcfilben: Fixed

taysea (approved):

PR #4483, opened 2024-11-26
Use container tokens in grommet-app

taysea:

#### What does this PR do?
Updates `theme.jsx` in sandbox/grommet-app to use "size.container" tokens instead if they exist.

Still need to check for size.content tokens since that demo app toggles between older token branches. In future, we can clean this up so it only uses current tokens.

**General UX Checks**
- [ ] Light & dark modes

**Accessibility Checks**
- [ ] Keyboard interactions

**Code Quality Checks**

### DIFF
--- a/sandbox/grommet-app/src/theme.jsx
+++ b/sandbox/grommet-app/src/theme.jsx
@@ -380,6 +380,32 @@ const buildTheme = tokens => {
     },
   };
 
+  const containerTokens = 'container' in large.hpe.size;
+  const size = breakpoint => ({
+    xxsmall: containerTokens
+      ? breakpoint.hpe.size.container.xxsmall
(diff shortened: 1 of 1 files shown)


sandbox/grommet-app/src/theme.jsx
```diff
@@ -380,6 +380,32 @@ const buildTheme = tokens => {
     },
   };
 
+  const containerTokens = 'container' in large.hpe.size;
```
halocline, line 383: This is for backwards compatibility in case the grommet-app is using an earlier version of tokens? Am I understanding correctly?

taysea: Yes exactly.

halocline (approved):

PR #4635, opened 2024-12-20
Add docs for typography system

taysea:
https://docs-typography-system--keen-mayer-a86c8b.netlify.app/design-tokens

**General UX Checks**
- [ ] Light & dark modes

**Accessibility Checks**
- [ ] Keyboard interactions

**Code Quality Checks**


jcfilben (commented):
When I search for typography, it feels a little unclear which page to go to. Could see users being confused why we have two pages for this. Need a way to clarify that one is for tokens and one is for grommet
Screenshot 2025-02-11 at 2 37 41 PM

taysea:
> When I search for typography, it feels a little unclear which page to go to. Could see users being confused why we have two pages for this. Need a way to clarify that one is for tokens and one is for grommet Screenshot 2025-02-11 at 2 37 41 PM

It's a good point -- I had the same thought when writing these docs. I almost wonder if the two pages should be consolidated

jcfilben:
> > When I search for typography, it feels a little unclear which page to go to. Could see users being confused why we have two pages for this. Need a way to clarify that one is for tokens and one is for grommet Screenshot 2025-02-11 at 2 37 41 PM
> 
> It's a good point -- I had the same thought when writing these docs. I almost wonder if the two pages should be consolidated

I feel like it would be nice to just have one page, but maybe it is a wider site architecture discussion. Currently it seems like the design tokens are in a separate area of the site. Do we want to generally keep things separate or try to blend them together. I wonder how other systems handle this that have design tokens

halocline (approved):

PR #4640, opened 2025-01-06
Add error check for invalid Figma variables references

### LINKED ISSUE #4655: Resolve incorrect modes coming back into Figma files

taysea:

#### What does this PR do?
This PR adjusts our scripts to throw errors when Figma variables make an invalid reference to an old file. This should reduce unexpected changes where Figma variables were referencing old files without our knowledge.

Now, anytime we attempt to push from Github to Figma OR pull from Figma to Github, these scripts will throw an error with any tokens that need to be fixed. The fix should occur within Figma (because this is where the reference IDs are established) and then pulled into Github.

The desired collection IDs are stored as Secrets provided to the Github Action, which the script uses to test for valid/invalid references.

Locally, tested this and originally saw 4 errors (see first screenshot). Then, resolved a reference in Figma and re-ran the script. Correctly, now there were only 3 errors (see second screenshot).

Screenshot 2025-01-06 at 2 49 29 PM

Screenshot 2025-01-06 at 2 50 00 PM

**General UX Checks**
- [ ] Light & dark modes

**Accessibility Checks**
- [ ] Keyboard interactions

**Code Quality Checks**

Closes #4655


design-tokens/src/utils.ts
```diff
@@ -49,3 +51,68 @@ export const nonComponentTokens: string[] = [
...
+      const variable = tokens.meta.variables[key];
+      Object.keys(variable.valuesByMode).forEach(j => {
+        const modeValue = variable.valuesByMode[j];
+        if (
+          typeof modeValue === 'object' &&
+          'id' in modeValue &&
+          invalidVariables.includes(modeValue.id)
+        ) {
+          console.error(
+            `🛑 Invalid collection reference for value of: ${variable.name}`,
```
halocline, line 108: The error message screen shot shown in the PR description reads `Invalid {collection.name} collection ...`, but I don't see that reflected here. I think the collection name context is helpful. Intended? Am I missing it?

halocline: Is there resolution detail we can provide accompanying this error message? e.g. add a corresponding console.log which states `${variable.name} is referencing an incorrect Figma file. Expecting to reference value from ${process.env[collectionKey]}`.

Would something like this be helpful? If I am a dev encountering this error and unfamiliar with the set up, additional specificity helping me towards resolution would be useful.

In my suggested messaging, its recommending specifying the Figma file. Would we be leaking anything sensitive in doing so?

halocline (commented):
Looks good. A couple comments related to messaging to potentially enhance dev experience and resolution.

design-tokens/src/utils.ts
taysea: The "color" reference in the screenshot was from an earlier stage of the PR where I only was checking color. Once adding other collections, I decided the inclusion of "color"/"dimension" etc. isn't adding value given you really just need the name of the token itself (and would've require more logic to include anyway so I removed).

Addressed your comment with an update to the error message. Having the collection key wouldn't necessarily be helpful, but I've added clarification that the error should be resolved in Figma so a dev with less context would know what to do.

halocline (approved):

PR #4785, opened 2025-02-13
All tokens - Nav menu tweaks

halocline:
[Deploy Preview](https://deploy-preview-4785--keen-mayer-a86c8b.netlify.app/design-tokens/all-design-tokens)

#### What does this PR do?

- Fixed primitives collection bug where 'base' was being referenced, but we renamed to primitives
- Some layout tweaks
  - Only occupy needed width
- Some nav menu styling tweaks
  - Trying to find styling which doesn't require borders (alternate is no background with single border on right)
- Changed nav menu behavior to only have one directory open at a time. This was to address the ability to support sticky menu, while avoiding menu item overflow / the need for an overflow scroll. Ensures the menu is always present without need for excessive div height to prevent overflow.
- Close Nav layer when menu item is selected.

**General UX Checks**
- [ ] Light & dark modes

**Accessibility Checks**
- [ ] Keyboard interactions

**Code Quality Checks**

### DIFF
--- a/aries-site/src/components/content/DesignTokensTable.js
+++ b/aries-site/src/components/content/DesignTokensTable.js
@@ -24,6 +24,70 @@ import {
 } from './designTokenUtils';
 import { DesignTokenContext } from './DesignTokenContext';
 
+const tokenColumns = [
+  {
+    property: 'id',
+    header: 'Preview',
+    render: datum => {
+      if (datum.type === 'color') return <ColorPreview datum={datum} />;
+
+      if (
+        datum.token.includes('size') ||
+        datum.token.includes('dimension') ||
+        datum.token.includes('spacing') ||
+        datum.token.includes('gap')
+      )
+        return <DimensionPreview datum={datum} />;
+      if (
+        datum.token.includes('radius') ||
+        datum.token.includes('borderRadius') ||
+        /border.*Radius/.test(datum.token)
+      )
+        return <RadiusPreview datum={datum} />;
+      if (datum.token.includes('border'))
+        return <BorderPreview datum={datum} />;
+      if (datum.token.includes('weight') || datum.token.includes('fontWeight'))
+        return <WeightPreview datum={datum} />;
+      if (
+        (datum.token.includes('font') && datum.token.includes('size')) ||
+        datum.token.includes('fontSize')
+      )
+        return <TextPreview datum={datum} />;
+      return '--';
+    },
+    size: 'auto',
+    pin: true,
+  },
+  {
+    property: 'token',
+    header: 'Token',
+    render: datum => (
+      <Box direction="row">
+        <Box
+          background="background-contrast"
+          pad="xsmall"
+          round="xsmall"
+          style={{ fontFamily: 'Menlo' }}
+        >
+          <Text size="xsmall">{datum.token}</Text>
+        </Box>
+      </Box>
+    ),
+  },
+  {
+    property: 'description',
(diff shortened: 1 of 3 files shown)


britt6612:
In the `layer` version on mobile when I open the Layer and click on another option for example `semantic color` I have to then click the exit button for the layer to close. I would expect once I click `semantic color` it would close and this would be the current state
https://github.com/user-attachments/assets/4f017a62-de4b-4d91-8619-d0898153ab14

aries-site/src/pages/design-tokens/all-design-tokens.js
```diff
@@ -64,10 +67,48 @@ const Nav = ({ active, setActive, tokens: tokensObj }) => {
...
+const NavPane = ({ children }) => {
+  const theme = useContext(ThemeContext);
+  return (
+    <Box
+      flex={false}
+      style={{
+        position: 'sticky',
+        top: theme.global.edgeSize.medium,
+      }}
+      height="130vh"
```
britt6612, line 79: I see why you added this `height` but also see an issue this is causing 


https://github.com/user-attachments/assets/c927ab17-83f5-413f-bf9f-3dad70d944a8

halocline: bummer. I thought I had addressed that. Lost it along the way. Will fix.

aries-site/src/components/content/DesignTokensTable.js
```diff
@@ -50,29 +114,31 @@ export const DesignTokensTable = ({ active, maxHeight, toolbar }) => {
             <DataSearch />
             {modeOptions?.length > 1 ? (
               // eslint-disable-next-line grommet/formfield-htmlfor-id
```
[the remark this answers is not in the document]
halocline: Not something I touched, but will remove.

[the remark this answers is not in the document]
halocline: removed

aries-site/src/pages/design-tokens/all-design-tokens.js
halocline: fixed.

aries-site/src/components/content/designTokenUtils.js
```diff
@@ -5,20 +5,20 @@ import * as tokens from 'hpe-design-tokens/docs';
...
-  // base, component, light, dark, etc.
+  // primitives, component, light, dark, etc.
   Object.keys(tokens[mode]).forEach(token => {
     const currentToken = tokens[mode][token];
 
     const parts = token.split('.');
     const category = parts[1];
     let level;
-    if (mode === 'base') level = 'primitive';
+    if (mode === 'primitives') level = 'primitives';
```
taysea, line 21: Thank you!

taysea:
I'm supportive of the adjustment to use "background-contrast" for the side nav.

The narrowness of the side navigation though feels a bit odd to me. I'm up for reducing it but what if we use the in-between off our new scale "256px"?

Personally, I liked the ability to open multiple directories at once to be able to get context across the different groups. I don't feel like there's anything wrong with an overflow happening as a result of me opening multiple folders.

Also the implementation seems to still have overflow issues:


https://github.com/user-attachments/assets/705d5ab9-a130-4995-bb63-38f912b81f42

Screenshot 2025-02-21 at 3 48 43 PM

taysea (approved):
I'm good with these changes.

PR #4840, opened 2025-03-06
A11y - Added eval and content for Card

halocline:
[Deploy Preview](https://deploy-preview-4840--keen-mayer-a86c8b.netlify.app/components/card?q=card#wcag-compliance)

#### What does this PR do?

Addresses evaluation and site content for: https://github.com/grommet/hpe-design-system/issues/4810

Additionally:
- Introduces 'not-applicable' status to fully account for complete evaluation record.
- Modifies logic generating WCAG compliance summary and detail content to account for:
  - 'not-applicable' status
  - WCAG grouping where 0 rules are applicable

#### Notifications
[Update]Card

[Accessibility]

### DIFF
--- a/aries-site/src/data/wcag/components.json
+++ b/aries-site/src/data/wcag/components.json
@@ -228,8 +228,8 @@
       {
         "rule": "2.3.3",
         "label": "Animation from Interactions",
-        "status": "not-applicable",
-        "notes": "I don't think hover elevation is considered here. Check with team."
+        "status": "failed",
+        "notes": "Due to the the intent of 2.3.3 and subtle nature of the elevation animation, applicability of this rule falls into a grey area as the subtlty of the elevation is not likely to trigger reactions this rule intends to prevent. That said, we can treat as 'does not meet AAA' until `prefer-reduced-motion` is supplied in Grommet or the HPE theme. 2.3.3 aims to prevent animations that could cause seizures or physical reactions, particularly for users with vestibular disorders or cognitive disabilities. The core concern is with animations that are unexpected or excessive. Adding elevation to a card on hover typically involves a subtle animation, such as a slight increase in size or a shadow effect. Whether this triggers WCAG 2.3.3 depends on the nature of the animation. Applicability - 10 Subtle and Short Animations: If the elevation animation is subtle, brief, and not overly dynamic, it's less likely to cause issues. Many standard hover effects fall into this category. 2) Excessive or Rapid Animations: If the animation involves rapid, flashing, or large-scale movements, it could potentially violate WCAG 2.3.3. 3) User Control: A key aspect of WCAG 2.3.3 is providing users with the ability to disable non-essential animations. Therefore, if your site includes such animations, it should ideally offer a mechanism to turn them off and make use of `prefers-reduced-motion` media query."
       },
       {
         "rule": "2.4.1",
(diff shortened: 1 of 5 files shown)


aries-site/src/data/wcag/components.json
```diff
@@ -48,5 +48,419 @@
...
+      {
+        "rule": "2.3.2",
+        "label": "Three Flashes",
+        "status": "not-applicable"
+      },
+      {
+        "rule": "2.3.3",
+        "label": "Animation from Interactions",
+        "status": "not-applicable",
+        "notes": "I don't think hover elevation is considered here. Check with team."
```
jcfilben, line 232: Kinda on the fence about this, agree it would be good to chat as a team. Maybe this is something we could bring up in our call Friday. Did a quick google search about if a hover elevation is considered an animation and here is what I got back.
Screenshot 2025-03-06 at 1 02 12 PM

We don't do anything in grommet with 'prefers-reduced-motion', feel like this is an area across grommet that we could improve upon

aries-site/src/data/wcag/components.json
```diff
@@ -48,5 +48,419 @@
...
+        "rule": "2.4.3",
+        "label": "Focus Order",
+        "status": "conditional",
+        "notes": "Depending on context, a proper role of `listitem` or wrapping in an <li> may be needed."
+      },
+      {
+        "rule": "2.4.4",
+        "label": "Link Purpose (In Context)",
+        "status": "conditional",
+        "notes": "Related cards sharing a common call to action such as 'Learn More' may need to be differentiated by using `aria-label` to provide context as to answer 'learn more about what?'."
```
jcfilben, line 254: Good note, feel like this is something we could be better at demonstrating in our card examples

aries-site/src/data/wcag/components.json
```diff
@@ -48,5 +48,419 @@
...
+      {
+        "rule": "2.4.7",
+        "label": "Focus Visible",
+        "status": "passed",
+        "notes": "When the card is interactive, focus is visible on hover and keyboard focus."
+      },
+      {
+        "rule": "2.4.8",
+        "label": "Location",
+        "status": "not yet evaluated"
```
jcfilben, line 276: I don't think this rule is applicable

aries-site/src/data/wcag/components.json
```diff
@@ -48,5 +48,419 @@
...
+      },
+      {
+        "rule": "2.4.9",
+        "label": "Link Purpose (Link Only)",
+        "status": "conditional",
+        "notes": "Related cards sharing a common call to action such as 'Learn More' may need to be differentiated by using `aria-label` to provide context as to answer 'learn more about what?'."
```
jcfilben, line 282: For this rule having an aria-label wouldn't be enough in the case where we have links with the text, 'learn more'. This rule requires that the link be identifiable from link text alone.

jcfilben:
Can we add the following to the PR description so it gets flagged as an updated section for people visiting the site?
```
#### Notifications
[Update]Card

[Accessibility]

```
https://github.com/grommet/hpe-design-system/wiki/Inline-Notifications-Guidance

aries-site/src/data/wcag/components.json
halocline: https://www.w3.org/WAI/WCAG22/Understanding/link-purpose-link-only#sufficient

aries-site/src/data/wcag/components.json
halocline: Marked as "failed" and expanded on rationale in notes.

aries-site/src/data/wcag/components.json
jcfilben: Thanks for flagging, I misunderstood when I read through previously. Agree aria-label is sufficient

jcfilben (approved):
Looks good!

PR #4917, opened 2025-03-20
Add tests for design token output formats

taysea:

#### What does this PR do?
- Adds tests to check output formats from build-style-dictionary. This will give greater confidence if we ever make adjustments/edits to our build step/transforms that outputs aren't changing in unexpected ways.
- Updates tests to use vitest instead of jest due to various configuration issues (which others on stack overflow were experiencing and were suggesting to just switch to vitest to avoid needing additional config), also this is a good place to get familiar with vitest which is growing in usage.
- Removed jest dependencies / config.
- Moved tests all into a `tests` directory.

**General UX Checks**
- [ ] Light & dark modes

**Accessibility Checks**
- [ ] Keyboard interactions

**Code Quality Checks**

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@
 
 # testing
 /coverage
+*/src/tests/build
 
 #lock files
 *package-lock.json
--- a/design-tokens/jest.config.cjs
+++ b/design-tokens/jest.config.cjs
@@ -1,6 +1,0 @@
-module.exports = {
-  moduleNameMapper: {
-    '^(\\.{1,2}/.*)\\.js$': '$1',
-  },
-  moduleFileExtensions: ['ts', 'js'],
-};
--- a/design-tokens/package.json
+++ b/design-tokens/package.json
@@ -30,7 +30,7 @@
     "release-stable": "node src/scripts/release-stable.js  --branch=design-tokens-stable",
     "paddingY:verify": "tsx src/scripts/verify-paddingY.ts",
     "paddingY:update": "tsx src/scripts/update-paddingY.ts",
-    "test": "jest"
+    "test": "vitest --watch=false"
   },
   "devDependencies": {
     "@babel/preset-env": "^7.26.0",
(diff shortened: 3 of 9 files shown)


britt6612:
need to resolve yarn conflict.


This makes sense for me to change to using Vitest I did read it does have its advantages in terms of speed. I did see for non-vite projects it may not be the best to use, this is not our case though. The code for the test looks good to me as far as failing if the generated file doesn't match the string. Given the trouble you encountered trying to use jest Im good with this direction.

britt6612 (approved):

taysea:
> I did see for non-vite projects it may not be the best to use, this is not our case though.

@britt6612 Could you share the link to where you read about this just for future reference?

britt6612:
> > I did see for non-vite projects it may not be the best to use, this is not our case though.
> 
> @britt6612 Could you share the link to where you read about this just for future reference?

Here is a few articles I read
https://www.divotion.com/blog/why-vitest-might-be-better-than-jest-for-your-javascript-projects
https://github.com/vitest-dev/vitest/discussions/3362/
https://dev.to/colin-williams-dev/testing-w-vitest-wo-vite-5820

I think from what I read it may require extra configuration than with using a Vite project. Also the performance Vitest an advantage is the performance but when used with non Vite project you may not get that benefit.

taysea:
> > > I did see for non-vite projects it may not be the best to use, this is not our case though.
> > 
> > 
> > @britt6612 Could you share the link to where you read about this just for future reference?
> 
> Here is a few articles I read https://www.divotion.com/blog/why-vitest-might-be-better-than-jest-for-your-javascript-projects [vitest-dev/vitest#3362](https://github.com/vitest-dev/vitest/discussions/3362) https://dev.to/colin-williams-dev/testing-w-vitest-wo-vite-5820
> 
> I think from what I read it may require extra configuration than with using a Vite project. Also the performance Vitest an advantage is the performance but when used with non Vite project you may not get that benefit.

Thank you!

halocline (approved):

PR #4959, opened 2025-03-25
Tag wcag documentation & file cleanup

jcfilben:
[Deploy Preview](https://deploy-preview-4959--keen-mayer-a86c8b.netlify.app/)

#### What does this PR do?
- Adds tag component documentation (I did the Tag assessment & documentation as part of a demo, just getting around to pushing up the changes)
- Changed the file structure a little. The `data/wcag/components` file was getting really long and messy so I instead created a `data/wcag/[component name].json` file for each component
- Fixed an issue where the info icon was being displayed on the accordion panel when it should have been displaying the green circle icon.

**General UX Checks**
- [ ] Light & dark modes

**Accessibility Checks**
- [ ] Keyboard interactions

**Code Quality Checks**

https://github.com/grommet/hpe-design-system/issues/4846
#### Notifications
[Update]Tag

[Accessibility]


aries-site/src/data/wcag/tag.json
```diff
@@ -0,0 +1,512 @@
...
+    "rule": "2.5.7",
+    "label": "Dragging Movements",
+    "status": "not-applicable",
+    "notes": ""
+  },
+  {
+    "rule": "2.5.8",
+    "label": "Target Size (Minimum)",
+    "status": "failed",
+    "notes": "x-small tag is 22px. We should consider adding guidance that when a tag is clickable x-small shouldn't be used."
```
jcfilben, line 372:
```suggestion
    "status": "passed",
    "notes": "x-small tag close button is 22px but it passes the spacing exception of the rule. You can have a 24x24 box around the target without intersecting other click targets."
```

aries-site/src/data/wcag/tag.json
```diff
@@ -0,0 +1,512 @@
...
+  {
+    "rule": "1.4.10",
+    "label": "Reflow",
+    "status": "passed",
+    "notes": ""
+  },
+  {
+    "rule": "1.4.11",
+    "label": "Non-text Contrast",
+    "status": "passed",
```
britt6612, line 161: what about border on the highlight version of Tag

aries-site/src/data/wcag/tag.json
```diff
@@ -0,0 +1,512 @@
...
+    "label": "No Keyboard Trap",
+    "status": "passed",
+    "notes": ""
+  },
+  {
+    "rule": "2.1.3",
+    "label": "Keyboard (No Exception)",
+    "status": "passed",
+    "notes": ""
+  },
```
britt6612, line 193: what about rule 2.1.4 character key shortcuts

aries-site/src/data/wcag/tag.json
```diff
@@ -0,0 +1,512 @@
...
+    "rule": "2.5.1",
+    "label": "Pointer Gestures",
+    "status": "not-applicable",
+    "notes": ""
+  },
+  {
+    "rule": "2.5.2",
+    "label": "Pointer Cancellation",
+    "status": "conditional",
+    "notes": "Marking conditional because it depends on the implementation, if there is an easy way to undo clicking a tag or closing it"
```
britt6612, line 336: agree here so just so I understand 

if I user clicks the remove on the tag but didnt mean to the implementor should 
provide a confirmation dialog or cancel option that lets the user confirm the action before the tag is deleted?

aries-site/src/data/wcag/tag.json
```diff
@@ -0,0 +1,512 @@
...
+  },
+  {
+    "rule": "2.5.3",
+    "label": "Label in Name",
+    "status": "passed",
```
britt6612, line 341: would this be passed? If we have a removable tag does the screen reader announce "remove tag (name)" I did not think but we can re check

aries-site/src/data/wcag/tag.json
britt6612: 👍

aries-site/src/data/wcag/tag.json
```diff
@@ -368,8 +368,8 @@
   {
     "rule": "2.5.8",
     "label": "Target Size (Minimum)",
-    "status": "failed",
-    "notes": "x-small tag is 22px. We should consider adding guidance that when a tag is clickable x-small shouldn't be used."
+    "status": "passed",
+    "notes": "x-small tag close button is 22px but it passes the spacing exception of the rule. You can have a 24x24 box around the target without intersecting other click targets."
```
britt6612, line 372: 👍

aries-site/src/data/wcag/tag.json
```diff
@@ -0,0 +1,512 @@
+[
+  {
+    "rule": "1.1.1",
+    "label": "Non-text content",
+    "status": "failed",
+    "notes": "There is no aria-label on the close icon button and no way for the implementer to provide one."
```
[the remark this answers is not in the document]
taysea: Was going to comment the same -- feels like we're passing but there could be an enhancement to let people pass their own.

aries-site/src/data/wcag/tag.json
```diff
@@ -0,0 +1,512 @@
...
+  },
+  {
+    "rule": "1.2.1",
+    "label": "Audio-only and Video-only (Prerecorded)",
+    "status": "not-applicable",
+    "notes": ""
```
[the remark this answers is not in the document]
jcfilben: Good question, I generated the json file by exporting the data from this excel file: https://hpe.sharepoint.com/:x:/r/teams/hpe-design/_layouts/15/Doc.aspx?sourcedoc=%7BE975E19F-3264-4F4C-BD0A-446F23F89277%7D&file=WCAG%20Rule%20Evaluation%20Template.xlsx&action=default&mobileredirect=true&DefaultItemOpen=1.
This is why the notes section is included for each rule even though it is empty.

I think for simplicity of exporting the data, I'm fine if we have the notes section still included even if it is blank. Especially since we are moving to separate json files for each component, the length of the json is less of a concern

aries-site/src/data/wcag/tag.json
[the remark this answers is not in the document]
jcfilben: Good call, I agree this is more of an enhancement than an issue. I'll update this to passing

aries-site/src/data/wcag/tag.json
```diff
@@ -0,0 +1,512 @@
...
+  {
+    "rule": "1.3.6",
+    "label": "Identify Purpose",
+    "status": "failed",
+    "notes": "The close icon doesn't have a descriptive aria-label"
+  },
+  {
+    "rule": "1.4.1",
+    "label": "Use of Color",
+    "status": "conditional",
```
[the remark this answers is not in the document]
jcfilben: Yeah I was on the fence about this. I was feeling like the highlight tag 'blue' treatment was a little more on the decorative side than informational. I marked as conditional because if there was a case where it is important for the user to understand that something is a highlight tag rather than a regular tag then that information should be conveyed in text.

Open to other thoughts/perspectives here, that's just what I was thinking

aries-site/src/data/wcag/tag.json
jcfilben: I didn't include the highlight tag as something that needed to meet contrast, I was viewing the blue of the highlight tag as more decorative than informative. See comment here: https://github.com/grommet/hpe-design-system/pull/4959#discussion_r2021526665

aries-site/src/data/wcag/tag.json
jcfilben: They don't necessairly need to provide a confirmation dialog or cancel option, the just need to make sure when building the tag if it includes interaction that one of the following is true:

- No Down-Event: The down-event of the pointer is not used to execute any part of the function;

- Abort or Undo: Completion of the function is on the up-event, and a mechanism is available to abort the function before completion or to undo the function after completion;

- Up Reversal: The up-event reverses any outcome of the preceding down-event;

- Essential: Completing the function on the down-event is essential.

aries-site/src/data/wcag/tag.json
jcfilben: This rule is less about the screenreader and more about the accessible name that accessibility tech uses to refer to the component by. The accessible name should match what we see on the screen. See in the below screenshot the text we see on the screen is 'name : value' and the accessible name is 'name: value'. This helps users who use speech based tech to reference the correct component
Screenshot 2025-03-31 at 12 07 55 PM

aries-site/src/data/wcag/tag.json
jcfilben: Good catch, added this in latest commit

aries-site/src/data/wcag/tag.json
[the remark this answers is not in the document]
jcfilben: Fixed

aries-site/src/data/wcag/tag.json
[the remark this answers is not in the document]
britt6612: in our site we have 
[Highlight tags](https://design-system.hpe.design/components/tag#highlight) can be used to draw attention to an element on the page. They are a decorative variant of the [static tag](https://design-system.hpe.design/components/tag#static). Highlights should only contain descriptions that are temporary, such as “New”, “Updated”. As such they should not be used a permanent design feature.

So I wonder since we are saying here it is a decorative variant but we are also saying these can be used to draw the attention of the user to an element on the page. If this didnt pass then would the underlying really be drawing someone attention. Im open to having this conditional but I wonder if we should call it out in the `hightlight tags` section of the page. @taysea 

also open to other thoughts here as well.

aries-site/src/data/wcag/tag.json
britt6612: Thanks for explaining

aries-site/src/data/wcag/tag.json
britt6612: good to know thank you

aries-site/src/data/wcag/tag.json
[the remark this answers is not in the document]
britt6612: that makes sense I didnt realize they were being exported from the excel file. Thank you it was more of a general question I wasnt concerned with having the empty string

aries-site/src/data/wcag/tag.json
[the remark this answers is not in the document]
taysea: I'm comfortable with "conditional" -- I like the rationale.

britt6612 (approved):

aries-site/src/data/wcag/tag.json
```diff
@@ -0,0 +1,518 @@
...
+    "rule": "1.3.5",
+    "label": "Identify Input Purpose",
+    "status": "not-applicable",
+    "notes": ""
+  },
+  {
+    "rule": "1.3.6",
+    "label": "Identify Purpose",
+    "status": "failed",
+    "notes": "The close icon doesn't have a descriptive aria-label"
```
taysea, line 96: I dropped a comment on this previously, the comment/any replies might've gotten lost in the refactor but I would say this passes. The svg has "Close" on it. Feels like we can enhance to allow more of a customization but would that be sufficient?

jcfilben: Agree, I'll update this to passed

aries-site/src/data/structures/components.js
```diff
@@ -1595,6 +1595,7 @@ export const components = [
   },
   {
     name: 'Tag',
+    accessibility: 'Failed WCAG 2.2 A',
```
taysea, line 1598:
```suggestion
    accessibility: 'Passed WCAG 2.2 AA',
```

Now that the other rule is "passed" for the close icon, can this be changed?

britt6612: ☝️ good catch!

jcfilben: Yes! updated

taysea (approved):
Looks good, thank you!

PR #4992, opened 2025-04-02
Temp solution for fetching wcag rules

jcfilben:
[Deploy Preview](https://deploy-preview-4992--keen-mayer-a86c8b.netlify.app/)

#### What does this PR do?
The link where we were fetching the data for the WCAG rules no longer exists. I looked around to see if there was an alternative but wasn't able to find something. As a temporary solution I copied the wcag.json file into the repo so we could point directly to that.

Currently the wcag rules section on the DS site doesn't work because of the broken fetch link
Screenshot 2025-04-02 at 2 24 40 PM

**General UX Checks**
- [ ] Light & dark modes

**Accessibility Checks**
- [ ] Keyboard interactions

**Code Quality Checks**

no
backwards compatible

### DIFF
--- a/aries-site/src/data/wcag/wcag.json
+++ b/aries-site/src/data/wcag/wcag.json
@@ -1,0 +1,5670 @@
+{
+  "principles": [
+    {
+      "id": "WCAG2:perceivable",
+      "num": "1",
+      "versions": ["2.0", "2.1", "2.2"],
+      "handle": "Perceivable",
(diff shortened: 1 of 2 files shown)


britt6612 (approved):
looks good for the mean time

taysea (approved):
That's a bummer, I'm sure there must be something else out there we can find. Maybe even reaching out to one of the maintainers?

I'm fine with this as an interim solution.

PR #5057, opened 2025-04-17
Added Main accessibility info

oliverHPE:

#### What does this PR do?
Adds accessibility information for 'main'

**General UX Checks**
- [ ] Light & dark modes

**Accessibility Checks**
- [ ] Keyboard interactions

**Code Quality Checks**

https://github.com/grommet/hpe-design-system/issues/5015

#### Notifications
[Update]Main

[Accessibility]

### DIFF
--- a/aries-site/src/data/structures/components.js
+++ b/aries-site/src/data/structures/components.js
@@ -1147,7 +1147,7 @@ export const components = [
   },
   {
     name: 'Main',
-    accessibility: 'Passed WCAG 2.2 AA',
+    accessibility: 'Passed WCAG 2.2 AAA',
     category: 'Layouts',
     description:
       'The Main component is where you define the location and layout of the primary context of your content.',
--- a/aries-site/src/data/structures/components.js
+++ b/aries-site/src/data/structures/components.js
@@ -1147,6 +1147,7 @@ export const components = [
   },
   {
     name: 'Main',
+    accessibility: 'Passed WCAG 2.2 AA',
     category: 'Layouts',
     description:
       'The Main component is where you define the location and layout of the primary context of your content.',
--- a/aries-site/src/data/wcag/main.json
+++ b/aries-site/src/data/wcag/main.json
@@ -1,0 +1,512 @@
+[
+    {
+      "rule": "1.1.1",
+      "label": "Non-text content",
+      "status": "not-applicable",
+      "notes": ""
+    },
+    {
+      "rule": "1.2.1",
+      "label": "Audio-only and Video-only (Prerecorded)",
+      "status": "not-applicable",
+      "notes": ""
+    },
+    {
+      "rule": "1.2.2",
+      "label": "Captions (Prerecorded)",
+      "status": "not-applicable",
+      "notes": ""
+    },
+    {
+      "rule": "1.2.3",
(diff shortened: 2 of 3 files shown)


aries-site/src/data/wcag/main.json
```diff
@@ -0,0 +1,512 @@
...
+    {
+      "rule": "1.3.1",
+      "label": "Info and Relationships",
+      "status": "passed",
+      "notes": ""
+    },
+    {
+      "rule": "1.3.2",
+      "label": "Meaningful Sequence",
+      "status": "not-applicable",
```
britt6612, line 71: @jcfilben should this be conditional?

aries-site/src/data/structures/components.js
```diff
@@ -1147,6 +1147,7 @@ export const components = [
   },
   {
     name: 'Main',
+    accessibility: 'Passed WCAG 2.2 AA',
```
britt6612, line 1150: I believe
```
If we don't pass AAA, then we say 'Passed WCAG 2.2 AA', but if we do meet AAA then we can say 'Passed WCAG 2.2 AAA'
```
so this would be passed wcag 2.2 AAA
@jcfilben

aries-site/src/data/wcag/main.json
jcfilben: I feel like the sequence things are read out in is not controlled directly by the main component so I would leave it as not-applicable

aries-site/src/data/structures/components.js
jcfilben: Yes this should be 'Passed WCAG 2.2 AAA'

aries-site/src/data/wcag/main.json
britt6612: ah okay that being said then should we change the Page component 
https://design-system.hpe.design/components/page#wcag-compliance 
for this rule to be `not-applicable`

aries-site/src/data/structures/components.js
britt6612, line 1150:
```suggestion
    accessibility: 'Passed WCAG 2.2 AAA',
```

aries-site/src/data/wcag/main.json
jcfilben: I could go either way on page, this rule is more related to the contents of the page but I felt like it was a nice thing to call out for people to be aware of as they at working at the page level so I was leaning towards flagging as conditional there. Its subjective though, open to other thoughts

britt6612: I see your point the way I looked at it was the same as Page where Main ensures that whatever content is inside of it is the primary content of the page. So then if the children inside Main are rendered in a logical DOM order kinda the same as with Page.

jcfilben: That makes sense, I would be open to marking this as conditional here too

britt6612: okay Im also open to either but I think if we decide conditional it should apply to both Main and Page or `not-applicable` should be for both as well

britt6612, line 71:
```suggestion
      "status": "conditional",
```

taysea (approved):

PR #5063, opened 2025-04-17
Add accessibility to skeleton

KennyAtHPE:
[Deploy Preview](https://deploy-preview-5063--keen-mayer-a86c8b.netlify.app/components/skeleton?q=skel)

#### What does this PR do?
Adding accessibility to skeleton

**General UX Checks**
- [ ] Light & dark modes

**Accessibility Checks**
- [ ] Keyboard interactions

**Code Quality Checks**

https://github.com/grommet/hpe-design-system/issues/5016
#### Notifications

[Update]Spinner

[Accessibility]


aries-site/src/data/wcag/skeleton.json
```diff
@@ -0,0 +1,513 @@
...
+    {
+      "rule": "1.3.6",
+      "label": "Identify Purpose",
+      "status": "not-applicable",
+      "notes": ""
+    },
+    {
+      "rule": "1.4.1",
+      "label": "Use of Color",
+      "status": "not-applicable",
```
britt6612, line 101: would this be applicable in the case we have the boxes with the light grey color loading?

aries-site/src/data/wcag/skeleton.json
```diff
@@ -0,0 +1,513 @@
...
+    {
+      "rule": "1.4.2",
+      "label": "Audio Control",
+      "status": "not-applicable",
+      "notes": ""
+    },
+    {
+      "rule": "1.4.3",
+      "label": "Contrast (Minimum)",
+      "status": "not-applicable",
```
britt6612, line 113: I would think it would apply

aries-site/src/data/wcag/skeleton.json
```diff
@@ -0,0 +1,513 @@
...
+    },
+    {
+      "rule": "1.4.5",
+      "label": "Images of Text",
+      "status": "not-applicable",
+      "notes": ""
+    },
+    {
+      "rule": "1.4.6",
+      "label": "Contrast (Enhanced)",
```
britt6612, line 130: what about progress bars or something with the like grey that we use think it has to be at least 3:1

aries-site/src/data/wcag/skeleton.json
```diff
@@ -0,0 +1,513 @@
...
+    {
+      "rule": "3.2.4",
+      "label": "Consistent Identification",
+      "status": "conditional",
+      "notes": ""
+    },
+    {
+      "rule": "3.2.5",
+      "label": "Change on Request",
+      "status": "not-applicable",
```
britt6612, line 437: conditional 
When the skeleton is replaced by the actual content, ensure this transition is expected by the user and doesn't introduce sudden changes to the UI that could cause confusion

aries-site/src/data/wcag/skeleton.json
MikeKingdom: The general contrast question is a good one. Skeletons are just placeholders and I'm not convinced they are something required to be visualized all that well. This seems like it needs a bit more research with what others in the industry are doing here.

aries-site/src/data/wcag/skeleton.json
MikeKingdom: Do you feel there is something we can be doing for this or is this really in the domain of the application where they are used? We really don't have enough information to be all that helpful (ie. we can't say what we're waiting for to be loaded etc.)

aries-site/src/data/wcag/skeleton.json
britt6612: agree I know "use of color" if its something decorative then it doesnt apply so we can do some more research

aries-site/src/data/wcag/skeleton.json
britt6612: I think the way I was thinking about this one was a skeleton component should only appear when a user clicks a button, or navigates to a section or something that is user requested. So if a skeleton was shown like on a auto -refresh I would think it may not pass since the skeleton would be unexpected

aries-site/src/data/wcag/skeleton.json
KennyAtHPE: I didn't think that this rule applied, because the skeleton color isn't being used to depict any specific meaning (ex. red = critical).

My interpretation of this rule was that IF color is being used to signify meaning, then there should be an alternative. But in this case, the grey isn't signifying anything.

aries-site/src/data/wcag/skeleton.json
KennyAtHPE: Doesn't this rule only apply for "text and text-like content?" Skeletons are just visual placeholders.

aries-site/src/data/wcag/skeleton.json
KennyAtHPE: Same comment above, not sure if this applies since skeletons are purely visual placeholders.

aries-site/src/data/wcag/skeleton.json
jcfilben: ^ Agree, I think 'not-applicable' is correct here

aries-site/src/data/wcag/skeleton.json
jcfilben: > Doesn't this rule only apply for "text and text-like content?" Skeletons are just visual placeholders.

Correct, I think 'not-applicable' is the right status

aries-site/src/data/wcag/skeleton.json
jcfilben: > Same comment above, not sure if this applies since skeletons are purely visual placeholders.

Agree, let's leave as not-applicable

aries-site/src/data/wcag/skeleton.json
jcfilben: I feel like this may fall slightly outside of the scope of the Skeleton component and be more of an application design thing but I would be okay with marking as conditional to help flag for people using a skeleton component that this is something they should be aware of.

KennyAtHPE: Agree that it'd probably be important to flag regardless.

aries-site/src/data/wcag/skeleton.json
britt6612: okay so this would fall under the decorative category then these rules would not apply

aries-site/src/data/wcag/skeleton.json
```diff
@@ -0,0 +1,513 @@
...
+    {
+      "rule": "1.4.10",
+      "label": "Reflow",
+      "status": "passed",
+      "notes": ""
+    },
+    {
+      "rule": "1.4.11",
+      "label": "Non-text Contrast",
+      "status": "failed",
```
britt6612, line 161: I wonder then given the notes if this would be conditional

aries-site/src/data/wcag/skeleton.json
britt6612: @KennyAtHPE do you wanna just add a note in the notes here

aries-site/src/data/wcag/skeleton.json
```diff
@@ -0,0 +1,513 @@
...
+      "notes": "It depends if skeletons are labeled as graphic objects, if so they must meet 3:1 contrast ratio. This seems like an open ended conversation and there isn't any actual classification for Skeletons. Relevant github discussion (https://github.com/w3c/wcag/issues/2048)."
```
britt6612, line 162: can we update the note here to why Skeletons do not need to meet contrast

britt6612: https://medium.com/@accessibilityjillie/day-1-skeleton-loaders-accessibility-ec750fcceba7

jcfilben (approved):
Looks good!

Issue #5013, opened 2025-04-08
Notification - DS site WCAG accessibility documentation


jcfilben:
Add Accessibility information to the Notification page on the Design System site.

- [x] Assess which [WCAG rules](https://www.w3.org/WAI/WCAG22/quickref/?versions=2.2) are applicable to the component and if they are passing/failing.
- [x] Create a pull request on the Design System site to add this information to the component page. Follow the format here: https://github.com/grommet/hpe-design-system/pull/5004/files. The finished result should look something like this: https://design-system.hpe.design/components/menu?q=menu#wcag-compliance.
- [ ] Create issues in the relevant repository (grommet, grommet-theme-hpe, etc.) for any rules where we are not meeting WCAG AA or if there is something observed that it feels like we should fix. We should create 1 issue per rule failure. Issues where we are not meeting AA should be tagged with 'bug'. All issues created through this process should have the 'accessibility' label and be added to the accessibility board.

The pull request should be up on Github within the first week of the sprint so that the second week of the sprint can be focused on review and polishing.

When your work is ready for review please get a first pass review from another member of the accessibility sprint and once that is done reach out to Jessica to flag that a final review is needed.

Resources and videos for this work can be found here: https://hpe-my.sharepoint.com/shared?id=%2Fteams%2Fhpe%2Ddesign%2FShared%20Documents%2FHPE%20Design%20System%2FAccessibility&listurl=https%3A%2F%2Fhpe%2Esharepoint%2Ecom%2Fteams%2Fhpe%2Ddesign%2FShared%20Documents

SeamusLeonardHPE:
https://v2.grommet.io/notification?theme=hpe

https://storybook.grommet.io/?path=/story/visualizations-notification-global--global&globals=theme:hpe

https://design-system.hpe.design/components/notification

luketa8:
First pass review done. Thanks!

SeamusLeonardHPE:
Thanks @luketa8 

@jcfilben @taysea can you have a look at notification for a final pass on the scoring (I may ask for guidance on any notes that we should publish after your review)

britt6612:
@SeamusLeonardHPE 
left comments on your figma to add issues

britt6612:
@SeamusLeonardHPE 

Just looks like we need to file these last 3 issues
1.4.6 
2.4.9
2.5.5

britt6612:
@SeamusLeonardHPE thank you! closing issue now

PR #5065, opened 2025-04-18
Notification WCAG file

SeamusLeonardHPE:
[Deploy Preview](https://deploy-preview-4959--keen-mayer-a86c8b.netlify.app/)

#### What does this PR do?
- Notification wcag info

**General UX Checks**
- [ ] Light & dark modes

**Accessibility Checks**
- [ ] Keyboard interactions

**Code Quality Checks**

https://github.com/grommet/hpe-design-system/issues/5013
#### Notifications
[Update]Notification

[Accessibility]

### DIFF
--- a/aries-site/src/data/wcag/notification.json
+++ b/aries-site/src/data/wcag/notification.json
@@ -1,0 +1,512 @@
+[
+  {
+    "rule": "1.1.1",
+    "label": "Non-text content",
+    "status": "passed",
+    "notes": ""
+  },
+  {
+    "rule": "1.2.1",
+    "label": "Audio-only and Video-only (Prerecorded)",
+    "status": "not-applicable",
+    "notes": ""
+  },
+  {
+    "rule": "1.2.2",
+    "label": "Captions (Prerecorded)",
+    "status": "not-applicable",
+    "notes": ""
+  },
+  {
+    "rule": "1.2.3",
+    "label": "Audio Description or Media Alternative (Prerecorded)",
+    "status": "not-applicable",
+    "notes": ""
+  },
+  {
+    "rule": "1.2.4",
+    "label": "Captions (Live)",
+    "status": "not-applicable",
+    "notes": ""
+  },
+  {
+    "rule": "1.2.5",
+    "label": "Audio Description (Prerecorded)",
+    "status": "not-applicable",
+    "notes": ""
+  },
+  {
+    "rule": "1.2.6",
+    "label": "Sign Language (Prerecorded)",
+    "status": "not-applicable",
+    "notes": ""
+  },
+  {
+    "rule": "1.2.7",
+    "label": "Extended Audio Description (Prerecorded)",
+    "status": "not-applicable",
+    "notes": ""
+  },
+  {
+    "rule": "1.2.8",
+    "label": "Media Alternative (Prerecorded)",
+    "status": "not-applicable",
+    "notes": ""
+  },
+  {
+    "rule": "1.2.9",
+    "label": "Audio-only (Live)",
+    "status": "not-applicable",
+    "notes": ""
+  },
+  {
+    "rule": "1.3.1",
+    "label": "Info and Relationships",
+    "status": "failed",
+    "notes": ""
+  },
+  {
+    "rule": "1.3.2",
+    "label": "Meaningful Sequence",
+    "status": "passed",
+    "notes": ""
+  },
+  {
+    "rule": "1.3.3",
+    "label": "Sensory Characteristics",
+    "status": "passed",
+    "notes": ""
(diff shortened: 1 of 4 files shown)


aries-site/src/data/wcag/notification.json
```diff
@@ -0,0 +1,512 @@
...
+  {
+    "rule": "3.2.4",
+    "label": "Consistent Identification",
+    "status": "not-applicable",
+    "notes": ""
+  },
+  {
+    "rule": "3.2.5",
+    "label": "Change on Request",
+    "status": "conditional",
```
britt6612, line 437: @taysea I was going back and forth on failed or conditional

britt6612: I would think this has to do with 
content changes automatically (Toast notification), the user should have the ability to stop, pause, or adjust that change..
thoughts?

taysea: I think conditional is fine here. For toast, we build in the onEsc behavior for someone to close it. If someone's preference were to be to not have toast notifications at all, this feels like it would be conditional to app implementation/user preferences.

britt6612: okay sounds good.

britt6612 (approved):

aries-site/src/data/wcag/notification.json
```diff
@@ -0,0 +1,512 @@
...
+    "rule": "2.1.3",
+    "label": "Keyboard (No Exception)",
+    "status": "passed",
+    "notes": ""
+  },
+  {
+    "rule": "2.2.1",
+    "label": "Timing Adjustable",
+    "status": "failed",
+    "notes": ""
```
taysea, line 198: Wait should this be conditional? It's available via prop: https://v2.grommet.io/notification#time

aries-site/src/data/wcag/notification.json
```diff
@@ -0,0 +1,512 @@
...
+  {
+    "rule": "2.2.2",
+    "label": "Pause, Stop, Hide",
+    "status": "passed",
+    "notes": ""
+  },
+  {
+    "rule": "2.2.3",
+    "label": "No Timing",
+    "status": "failed",
```
taysea, line 209: Would this also be conditional? https://v2.grommet.io/notification#toast Implementor could set `autoClose={false}`, meaning that could be exposed as a user preference in an app.

aries-site/src/data/wcag/notification.json
```diff
@@ -0,0 +1,512 @@
...
+  {
+    "rule": "2.4.3",
+    "label": "Focus Order",
+    "status": "passed",
+    "notes": ""
+  },
+  {
+    "rule": "2.4.4",
+    "label": "Link Purpose (In Context)",
+    "status": "failed",
```
taysea, line 269: @britt6612 Which "link" does this refer to? https://v2.grommet.io/notification#props All anchor props are supported here.

aries-site/src/data/wcag/notification.json
britt6612: For AA compliance: Avoid auto-dismiss toasts unless users can adjust or disable the timer.
Alternatively, switch toast behavior to:
→ "Remains until dismissed" OR
→ "Includes pause/dismiss options when focused"


users cant really adjust that? or would it be the case that someone implementing can?
Disappears automatically after a few seconds
- Has no visible controls to extend/pause
- Offers no user setting to change timing

aries-site/src/data/wcag/notification.json
britt6612: ah yeah then that way the user could close on their own?

aries-site/src/data/wcag/notification.json
britt6612: For the `anchors` in this component we only give implementors access to the label prop, we should give them access to set the aria-label as well.

☝️ what Jessica mentioned

aries-site/src/data/wcag/notification.json
britt6612: in that case I would agree conditional

aries-site/src/data/wcag/notification.json
taysea: Hmm -- we can see what @jcfilben thinks next week. My feeling is that the grommet component offers the ability to. So it would be "conditional" for the product team to provide the user some option in their "settings".

aries-site/src/data/wcag/notification.json
taysea: I would still think this is conditional because shouldn't the "label" be something meaningful? The label is read out if no aria-label is present anyway.

PR #5093, opened 2025-04-24
Update `dataVis.categorical` palette

### LINKED ISSUE #5033: Refine dataVis-categorical palette
Based on feedback received, refine the dataVis-categorical palette to meet these success criteria:
- Color meets 3:1 on background-front and background-back
- Colors are distinguishable across various color-blindnesses 
- Colors would not be confused with status colors (when status colors are present on same screen or absent)
- First color is desirable if grabbed as the default for a meter, chart, etc.
- Overall palette aligns with desired brand “feel”

Deliverable:
- [x] Proposal: https://www.figma.com/design/IWBHg4LxqZA9zO9tRIPMIM/Color-palette-testing?node-id=2189-11338&t=eJCqeUHysBJQIX6d-1

taysea:

#### What does this PR do?
Based on theme feedback, updating `datavis.categorical` palette. Approach was to replace primitive values with closest match. If no primitive value has close, just hard code for now in semantic files (this is our current approach given we haven't had a chance to deeply systematize the primitive palette).

Also added docs for the last datavis color (which was missing from docs example).

Has received approval from Chris, but leaving as draft until Friday to allow time for final feedback from a few product designers. Will merge on Monday if no feedback received.

design-tokens/tokens/primitive/primitives.default.json

In Figma, contrast values have been checked: https://www.figma.com/design/IWBHg4LxqZA9zO9tRIPMIM/Color-palette-testing?node-id=2165-5332&t=eJCqeUHysBJQIX6d-4
In local DS site, checked that values map through to grommet-theme-hpe as expected:

Screenshot 2025-04-24 at 1 30 17 PM

**General UX Checks**
- [ ] Light & dark modes

**Accessibility Checks**
- [ ] Keyboard interactions

**Code Quality Checks**

Closes #5033

### DIFF
--- a/aries-site/src/data/color.js
+++ b/aries-site/src/data/color.js
@@ -449,6 +449,11 @@ const graphColorsLight = [
     value: colors['graph-6'].light,
     hex: colors['graph-6'].light,
   },
+  {
+    name: 'graph-7',
+    value: colors['graph-7'].light,
+    hex: colors['graph-7'].light,
+  },
 ];
 
 const graphColorsDark = [
@@ -486,6 +491,11 @@ const graphColorsDark = [
     name: 'graph-6',
     value: colors['graph-6'].dark,
     hex: colors['graph-6'].dark,
+  },
+  {
+    name: 'graph-7',
+    value: colors['graph-7'].dark,
+    hex: colors['graph-7'].dark,
   },
 ];
 
--- a/design-tokens/tokens/primitive/primitives.default.json
+++ b/design-tokens/tokens/primitive/primitives.default.json
@@ -1286,7 +1286,7 @@
         },
         "gold1": {
           "$type": "color",
-          "$value": "#bd8d09",
+          "$value": "#b0840d",
           "$description": "",
           "$extensions": {
             "com.figma": {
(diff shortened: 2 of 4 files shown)


halocline (approved):
Quick question re: primitive color name space. We are good with names like `gold1` instead of pointing to various values in our core color palettes e.g `yellow.300`?

NVM, using the `dataVis.gold1` namespace. If we wanted to adjust in the future I think we're okay because we state primitives shouldn't be used directly and thus not part of supported API, correct?

taysea:
> Quick question re: primitive color name space. We are good with names like `gold1` instead of pointing to various values in our core color palettes e.g `yellow.300`?
> 
> NVM, using the `dataVis.gold1` namespace. If we wanted to adjust in the future I think we're okay because we state primitives shouldn't be used directly and thus not part of supported API, correct?

Correct, exactly. And we can certainly revisit once we "systematize" the primitive set in the future if we'd prefer datavis colors to come from the core palettes. But at the moment yes they are a separate set, but the naming is not something we're too focused on given we're not recommending people to use directly.

PR #5165, opened 2025-05-15
Footer a11y guidance

luketa8:
[Deploy Preview](https://deploy-preview-5165--keen-mayer-a86c8b.netlify.app/components/footer)

### DIFF
--- a/aries-site/src/data/structures/components.js
+++ b/aries-site/src/data/structures/components.js
@@ -830,7 +830,7 @@ export const components = [
   },
   {
     name: 'Footer',
-    accessibility: 'Passed WCAG 2.2 AA',
+    accessibility: 'Passed WCAG 2.2 AAA',
     category: 'Layouts',
     description:
       'Footer is a Box with a set of preset properties. It is usually placed at the bottom of the page, which makes it predictable for users to find certain information or actions.',
--- a/aries-site/src/data/structures/components.js
+++ b/aries-site/src/data/structures/components.js
@@ -830,6 +830,7 @@ export const components = [
   },
   {
     name: 'Footer',
+    accessibility: 'Passed WCAG 2.2 AA',
     category: 'Layouts',
     description:
       'Footer is a Box with a set of preset properties. It is usually placed at the bottom of the page, which makes it predictable for users to find certain information or actions.',
--- a/aries-site/src/data/wcag/footer.json
+++ b/aries-site/src/data/wcag/footer.json
@@ -1,0 +1,512 @@
+[
+  {
+    "rule": "1.1.1",
+    "label": "Non-text content",
+    "status": "not-applicable",
+    "notes": ""
+  },
+  {
+    "rule": "1.2.1",
+    "label": "Audio-only and Video-only (Prerecorded)",
+    "status": "not-applicable",
+    "notes": ""
+  },
+  {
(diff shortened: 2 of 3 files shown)


aries-site/src/data/wcag/footer.json
```diff
@@ -0,0 +1,512 @@
...
+    "notes": ""
+  },
+  {
+    "rule": "1.3.3",
+    "label": "Sensory Characteristics",
+    "status": "not-applicable",
+    "notes": ""
+  },
+  {
+    "rule": "1.3.4",
```
luketa8, line 81: @jcfilben 
I think we are passing this rule since we do not lock the footer to one specific orientation.

aries-site/src/data/wcag/footer.json
```diff
@@ -0,0 +1,512 @@
...
+    "notes": ""
+  },
+  {
+    "rule": "2.3.3",
+    "label": "Animation from Interactions",
+    "status": "not-applicable",
+    "notes": ""
+  },
+  {
+    "rule": "2.4.1",
```
luketa8, line 249: @jcfilben 
Not required since we are using semantic HTML `<footer>`, but Footer accessibility can be improved by using `role="contentinfo"` and, if there are multiple footers on the page, adding an `aria-label` or `aria-labelledby` to distinguish them.

luketa8 (commented):
@jcfilben 
Ready for review

aries-site/src/data/wcag/footer.json
jcfilben: Yeah I think marking this as passed makes sense

aries-site/src/data/wcag/footer.json
jcfilben: It seems like using the native html footer element is preferred over using `role="contentinfo"` based on the infomation on this page: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/contentinfo_role.

If we add `role="contentinfo"` to the footer in what way does the accessibility experience improve?

aries-site/src/data/wcag/footer.json
```diff
@@ -0,0 +1,512 @@
...
+  {
+    "rule": "3.3.9",
+    "label": "Accessible Authentication (Enhanced)",
+    "status": "not-applicable",
+    "notes": ""
+  },
+  {
+    "rule": "4.1.2",
+    "label": "Name, Role, Value",
+    "status": "not-applicable",
```
jcfilben, line 503:
```suggestion
    "status": "passed",
```

Footer has the role 'contentinfo' by default

aries-site/src/data/structures/components.js
```diff
@@ -830,6 +830,7 @@ export const components = [
   },
   {
     name: 'Footer',
+    accessibility: 'Passed WCAG 2.2 AA',
```
jcfilben, line 833: Looks like we are meeting AAA

jcfilben, line 833:
```suggestion
    accessibility: 'Passed WCAG 2.2 AAA',
```

luketa8:
Thanks! 
Committed suggested changes @jcfilben

jcfilben (approved):
Looks good!

PR #5173, opened 2025-05-20
Tip WCAG eval

KennyAtHPE:
[Deploy Preview](https://deploy-preview-5173--keen-mayer-a86c8b.netlify.app/components/tip?q=tip)

#### What does this PR do?
Adding Tip WCAG evaluation

**General UX Checks**
- [ ] Light & dark modes

**Accessibility Checks**
- [ ] Keyboard interactions

**Code Quality Checks**


aries-site/src/data/wcag/tip.json
```diff
@@ -0,0 +1,512 @@
...
+    "rule": "1.4.7",
+    "label": "Low or No Background Audio",
+    "Column1": "not-applicable",
+    "notes": ""
+  },
+  {
+    "rule": "1.4.8",
+    "label": "Visual Presentation",
+    "Column1": "conditional",
+    "notes": "Depends on the length of the text, but it's recommended to keep Tips short so shouldn't apply in a majority of use cases."
```
jcfilben, line 144: I think this is passing

aries-site/src/data/wcag/tip.json
```diff
@@ -0,0 +1,512 @@
...
+    "rule": "2.2.6",
+    "label": "Timeouts",
+    "Column1": "not-applicable",
+    "notes": ""
+  },
+  {
+    "rule": "2.3.1",
+    "label": "Three Flashes or Below Threshold",
+    "Column1": "conditional",
+    "notes": "Dependent on tip animation"
```
jcfilben, line 234: I don't think we support animation on Tip at the moment

aries-site/src/data/wcag/tip.json
```diff
@@ -0,0 +1,512 @@
...
+  {
+    "rule": "2.5.7",
+    "label": "Dragging Movements",
+    "Column1": "not-applicable",
+    "notes": ""
+  },
+  {
+    "rule": "2.5.8",
+    "label": "Target Size (Minimum)",
+    "Column1": "not-applicable",
```
jcfilben, line 371: Conditional based on the trigger used for tip

aries-site/src/data/wcag/tip.json
```diff
@@ -0,0 +1,512 @@
...
+    "rule": "3.3.9",
+    "label": "Accessible Authentication (Enhanced)",
+    "Column1": "not-applicable",
+    "notes": ""
+  },
+  {
+    "rule": "4.1.2",
+    "label": "Name, Role, Value",
+    "Column1": "conditional",
+    "notes": "Tips must be given the correct aria role consistently"
```
jcfilben, line 504: Agree this is conditional
I think we could make an enhancement issue to improve the out of the box behavior. Users can pass `role='tooltip'` through `dropProps` in Tip to pass this rule, but we could enhance Grommet to add this automatically for them.
Can we file an enhancement issue for this?

aries-site/src/data/wcag/tip.json
```diff
@@ -0,0 +1,512 @@
...
+    "rule": "1.2.9",
+    "label": "Audio-only (Live)",
+    "Column1": "not-applicable",
+    "notes": ""
+  },
+  {
+    "rule": "1.3.1",
+    "label": "Info and Relationships",
+    "Column1": "conditional",
+    "notes": "Tip needs to be given an aria-label, aria-labelledby, or aria-describedby so that the tip can clearly be associated with its trigger."
```
jcfilben, line 66: Can we file issues for the following:

- on grommet-site add an issue to improve the documentation to make it more clear that an `id` needs to be passed through `dropProps` and the tooltip trigger should refence this id via `aria-describedby`

- on grommet create an issue to check all components that use Tip internally and make sure we follow the `id` + `aria-describedby` pattern.

- on grommet create an issue to update storybook examples to demonstrate the `id` + `aria-describedby` pattern.

aries-site/src/data/wcag/tip.json
KennyAtHPE: @jcfilben I had seen this [example](https://storybook.grommet.io/?path=/story/controls-tip-animated--animated&globals=theme:hpe) in the storybook. It appears to be a simple fadein, which wouldn't be a concern for this rule but just wanted to be sure before changing this rule status.

aries-site/src/data/wcag/tip.json
KennyAtHPE: @jcfilben Wouldn't this rule apply to the trigger only, not necessarily the tip itself? It seems like it applies indirectly to the tip. I don't have a hard opinion on this either way but just want to double check.

aries-site/src/data/wcag/tip.json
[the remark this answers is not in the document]
jcfilben: I think it is still conditional, the implementer would need to pass in `aria-describedby` to the tooltip trigger. Grommet can't handle this part because the value of `aria-describedby` needs to match the `id` that the implementer puts on the tooltip drop

aries-site/src/data/wcag/tip.json
```diff
@@ -0,0 +1,512 @@
+[
+  {
+    "rule": "1.1.1",
+    "label": "Non-text content",
+    "Column1": "conditional",
+    "notes": "Implementor needs to apply the correct aria role"
```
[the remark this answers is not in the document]
jcfilben: Yes let's mark as not-applicable

aries-site/src/data/wcag/tip.json
jcfilben: Good find, this story is a bit of a custom build but I would be fine leaving the animation related rules as conditional

aries-site/src/data/wcag/tip.json
jcfilben: Yeah that's fair I think it does apply to trigger only, I'm good leaving this as not-applicable

jcfilben (approved):
Looks good!

PR #5297, opened 2025-09-04
Add announcements to Layer example

jcfilben:

**General UX Checks**
- [ ] Light & dark modes

**Accessibility Checks**
- [ ] Keyboard interactions

**Code Quality Checks**

### DIFF
--- a/aries-site/src/examples/components/layer/LayerExample.js
+++ b/aries-site/src/examples/components/layer/LayerExample.js
@@ -1,5 +1,5 @@
-import React from 'react';
-import { Button, Box, Layer } from 'grommet';
+import React, { useContext } from 'react';
+import { AnnounceContext, Button, Box, Layer } from 'grommet';
 import { LayerHeader } from 'aries-core';
 import { LayerForm } from './ConfigurationForm';
 import {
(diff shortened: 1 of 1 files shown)


aries-site/src/examples/components/layer/LayerExample.js
```diff
@@ -9,33 +9,50 @@ import {
...
+    </ConfirmationProvider>
+  );
+};
 
 const AddApplication = ({ ...rest }) => {
+  const announce = useContext(AnnounceContext);
   const { onClose } = useConfirmation();
 
+  const closeLayer = () => {
+    announce('Add application modal closed.', 'polite');
```
halocline, line 48: Only thing I noticed was that this announce was difficult to perceive, because the button's label would immediately get read.

The "modal opened" reads great.

What's your thought on "aggressive" vs. "polite" in this scenario?

aries-site/src/examples/components/layer/LayerExample.js
```diff
@@ -9,33 +9,50 @@ import {
...
-          {showLayer ? <AddApplication /> : null}
-          {showConfirmation ? <DoubleConfirmation title="application" /> : null}
-        </>
-      )}
-    </ConfirmationContext.Consumer>
-  </ConfirmationProvider>
-);
+export const LayerExample = () => {
+  const announce = useContext(AnnounceContext);
+
```
halocline, line 14: Can we do something like:
```suggestion
const modalTitle = "Add application modal";
```

Then in each of the announces:
```
announce(`${modalTitle} opened.`, 'polite');

....

announce(`${modalTitle} closed.`, 'polite');
```

This way the messages are synched / aligned.

aries-site/src/examples/components/layer/LayerExample.js
jcfilben: Yes tested both 'polite' and 'assertive' here, but in both cases the label's announcement takes precedence. See comment in PR description

> The close layer announcement gets interrupted by the button label announcement, no great way around this unless we delayed moving focus to the button after the layer closes which isn't ideal. So I left as is, I think the more important announcement is when the layer opens.

PR #5478, opened 2025-10-30
Selector box shadow change

jcfilben:

#### What does this PR do?
Add another visual indicator besides just color that the selector is selected.

NOTE: I think this is an okay solution for now but we should probably revisit at some point and explore more options because in some high contrast modes box shadows are removed.

Screenshot 2025-10-30 at 1 58 47 PM

**General UX Checks**
- [ ] Light & dark modes

**Accessibility Checks**
- [ ] Keyboard interactions

**Code Quality Checks**

### DIFF
--- a/aries-core/src/js/components/core/Selector/Selector.js
+++ b/aries-core/src/js/components/core/Selector/Selector.js
@@ -12,7 +12,7 @@ const StyledBox = styled(Box)`
   ${props =>
     props.selected &&
     `box-shadow: inset 0 0 0 ${
-      props.theme.global.edgeSize['5xsmall']
+      props.theme.global.borderSize.small
     } ${normalizeColor(props.border.color, props.theme)};
     };`}
 `;


jcfilben:
Here is with the edge size change (edgeSize.5xsmall)

Screenshot 2025-10-30 at 2 38 34 PM

aries-core/src/js/components/core/Selector/Selector.js
```diff
@@ -10,8 +11,10 @@ import { SelectorHeader } from './SelectorHeader';
 const StyledBox = styled(Box)`
   ${props =>
     props.selected &&
-    // eslint-disable-next-line max-len
-    `box-shadow: inset 0 0 0 1px ${props.theme.global.colors['border-selected']};`}
+    `box-shadow: inset 0 0 0 ${
+      props.theme.global.edgeSize['5xsmall']
```
halocline: I think I misled you with edgeSize. Don't we have [border size](https://github.com/grommet/grommet-theme-hpe/blob/1205e72a8c493ab365b817722e8d5f681c2c848e/src/js/themes/hpe.js#L127)?

Also, shouldn't these dimensions be consuming design tokens instead of hard coded? If so, can we log an issue?

```suggestion
      props.theme.global.borderSize['small']
```

jcfilben: Oops yes we should use borderSize, just pushed up that change. Here is an updated screenshot:

Screenshot 2025-10-30 at 3 43 14 PM

> Also, shouldn't these dimensions be consuming design tokens instead of hard coded? If so, can we log an issue?

Yes, I'll file a ticket for this

jcfilben: Ticket created https://github.com/grommet/grommet-theme-hpe/issues/558

halocline (approved):

PR #5583, opened 2025-11-24
upgrade Storybook v10

britt6612:

#### What does this PR do?
This PR updates our current storybook from version 8 to use version 10.

This PR also changes the format of the stories themselves, our old format was the CSF2 however for version 10 of storybook they recommend using the CSF3 format, which the `meta` object is reuqired in the CSF3. More info can be found [here](https://storybook.js.org/docs/8/api/csf#upgrading-from-csf-2-to-csf-3) 

Additionally, we had to rename Storybook configuration files from .js to .mjs because Storybook 10 requires ESM-only builds. The .mjs extension explicitly tells Node.js these are ES modules, which is necessary for Storybook's new ESM-only architecture.
upgrades storybook v10 on dev board
package.json
run storybook locally 
I needed to change imports in some example files that were pointing to `ContentPane`.

This caused a "process is not defined" error in Storybook because importing from the barrel export (layouts/index.js) loads the entire folder—including files like `DocsPageHeader.js` that import `next/link`. Since Storybook runs in the browser this breaks the build.

I had to change to use the direct import of the ContentPane. 
This imports only the ContentPane component, avoiding the Next.js dependencies.

Previously this was working fine in v8 because 
Storybook v8:
- Had built-in polyfills for Node.js globals like process
- Automatically shimmed process.env
- More forgiving with Next.js code in the browser

Storybook v10 
- Removed automatic Node.js polyfills (cleaner bundle)
- Stricter webpack configuration
-  Requires explicit handling of Node.js modules
-  Forces better separation of browser vs server code

Renamed: .storybook/main.js → .storybook/main.mjs
Renamed: .storybook/preview.js → preview.mjs
Why: Storybook 10 is ESM-only and requires the .mjs extension for explicit ES module recognition by Node.js

none
no
backwards compatible

### DIFF
--- a/.github/workflows/sync-figma-to-tokens.yml
+++ b/.github/workflows/sync-figma-to-tokens.yml
@@ -30,11 +30,11 @@ jobs:
       - name: Clone repo
         uses: actions/checkout@v3
 
-      - name: Upgrade stable branch dependencies
-        run: pnpm upgrade grommet && pnpm upgrade grommet-icons
-
       - name: Install dependencies
         run: pnpm install
+
+      - name: Upgrade stable branch dependencies
+        run: pnpm upgrade grommet && pnpm upgrade grommet-icons
 
       - name: Sync variables in Figma file to tokens
         run: pnpm sync-figma-to-tokens -- --output tokens
--- a/aries-site/src/examples/foundation/content-container-sizes/ContainerSizes.jsx
+++ b/aries-site/src/examples/foundation/content-container-sizes/ContainerSizes.jsx
@@ -20,7 +20,10 @@ export const ContainerSizes = () => {
         flex
         overflow="auto"
       >
-        <Box flex /> {/* to center the container, which justify content properly when overflowing */}
+        {/*
+          center the container, which justify content properly when overflowing
+        */}
+        <Box flex /> 
         <Box
           height={{ min: height, max: height }}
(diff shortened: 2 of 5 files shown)


aries-core/.storybook/main.js
jcfilben: Getting this error in the console when I run storybook:

Screenshot 2025-12-02 at 3 30 02 PM

britt6612: hmm let me check on my side

aries-core/src/stories/navigation/navigation.stories.tsx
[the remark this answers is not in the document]
britt6612: opps yes Ill put them back

aries-core/.storybook/main.mjs
britt6612: @jcfilben 

Storybook 10 requires ESM-only builds, which means all configuration files must be proper ES modules. 

What Changed in Our Setup
File renames required:

```
.storybook/main.js → .storybook/main.mjs
.storybook/preview.js → .storybook/preview.mjs
```

jcfilben (approved):
Looks good!

britt6612:
> Looks good.
> 
> Only thing I noticed is the images in "Inline Notifications -> Promotion" are broken

since this is the same behavior as now merging then we can look at the stories when we move things around

Issue #5596, opened 2025-12-02
Move aries-core -> shared


MikeKingdom:
### What does this request relate to?

docs

### Describe the feature you'd like to request.

Move the `aries-core` directory into `shared/aries-core.

- [ ] update `.husky/pre-commit`
- [ ] update aries-site refs in package.json

### How urgent is this request?

None

PR #5679, opened 2026-01-08
[repo-restructure] move aries-core to shared

Sulaymon333:
**Move aries-core directory into shared/aries-core.**

**Overview**
Relocates the `aries-core` package from the repository root to `shared/aries-core`, consolidating shared packages and improving workspace organization

**Changes Made**
Directory & Package Structure

- Moved `aries-core` → `shared/aries-core`
- Renamed package from `aries-core` to `@shared/aries-core`
- Updated pnpm-workspace.yaml to remove explicit entry (covered by shared/* pattern)

**Package Dependencies**
- Added @shared/hooks workspace dependency to package.json
- Added @types/node devDependency to support build scripts
- Updated both @shared/aries-core and hpe-design-tokens to use @types/node from catalog for build script type support
- Updated aries-site dependency from aries-core: workspace:* to @shared/aries-core: workspace:*

**Configuration Updates**
- Fixed relative path in package.json build script (../aries-site → ../../aries-site)
- Updated next.config.mjs to transpile @shared/aries-core
- Updated pre-commit hook to reference @shared/aries-core
- Updated root package.json storybook script for @shared/aries-core

**Import Updates**

- Updated .js files imports across aries-site from 'aries-core' to '@shared/aries-core'
- Updated direct path imports to use scoped package name ('@shared/aries-core/src/js/...')

closes https://github.com/grommet/hpe-design-system/issues/5596

### DIFF
--- a/aries-site/src/examples/components/header/HeaderNavigationExample.js
+++ b/aries-site/src/examples/components/header/HeaderNavigationExample.js
@@ -1,7 +1,6 @@
 import React, { useContext } from 'react';
 import { Button, Header, Menu, Nav, ResponsiveContext } from 'grommet';
-// eslint-disable-next-line max-len
-import { AppIdentity } from '@shared/aries-core/src/js/components/helpers/AppIdentity';
+import { AppIdentity } from '@shared/aries-core';
 
 const items = [
   { label: 'Label 1' },


aries-site/src/examples/components/header/HeaderNavigationExample.js
```diff
@@ -1,6 +1,7 @@
 import React, { useContext } from 'react';
 import { Button, Header, Menu, Nav, ResponsiveContext } from 'grommet';
-import { AppIdentity } from 'aries-core/src/js/components/helpers/AppIdentity';
+// eslint-disable-next-line max-len
+import { AppIdentity } from '@shared/aries-core/src/js/components/helpers/AppIdentity';
```
MikeKingdom: I'm surprised this works, but either way we shouldn't be reaching down into the src level. Instead, we need to make sure AppIdentity is exported so this can instead be
```jsx
import { AppIdentity } from '@shared/aries-core`;
```

MikeKingdom (commented):
Looking pretty good. Just a couple of comments. I still need to test it.

BTW, soon after this goes in I'll move all the `examples` to `aries-core` to break the cyclic dependency between `aries-core` and `aries-site`.

MikeKingdom (approved):
Looks good. Tested and works fine (aries-site, aries-core storybook, etc.)

Once merged the netlify config for the aries-core storybook needs to be tweaked to use the new location. This is what's failing in the "checks" for this PR but we can't change that config until it's merged.

britt6612 (approved):

PR #5690, opened 2026-01-13
Fix circular dependency when importing @hpe-design/icons-grommet

### LINKED ISSUE #5623: Importing icon in ariesTheme causes error
When trying to change the import icon from:
```
import { Info } from 'grommet-icons';
```

to: 

```
import { Info } from '@hpe-design/icons-grommet';
```

It fails with the following error 

Image

britt6612:
[Deploy Preview](https://deploy-preview-5690--keen-mayer-a86c8b.netlify.app/)

#### What does this PR do?
Resolves a circular dependency that prevented using `@hpe-design/icons-grommet` in the site theme, which previously forced us to rely on the legacy `grommet-icons` package.

The Problem:
When importing icons from `@hpe-design/icons-grommet`in `aries.js the site would crash with:
```
Cannot read properties of undefined (reading 'palettes')
```

Solution:
Break the circular dependency by having `color.js` import directly from `grommet-theme-hpe` instead of the custom Aries theme.

This is safe because:
- The Aries theme only adds UI-specific customizations (e.g. feedback, buttonGroup)
- It does not override any color, elevation, or layer values used by `color.js`
- Color examples render the same values from `grommet-theme-hpe` as they do from Aries

Changes:
- `color.js`: Updated import from aries → grommet-theme-hpe
- `aries.js`: Switched from grommet-icons → @hpe-design/icons-grommet

closes #5623 
color.js
aries.js
- Run the site locally and verify it loads without crashing
- Confirm color examples render as expected
This was caused by a circular dependency:
ColorPalettes.js → imports colorExamples from index.js index.js → exports from both color.js and structure.js color.js → imports aries theme aries.js → imports Info icon from @hpe-design/icons-grommet structure.js → also imports icons from @hpe-design/icons-grommet Circular dependency → module loading deadlock → colorExamples becomes undefined
no
yes


MikeKingdom (commented):
Given how the color stuff is used couldn't it just be derived from the ThemeContext value in ColorPallettes.js? Hardwiring it to a theme instance is a bit fragile

Issue #5674, opened 2026-01-08
Menu - PR creating and approval


julauxen:
Goal: create a PR for all the changes agreed upon on #5667 and seek approval.

PR #5702, opened 2026-01-19
Menu content update

julauxen:
[Deploy Preview](https://deploy-preview-5702--keen-mayer-a86c8b.netlify.app/)
Actual page: https://deploy-preview-5702--keen-mayer-a86c8b.netlify.app/components/menu

#### What does this PR do?
Updates the menu content according to [this design.](https://www.figma.com/design/gkAlhfnBI1OrqAB2sVHObp/Content-audit?node-id=463-730&t=utpqMg2ZiD3FVMNn-1) 

Closes [#5674](https://github.com/grommet/hpe-design-system/issues/5674)

[Menu page](https://deploy-preview-5702--keen-mayer-a86c8b.netlify.app/components/menu)

### DIFF
--- a/apps/docs/src/examples/components/menu/MenuRecordActionsExample.js
+++ b/apps/docs/src/examples/components/menu/MenuRecordActionsExample.js
@@ -48,11 +48,18 @@ export const MenuRecordActionsExample = () => {
         <List
           data={serverGroups}
           defaultItemProps={{
-            pad: { vertical: 'xsmall' },
+            pad: { vertical: '4xsmall' },
           }}
         >
           {(datum, index) => (
-            <Box direction="row" justify="between" align="start">
+            <Box
+              direction="row"
+              justify="between"
+              align="start"
+              background="background-front"
+              pad="small"
+              round="small"
+            >
               <Box gap="3xsmall" fill>
                 <TextEmphasis>{datum.name}</TextEmphasis>
                 <Box gap="5xsmall">


luketa8:
@julauxen Added a first pass at the anatomy diagram. Let's sync tomorrow to move forward.

apps/docs/src/data/structures/components.js
```diff
@@ -864,7 +864,7 @@ export const components = [
     accessibility: 'Passed WCAG 2.2 AA',
     category: 'Controls',
     description:
-      'Menu is a component that contains a list of actions. When a user clicks an item in the menu, the menu closes and the action is executed.',
+      'A menu presents a set of related actions for a user to choose from.',
```
halocline: Trying to avoid ending a sentence with a preposition.

```suggestion
      'A menu presents a set of related actions from which a user may choose.',
```

apps/docs/src/pages/components/menu.mdx
```diff
@@ -27,102 +32,197 @@ import {
...
+| **3a** | **Item label**     | Communicates the specific action to the user.                        | <span><Checkmark a11yTitle="true" size="small" /></span> | --                                                                                                                                                                                  |
+| **3b** | **Item icon**      | Provides a quick visual reference for the action.                    |                         Optional                         | --                                                                                                                                                                                  |
+| **4**  | **Divider**        | Separates groups of related actions.                                 |                         Optional                         | --                                                                                                                                                                                  |
+
+## Types
 
-To create a predictable experience for users, the following are guidelines for using Menu.
+### Default
```
halocline: Where did we land on showing a menu other than the default button treatment? If I were to consume this page, I would not pick up on that menus can use any button kind styles.

Can we repurpose the "Toolbar" type and show a variety of menu kinds?

Also, I feel like folks are reaching for `secondary` instead of `toolbar` when adding menus to toolbars because the menu doesn't have enough affordance and users are not seeing it. Should we consider modifying the `toolbar` treatment.

cc: @oliverHPE

halocline: Given where we are in the sprint, I think that is fine. If we were mid-sprint, I'd say add now because we are already in that headspace and its cheap and easy to implement.

Issue #5710, opened 2026-01-22
Design System Storybook MVP


britt6612:
- [x] For every Design System component create one story that has controls. 
        One default story per component
        Controls enabled for relevant props
- [x] Group new stories as components.
- [x] Group current stories as patterns.

seanhillweb:
I'll work on a Button Story.

seanhillweb:
I'll work on the Menu story next.

britt6612:
### Done
Box
Anchor
Grid
Card
Footer
Header
Layer
Main
Page
PageHeader


### InProgress:
Accordion 
Menu
Button

### TODO:
Pagination 
Search 
Tabs
Tip
ToggleGroup
CheckBox
CheckBoxGroup
DateInput
FileInput
MaskedInput
RadioButtonGroup
RangeInput
Select 
SelectMultiple
TextArea
TextInput
Data & Friends
Avatar 
DataTable
NameValueList
Notification
Skeleton
Spinner
Tag

britt6612:
Every component is listed within the storybook. This was MVP
https://deploy-preview-5822--unrivaled-bublanina-3a9bae.netlify.app/?path=/story/components-accordion--default

will create any follow on tickets after storybook review

PR #5754, opened 2026-02-03
DS Storybook: Button Story

seanhillweb:
[Deploy Preview](https://deploy-preview-5754--unrivaled-bublanina-3a9bae.netlify.app/?path=/story/components-button--primary)

#### What does this PR do?
Added initial scaffolding for Button story.

### DIFF
--- a/shared/aries-core/src/stories/components/Button.stories.tsx
+++ b/shared/aries-core/src/stories/components/Button.stories.tsx
@@ -1,32 +1,32 @@
 import React from 'react';
 import type { Meta, StoryObj } from '@storybook/react';
 import { Button } from 'grommet';
-import { tShirtSizes, iconArg } from '../utils/commonArgs';
+import {
+  a11yTitleArg,
+  disabledArg,
+  fillArg,
+  iconArg,
+  labelArg,
+  tShirtSizes,
+} from '../utils/commonArgs';
 
 const meta = {
   title: 'Components/Button',
   component: Button,
-  // TODO: Verify if 'centered' is appropriate for Button stories, currently does not work as expected due to theme overrides
+  // TODO: Currently does not work as expected due to theme overrides, when supported, remove comments
   // parameters: {
   //   layout: 'centered',
   // },
   argTypes: {
-    a11yTitle: {
-      control: { type: 'text' },
-    },
+    a11yTitle: a11yTitleArg,
     active: {
       control: { type: 'boolean' },
     },
     busy: {
       control: { type: 'boolean' },
     },
-    disabled: {
-      control: { type: 'boolean' },
-    },
-    fill: {
-      control: { type: 'boolean' },
-      options: [true, false, 'horizontal', 'vertical'],
-    },
+    disabled: disabledArg,
+    fill: fillArg,
     href: {
       control: { type: 'text' },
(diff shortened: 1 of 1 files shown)


shared/aries-core/src/stories/components/Button.stories.tsx
```diff
@@ -0,0 +1,84 @@
...
+import { Button } from 'grommet';
+import { tShirtSizes, iconArg } from '../utils/commonArgs';
+
+const meta = {
+  title: 'Components/Button',
+  component: Button,
+  // TODO: Verify if 'centered' is appropriate for Button stories, currently does not work as expected due to theme overrides
+  // parameters: {
+  //   layout: 'centered',
+  // },
```
jcfilben: For the MVP version of this storybook let's not worry about centering for now. We can leave things aligned to the top left corner and then revisit centering later

jcfilben (approved):

shared/aries-core/src/stories/components/Button.stories.tsx
```diff
@@ -0,0 +1,84 @@
...
+    active: {
+      control: { type: 'boolean' },
+    },
+    busy: {
+      control: { type: 'boolean' },
+    },
+    disabled: {
+      control: { type: 'boolean' },
+    },
+    fill: {
```
britt6612: @jcfilben 

Im wondering about `fill` we usually dont use this prop on Button

shared/aries-core/src/stories/components/Button.stories.tsx
```diff
@@ -0,0 +1,84 @@
...
+import { tShirtSizes, iconArg } from '../utils/commonArgs';
+
+const meta = {
+  title: 'Components/Button',
+  component: Button,
+  // TODO: Verify if 'centered' is appropriate for Button stories, currently does not work as expected due to theme overrides
+  // parameters: {
+  //   layout: 'centered',
+  // },
+  argTypes: {
```
jcfilben: Actually one more note, can we reach for commonArgs for some of these?

shared/aries-core/src/stories/components/Button.stories.tsx
seanhillweb: Following up on this, I saw `fillArgs` listed in the `commonArgs` file and decided to follow that pattern. It would still include true, false, and vertical - but feels like something to cleanup after MVP?

shared/aries-core/src/stories/components/Button.stories.tsx
seanhillweb: I'm adding matching args I see listed in commonArgs.

seanhillweb: One more thought, some args like `active` or `busy` are just a boolean, and I wasn't sure if that should be included in commonArgs. I've left those as static within the story.

shared/aries-core/src/stories/components/Button.stories.tsx
jcfilben: I think it makes sense to use `fillArgs` from `commonArgs` here. After MVP I think it is worth checking with the designers to see what options they envision being used for hpe apps.

PR #5786, opened 2026-02-06
DS storybook - ToggleGroup story

jcfilben:
https://deploy-preview-5786--unrivaled-bublanina-3a9bae.netlify.app/?path=/story/components-togglegroup--default

#### What does this PR do?
Add story for togglegroup component

### DIFF
--- a/shared/aries-core/src/stories/components/ToggleGroup.stories.tsx
+++ b/shared/aries-core/src/stories/components/ToggleGroup.stories.tsx
@@ -7,7 +7,7 @@ const meta = {
   component: ToggleGroup,
   argTypes: {
     multiple: {
-      control: { type: 'object' },
+      control: { type: 'boolean' },
     },
     options: {
       control: { type: 'object' },


Sulaymon333 (changes requested):
Thanks for the PR! 
I would suggest tightening the Storybook typing to prevent several issues: type narrowing, preserve inferred precision for better intellisense, invalid keys - catching of bogus props during compile time, prop drift and so on. 

**Benefits of using Meta + satisfies + StoryObj<typeof meta>:**
- Type narrowing & precision: `satisfies` validates without widening types, preserving inference for better intellisense
- Compile-time safety: Catch invalid props in argTypes and args during type-checking, not at runtime
- Refactor safety: Any changes to ToggleGroup props immediately surface type errors in meta and story args, preventing silent drift

**Suggested changes:**

1. Type meta as satisfies Meta<typeof ToggleGroup> (validates against component props)
2. Remove redundant `ToggleGroupExtendedProps` annotation from render
3. Use satisfies Story on the story object (validates without widening type inference)

This aligns with [Storybook's TypeScript best practices](https://storybook.js.org/docs/writing-stories/typescript), which recommend native type inference via satisfies.

I hope we can apply the same approach with other similar PRs.

shared/aries-core/src/stories/components/ToggleGroup.stories.tsx
```diff
@@ -0,0 +1,34 @@
+import React from 'react';
+import type { Meta, StoryObj } from '@storybook/react';
+import { ToggleGroup } from 'grommet';
+
+const meta = {
+  title: 'Components/ToggleGroup',
+  component: ToggleGroup,
+  argTypes: {
```
britt6612: what about `defaultValue` prop

jcfilben: Couldn't get this to work for both single and multiple selection because I needed to support both string and array types

shared/aries-core/src/stories/components/ToggleGroup.stories.tsx
```diff
@@ -0,0 +1,34 @@
...
+    multiple: {
+      control: { type: 'object' },
```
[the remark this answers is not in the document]
jcfilben: Oops yes this should be boolean

PR #5815, opened 2026-02-13
DS Storybook - Add DataTableColumns story

britt6612:

#### What does this PR do?
adds  DataTableColumns story


shared/aries-core/src/stories/components/DataTableColumns.stories.tsx
```diff
@@ -0,0 +1,76 @@
...
+          {
+            property: 'paid',
+            label: 'Paid',
+          },
+        ]}
+      />
+    </Data>
+  ),
+  args: {
+    drop: true,
```
jcfilben: The drop view is a bit odd since it tried to stretch across the screen. Wondering if we set this to false by default

jcfilben: Agree with keeping things lean for MVP, we can revisit later if needed

Sulaymon333 (approved):

PR #5838, opened 2026-02-20
DS Storybook - Add DataTable story

britt6612:

#### What does this PR do?
adds datatable example
related to https://github.com/grommet/hpe-design-system/issues/5710


seanhillweb (approved):
Removed comment.

Sulaymon333:
Should we consider adding `%` to the "Complete" column? So that the table is more readable and it is in line with what we would want to recommend as guidance
image

There is a similar example on DS site on [DataTable page](https://design-system.hpe.design/components/datatable)
image

PR #5841, opened 2026-02-21
[Monorepo] Copilot instructions - Added initial instructions

halocline:
#### What does this PR do?

This pull request adds detailed Copilot instructions for the HPE Design System monorepo. The new `.github/copilot-instructions.md` file provides comprehensive documentation on the project structure, development workflows, design token architecture, UI conventions, and pull request guidelines, making it easier for contributors and Copilot to understand and work within the repo.

Documentation additions:

* Added `.github/copilot-instructions.md` with an overview of the monorepo structure, listing key workspace roots and packages.
* Documented essential development commands, including workspace builds, Storybook launches, and test execution.
* Provided step-by-step instructions for adding component pages to the docs site, including file locations and required patterns.

Design token and UI conventions:

* Explained the design token architecture, build process, and Figma sync commands, covering token layers and variant handling.
* Outlined UI framework conventions, theming, icon usage, and static export configuration for the docs site.

Pull request and versioning guidelines:

* Clarified branch naming, PR title/description formatting, and linking issues in PRs.
* Detailed versioning and publishing processes using Changesets and workspace catalogs.

Key file locations:

* Listed important

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -15,7 +15,8 @@ pnpm workspace with four workspace roots (`apps/*`, `packages/**`, `sandbox/*`, 
 | `shared/aries-core`          | `@shared/aries-core`         | Shared React components + Storybook           |
 | `shared/hooks`               | `@shared/hooks`              | Shared React hooks (TypeScript, Vitest)       |
 | `sandbox/grommet-app`        | —                            | Prototype app for testing components/tokens   |
-| `sandbox/mcp-ui`             | —                            | Future development — do not treat as active   |
+| `sandbox/native-web`         | —                            | Prototype sandbox app (native-web)            |
+| `sandbox/tailwind-app`       | —                            | Prototype sandbox app (Tailwind)              |
 
 Shared dependency versions are managed through `pnpm-workspace.yaml` `catalog:` entries — use `catalog:` references in `package.json` instead of pinned versions for shared deps like `grommet`, `react`, `styled-components`.
 
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -2,7 +2,7 @@
 
 ## Monorepo Overview
 
-pnpm workspace with five workspace roots (`apps/*`, `packages/**`, `sandbox/*`, `shared/*`). Key packages:
+pnpm workspace with four workspace roots (`apps/*`, `packages/**`, `sandbox/*`, `shared/*`). Key packages:
 
 | Path                         | Package                      | Purpose                                       |
 | ---------------------------- | ---------------------------- | --------------------------------------------- |
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -9,6 +9,7 @@ pnpm workspace with five workspace roots (`apps/*`, `packages/**`, `sandbox/*`, 
 | `apps/docs`                  | `docs`                       | Next.js 15 documentation site (static export) |
 | `packages/hpe-design-tokens` | `hpe-design-tokens`          | Design tokens built with Style Dictionary v4  |
 | `packages/icons-grommet`     | `@hpe-design/icons-grommet`  | HPE icons for Grommet (Vite build)            |
+| `packages/icons-svg`         | `@hpe-design/icons-svg`      | HPE icons in raw SVG format (Vite build)      |
 | `packages/codemods`          | `hpe-design-system-codemods` | JSCodeshift transforms for migrations         |
 | `shared/aries-core`          | `@shared/aries-core`         | Shared React components + Storybook           |
 | `shared/hooks`               | `@shared/hooks`              | Shared React hooks (TypeScript, Vitest)       |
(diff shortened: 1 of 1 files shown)


Sulaymon333:
Looks good to me with solid foundation and coverage.

> This looks so far so good as a start Im sure we can expand on it as we build more with it. couple things that came to mind that we dont need to address now was maybe adding any formatting rules. Maybe for TS avoid using `any` unless unavoidable, Prefer functional components with arrow syntax.

Building on the comment on TypeScript `any` usage, our current codebase has usage of `any` in some places like `tokens`, some `stories` codes, etc. So, it might be almost unavoidable in some cases. So, I think one area that would really benefit from explicit guidance for copilot is building on the TypeScript conventions for `shared/hooks` so that Copilot can avoid generating wildly inconsistent TS without such model example since the existing hooks in this repo follow a clear pattern that's worth documenting for Copilot to follow in the `copilot-instructions.md` file.

Looking at the current setup for `shared/hooks`: `tsconfig.json` enables both `strict` and `noUncheckedIndexedAccess`, and the lint script is `eslint . --max-warnings 0` with `tseslint.configs.recommended` extended - which means any already fails CI in shared/hooks. That's the de facto rule, but Copilot has no way to know that without us saying so.

The more useful guidance is _what to reach for instead of_ `any`, and the existing hooks already model it. So, we can add a section covering Typescript Conventions, see below:

``````md
## TypeScript Conventions (`shared/hooks`)

`shared/hooks` is strict (`strict: true`, `noUncheckedIndexedAccess: true`) and lints with `--max-warnings 0`, so `any` will fail CI. When generating hooks, follow the patterns already established in `useLocalStorage`, `useSessionStorage`, and `useInert`:

- **For caller-supplied shapes, use generics with sensible defaults**, not `any`:
```ts
  // ✅ matches existing pattern in useLocalStorage / useSessionStorage
  export const useLocalStorage = <T>(key: string, initialValue: T) => { ... }

  // ✅ matches existing pattern in useInert
  export const useInert = <T extends HTMLElement = HTMLDivElement>(): RefObject<T | null> => { ... }
```
- **For dynamic values (e.g. `JSON.parse` results)**, narrow into the generic `T` rather than typing as `any` — see how `useLocalStorage` round-trips through `JSON.parse` while staying generic.
- **Return tuples as `as const`** so callers get the precise `[value, setter]` shape.
- If `any` is genuinely unavoidable in a narrow case, prefer `unknown` first and narrow with a type guard; reach for `any` only with an inline `// eslint-disable-next-line` and a comment explaining why.
``````

The robust convention above goes beyond the "use generics, not `any`" rule alone — which on its own probably covers ~90% of cases — and helps us deliver type-safe code consistently. Being intentional about what our TypeScript conventions look like makes it clear to Copilot how to handle the edges where it would otherwise reach for `any` (e.g. `JSON.parse` results, tuple returns, and the rare genuine escape hatch). Grounding the guidance in the existing hooks also makes it self-reinforcing as the codebase grows.

File: .github/copilot-instructions.md [hunk 1]
```diff
@@ -0,0 +1,107 @@
...
+```
+
+## UI Framework Conventions
+
+- **Always use Grommet components** (`Box`, `Button`, `Text`, etc. from `grommet`) — not custom HTML elements.
+- **Icons**: Use `@hpe-design/icons-grommet`, not `grommet-icons`. Run `npx hpe-design-system-codemods migrate-grommet-icons-to-hpe <path>` to migrate.
+- **Theming**: Extend `hpe` theme from `grommet-theme-hpe` via `deepMerge(hpe, {...})`. See `apps/docs/src/themes/aries.js`.
+- **Dark mode**: Implemented via `ThemeMode` component (`apps/docs/src/layouts/main/ThemeMode.js`); token files have separate `.light.json`/`.dark.json` variants.
+- Docs site uses `output: 'export'` (static HTML) in `apps/docs/next.config.mjs` — no server-side rendering at runtime.
+
```
Sulaymon333, line 75: we can consider adding a new sibling section right after this section

``````md
## TypeScript Conventions (`shared/hooks`)

`shared/hooks` is strict (`strict: true`, `noUncheckedIndexedAccess: true`) and lints with `--max-warnings 0`, so `any` will fail CI. When generating hooks, follow the patterns already established in `useLocalStorage`, `useSessionStorage`, and `useInert`:

- **For caller-supplied shapes, use generics with sensible defaults**, not `any`:
```ts
  // ✅ matches existing pattern in useLocalStorage / useSessionStorage
  export const useLocalStorage = <T>(key: string, initialValue: T) => { ... }

  // ✅ matches existing pattern in useInert
  export const useInert = <T extends HTMLElement = HTMLDivElement>(): RefObject<T | null> => { ... }
```
- **For dynamic values (e.g. `JSON.parse` results)**, narrow into the generic `T` rather than typing as `any` — see how `useLocalStorage` round-trips through `JSON.parse` while staying generic.
- **Return tuples as `as const`** so callers get the precise `[value, setter]` shape.
- If `any` is genuinely unavoidable in a narrow case, prefer `unknown` first and narrow with a type guard; reach for `any` only with an inline `// eslint-disable-next-line` and a comment explaining why.
``````

Issue #5760, opened 2026-02-04
center storybook stories


britt6612:
Explore if there are default options within storybook that we can add padding around examples or center them.

PR #5847, opened 2026-02-23
DS - storybook center stories & group patterns

britt6612:

#### What does this PR do?
This PR fixes the stories so they are centered and groups our pre- existing patterns into the patterns 
closes https://github.com/grommet/hpe-design-system/issues/5760
related to https://github.com/grommet/hpe-design-system/issues/5710

### DIFF
--- a/shared/aries-core/.storybook/preview.mjs
+++ b/shared/aries-core/.storybook/preview.mjs
@@ -17,7 +17,7 @@ export default {
           themeMode={mode}
           background={context.globals.background?.value}
         >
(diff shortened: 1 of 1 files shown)


britt6612:
> Nice change, I'm wondering if there is a way we can opt out of this for certain stories. For example the Header story looks a bit squished and the Tabs story jumps around when I switch tabs
> 
>  Screen.Recording.2026-02-23.at.4.46.36.PM.mov

yes I believe there is a way we can make a list of components then I can opt those out

PR #5853, opened 2026-02-24
Poc playground audit

SeamusLeonardHPE:
Playground UX test

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -75,4 +75,12 @@ module.exports = {
     quotes: [2, 'single', { avoidEscape: true }],
     'arrow-body-style': 0,
   },
+  overrides: [
+    {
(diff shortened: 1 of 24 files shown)


aries-site/src/examples/components/anchor/AnchorPlayground.js
```diff
@@ -0,0 +1,145 @@
+/* eslint-disable react/prop-types, max-len */
+import React from 'react';
+import { Anchor, Box } from 'grommet';
+import {
+  anchorArgTypes,
+  Default as AnchorDefault,
+} from '@shared/aries-core/src/stories/components/Anchor.stories';
```
MikeKingdom: Should not be reaching into the file structure of a separate package.

Issue #5864, opened 2026-02-27
DS Storybook - accessibility violations


jcfilben:
There are accessibility violations on the following stories:

- [x] Data/Data
- [x] Dashboards/Dashboard
- [x] Dashboards/Three Column
- [x] Dashboards/Two Column
- [x] Empty State/Card
- [x] Forms/Shipping
- [x] Inline Notifications/Page Banner
- [x] Lists/Screen

britt6612:
all fixes are complete

PR #5892, opened 2026-03-06
DS Storybook - fix accessibility issue on empty state story

britt6612:

#### What does this PR do?
fixes the empty state accessibility issue
related to #5864

### DIFF
--- a/apps/docs/src/examples/templates/empty-state/CardEmptyState.js
+++ b/apps/docs/src/examples/templates/empty-state/CardEmptyState.js
@@ -1,12 +1,5 @@
 import React from 'react';
-import {
-  Box,
-  // Main,
-  Grid,
-  Page,
-  PageContent,
-  PageHeader,
-} from 'grommet';
+import { Box, Main, Grid, Page, PageContent, PageHeader } from 'grommet';
 import { Card } from '../Card';
 import { ErrorManagementEmptyState } from './ErrorManagementEmptyState';
 import { CostByMonth, CostByYear } from '../dashboards';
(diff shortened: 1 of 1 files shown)


apps/docs/src/examples/templates/empty-state/CardEmptyState.js
```diff
@@ -16,19 +9,17 @@ export const CardEmptyState = () => {
...
-          <CostByMonth period="Last Year" />
-          <CostByYear period="Lifetime" />
-          <Card align="center" title="Cost by service" level={2}>
-            <Box pad={{ vertical: 'medium' }}>
-              <ErrorManagementEmptyState level={3} />
-            </Box>
-          </Card>
-        </Grid>
-        {/* </Main> */}
+        <Main>
```
halocline: I'm not sure this is the right placement for main... Shouldn't main include the H1? I would expect Main to have Page as a child.

Related thread: https://hpe.slack.com/archives/C04LMJWQT/p1771018413935409

halocline: Modify `<Page pad={{ bottom: '3xlarge' }} />` so that content has room to breathe.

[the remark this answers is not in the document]
britt6612: which one would we recommend? In the site the background that is used is grey so this example looks different. Not sure if we will break site example if we change the `background` or `elevation` but for my knowledge do we prefer one over the other?

Image

seanhillweb: Ah, I was unaware of this pattern. I saw other pull requests that making the same mistake. @britt6612 Would you like some assistance resolving?

PR #5897, opened 2026-03-06
DS Storybook - fix popover accessibility

britt6612:
[Deploy Preview](https://deploy-preview-5897--unrivaled-bublanina-3a9bae.netlify.app/?path=/story/patterns-popover--popover-inline)

#### What does this PR do?
This pull request makes small accessibility improvements to the popover examples by ensuring that the `aria-controls` attribute is only set when the popover is visible. This prevents referencing non-existent elements in the accessibility tree.
related to https://github.com/grommet/hpe-design-system/issues/5864
popover stories 
Updated the `aria-controls` attribute in both `PopoverInlineExample` and `PopoverSimpleExample` components to conditionally reference the popover element only when it is shown.

### DIFF
--- a/shared/aries-core/src/stories/popover.stories.js
+++ b/shared/aries-core/src/stories/popover.stories.js
@@ -15,6 +15,9 @@ export default meta;
 
 export const PopoverInline = {
   render: () => <PopoverInlineExample />,
+  parameters: {
+    background: 'background-back',
+  },
 };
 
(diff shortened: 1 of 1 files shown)


britt6612:
putting this to draft just noticed the cards need background here too 

Screenshot 2026-03-06 at 2 54 47 PM

Issue #5867, opened 2026-02-27
DS storybook team review and gather feedback


jcfilben:
Before we announce the DS storybook review with the team and do a final polish.

Storybook available here: https://unrivaled-bublanina-3a9bae.netlify.app/?path=/story/components-accordion--default

Focus review on components stories. 

Review areas:
Do we have the correct set of props exposed, are these all items that we are comfortable with DS users adjusting?
Are the examples DS quality (aka things we are okay with being copy/pasted and referenced within HPE) both from a design perspective and from a code perspective?
How is the experience overall using the control panel?

MikeKingdom:
It's looking pretty good. I still need to spend more time on it. On thing I've run into for a while is it flashes light then dark for each story when your system default is dark mode.

jcfilben:
Thanks for the feedback @oliverHPE 
It is possible to have controls for Selector and Popover but at the moment for MVP we were just going to expose controls for the component stories. So of the pattern stories get more complex and are not something we want to support controls on, but I could see the value in having controls on Selector and Popover. Something we could revisit.

PR #5903, opened 2026-03-09
DS Storybook - incorporate feedback from team review part 2

jcfilben:
#### What does this PR do?
Address more feedback items from the team review https://github.com/grommet/hpe-design-system/issues/5867

- Fixed the button so that it doesn’t stretch across the page in DataTableColumns
- Set default state for booleans
- Alphabetize args
- Fixed Menu story so menu doesn’t stretch across the screen
- Button, added badge and tip prop

### DIFF
--- a/shared/aries-core/src/stories/components/Button.stories.tsx
+++ b/shared/aries-core/src/stories/components/Button.stories.tsx
@@ -52,7 +52,7 @@ const meta = {
     },
     tip: {
       control: { type: 'text' },
-    }
+    },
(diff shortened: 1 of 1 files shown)


Sulaymon333 (commented):
The PR looks good to me except for the Box component with a lot of pros that might lead to high cognitive load. Using Storybook native Category feature for component with long props? That will definitely help with reducing cognitive load IMHO. See possible preview and implementation below.

PR #5919, opened 2026-03-13
Eslint centralization/modernization

### LINKED ISSUE #5604: Consolidate eslint versions and config
Currently the `eslint` config is somewhat distributed among the various workspaces in the hpe-design-system repo and there are a couple different versions (one fairly old).

The alternative is to consider switching to Biomejs

- [ ] Move eslint config to the root directory
- [ ] Adjust any dependencies for eslint in package.json's to the root `pnpm-workspace.yaml` dependency catalogs

Sulaymon333:

#### What does this PR do?
Eslint centralization/modernization

closes #5604


MikeKingdom (commented):
Good progress. There does seem to be some missing config.

In some directories lint isn't really working or at least the rules aren't correct.

In tailwind-app, if I add an unused variable in App.jsx and run `pnpm lint` eslint runs and doesn't identify any issues. If I do something similar in a .js file it does complain like it should. Same with sandbox/grommet-app.

Similarly, in aries-core .tsx files don't seem to be linted.

packages/icons-grommet does seem to support lint on .ts and .jsx.

In shared/hooks it seems misconfigured:
```bash
$ pnpm lint

> @shared/hooks@0.0.0 lint C:\git\grommet\hpe-design-system3\shared\hooks
> eslint . --max-warnings 0


Oops! Something went wrong! :(

ESLint: 9.39.1

You are linting ".", but all of the files matching the glob pattern "." are ignored.
```

apps/design-tokens-manager seems to need some more config:

Image

apps/docs seems to be linting although in vscode it doesn't highlight the problem lines.  (It does for other files with eslint warnings)

Sulaymon333:
> apps/docs seems to be linting although in vscode it doesn't highlight the problem lines. (It does for other files with eslint warnings) 

Which of the files had the issue. Sometimes, I notice VSCode delays linting and won't highlight issues especially after changing eslint config settings. I simply reload it and it starts linting again.

shared/hooks/package.json
```diff
@@ -20,7 +20,7 @@
     "jsdom": "^26.1.0",
     "minimatch": "^10.0.3",
```
britt6612, line 27: do we need to have a local eslint file as well for the hooks? similar to the other folders.

eslint.config.mjs
```diff
@@ -0,0 +1,144 @@
...
+        'single',
+        {
+          avoidEscape: true,
+        },
+      ],
+
+      'arrow-body-style': 0,
+    },
+  },
+  // TypeScript recommended rules — scoped to TS files only
```
britt6612, line 127: Is the TS linting going to be addressed in a follow-up? I tested adding an unused variable to a .tsx file in aries-core and it wasn't caught. Just want to make sure we are tracking future work as well if thats the case

britt6612: I see your above comment 
We are not linitng .tsx file today in v8 eslint config. To start linting .tsx file, we need to set it up.

do we have an issue for this? @Sulaymon333

britt6612:
are we fixing the errors we are getting when we run the `pnpm eslint`?

when I ran this I got the following:

```
/Users/archibeq/Documents/hpe-design-system/shared/aries-core/tools/release-stable.js
  25:1   error    This line has a length of 86. Maximum allowed is 80                            max-len
  25:54  warning  Unused eslint-disable directive (no problems were reported from 'no-console')
  28:3   warning  Unused eslint-disable directive (no problems were reported from 'no-console')

✖ 949617 problems (919351 errors, 30266 warnings)
  598895 errors and 3757 warnings potentially fixable with the `--fix` option.
```

MikeKingdom:
> > Similarly, in aries-core .tsx files don't seem to be linted.
> 
> We are not linitng .tsx file today in v8 eslint config. To start linting .tsx file, we need to set it up.
> 
> I added locally eslint config scoped to aries-core for testing to cover .tsx files and there are bunch of errors that looks genuine and needs fixes (about 11 errors and 28 warning fixable with via `--fix`). That can be a follow on ticket to this if needed or can be fixed in this PR as well.

These all look very simple so I'd argue for turning on tsx and fixing them in this PR. It can be a follow on, but the point of this ticket was get basically all the workspaces linting that make sense and with as much common config as possible (rules and file types as well as versions.)

shared/hooks/package.json
MikeKingdom: > do we need to have a local eslint file as well for the hooks? similar to the other folders.

I believe we do and should. I thought it was linting previously (has lint target in its package.json) but it does nothing now:
```

Oops! Something went wrong! :(

ESLint: 9.39.1

You are linting ".", but all of the files matching the glob pattern "." are ignored.
```

MikeKingdom:
> are we fixing the errors we are getting when we run the `pnpm eslint`?

`pnpm eslint` basically just runs the `eslint` cli command from the root and doesn't have the `maxerrors` or `src` path (and maybe some of the other local config) so it seems to complain about more files that the individual checks aren't looking at. That seems somewhat worthwhile for local dev although Copilot seems to agree that the individual workspace invocations may be better for CI (it would do `pnpm -r lint`)

MikeKingdom (commented):
It seems like mainly what's remaining for me is the `shared/hooks` lint and tsx config and cleanup in aries-core.

Sulaymon333:
> It seems like mainly what's remaining for me is the `shared/hooks` lint and tsx config and cleanup in aries-core.

@MikeKingdom  @britt6612 

1. aries-core - eslint is added and related files with linting errors are fixed via this [commit](https://github.com/grommet/hpe-design-system/pull/5919/changes/07c4e982418d045f4fde54b60eda0a459d28af65)
2. enable linting for .tsx files inside docs and fix related files via this [commit ](https://github.com/grommet/hpe-design-system/pull/5919/changes/ea8aa9ef18308d25b8b8b97ade3710783df33292)
3. enable linting for shared/hooks and fix related files via this [commit](https://github.com/grommet/hpe-design-system/pull/5919/changes/3dbe99b2a61079b3efd49f9613d81d4952658531)

MikeKingdom (commented):
A couple of fundamental things I'm still noticing:

1.  Only `apps/docs` is using the rules settings in the root `eslint.config.mjs`
1.  Some workspaces are still using typescript-eslint (recommendation is to be using eslint at this point like `shared/hooks` is now and share the rule defs in the root `eslint.config.mjs`.)

MikeKingdom:
> A couple of fundamental things I'm still noticing:
> 
> 1. Only `apps/docs` is using the rules settings in the root `eslint.config.mjs`
> 2. Some workspaces are still using typescript-eslint (recommendation is to be using eslint at this point like `shared/hooks` is now and share the rule defs in the root `eslint.config.mjs`.)

I added a commit that makes the eslint rules shared for apps/docs, shared/hooks and shared/aries-core. I recommend cleaning up the other workspaces too to share the rules config

MikeKingdom (approved):

eslint.config.mjs
Sulaymon333: done.

shared/aries-core/src/stories/navigation/content/NavigationPanel.tsx
```diff
@@ -39,10 +39,10 @@ export const NavigationPanel = ({
     open: expanded,
   };
 
-  const messages = {
-    layerOpen: `${navTitle} navigation opened.`,
-    layerClose: `${navTitle} navigation closed.`,
-  };
+  // const messages = {
```
britt6612, line 42: @Sulaymon333 this was not commented out prior to this PR?

Sulaymon333: @britt6612 yes it was not. Eslint flagged it as unused variable that is why it is commented out in this PR. 

The actual [usage](https://github.com/grommet/hpe-design-system/blob/bb2a6cb18e2a448e9061a09936bd5e5d3a3a22b2/shared/aries-core/src/stories/navigation/content/NavigationPanel.tsx#L57) (lines 57, 59, 64 & 96) in the code today are all commented out.

britt6612: okay yes thank you for pointing that out thats fine then!

britt6612:
Screenshot 2026-04-20 at 10 58 10 PM

Getting this error when I run the lint

britt6612:
@Sulaymon333 I forget are we worried about these errors currently?

Screenshot 2026-04-20 at 11 06 09 PM

Sulaymon333:
> Screenshot 2026-04-20 at 10 58 10 PM
> Getting this error when I run the lint

@britt6612 This came as a result of merging latest changes from master. It is now addressed via this [commit](https://github.com/grommet/hpe-design-system/pull/5919/changes/32ad12a83e24f2df8f8cd54099ecc3a8931d255e)

Sulaymon333:
> @Sulaymon333 I forget are we worried about these errors currently?
> 
> Screenshot 2026-04-20 at 11 06 09 PM

Please can you run the docs site via `pnpm --filter docs start`? That should build latest changes in hooks package and will resolve the import.

shared/hooks/eslint.config.mjs
```diff
@@ -31,6 +31,7 @@ export default defineConfig(
         name: 'override-rules',
         rules: {
             ...sharedBaseRules,
+            'react-hooks/immutability': 'off',
```
MikeKingdom: This rule shouldn't be disabled (and neither probably should be the one below.) Instead fix the problem or use an inline comment to disable the check

Sulaymon333: @MikeKingdom I fixed the eslint error and add inline disable rule for `'react-hooks/set-state-in-effect': 'off',` All done via this [commit](https://github.com/grommet/hpe-design-system/pull/5919/changes/22811390b2b6136f74f463544156f506b735ee57)

Sulaymon333:
@MikeKingdom I added `--cache` flag to improve eslint timing on local development via this [commit](https://github.com/grommet/hpe-design-system/pull/5919/changes/b837156515cf6cdb7207f9d96b90e9ba8735e8b5). More info - https://eslint.org/docs/latest/use/command-line-interface#caching.

I gain a huge performance (17s to 2s for docs site) after adding `--cache` since eslint lint will only check files that has changed since last run. See attached images below (with and without the flag)

Another useful flag is `--concurrency`. I think CI linting workflow will benefit from the `--concurrency` flag - https://eslint.org/docs/latest/use/command-line-interface#--concurrency 

**Without `--cache` flag**
image

**With `--cache` flag**
image

cc: @britt6612

MikeKingdom (approved):

britt6612 (approved):

Issue #5597, opened 2025-12-02
Move aries-site examples -> aries-core


MikeKingdom:
Currently there is a storybook in `aries-core` that shows examples from `aries-site`. Since `aries-site` depends on `aries-core` and `aries-core`'s storybook tries to use examples in `aries-site` this causes a circle dependency.

A better solution is to move the examples either into `aries-core` (and export them) or their own package and let `aries-site` reference them from there. Also the storybook should be moved to `apps/examples`. The new `apps` directory is where the various netlify-able apps are intended to be.

- [ ] Move examples from `aries-site` to `aries-core`
- [ ] Move storybook from `aries-core` to `apps/examples`
- [ ] Update netlify config for the storybook
- [ ] Update references
- [ ] Add `apps` to pnpm-workspace.yaml` as needed.

PR #5927, opened 2026-03-17
DS Storybook - Add code to pattern stories

jcfilben:
#### What does this PR do?
Adds code to the code panel section in storybook for the pattern stories

### DIFF
--- a/shared/aries-core/.storybook/main.mjs
+++ b/shared/aries-core/.storybook/main.mjs
@@ -69,12 +69,15 @@ const config = {
       resourceQuery: /raw/, // Matches files imported with ?raw
       type: 'asset/source', // Loads as a raw string
     });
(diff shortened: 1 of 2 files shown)


britt6612 (approved):
I know this `?raw` import approach is a webpack-specific workaround to load example source files as plain strings for Storybook code panel, so couple things to keep in mind.


 1. If example files are moved out of aries-site (apps/docs) into a new location, the `apps/docs` alias and all `?raw` import paths in the story files will need to be updated to reflect the new location.

2. If we switch from webpack to Vite, this custom rule is no longer needed.
Vite supports `?raw` imports natively and ships TypeScript types for them
via `vite/client`.

Given the above and we need this to release MVP approving PR
https://github.com/grommet/hpe-design-system/issues/5597

PR #5977, opened 2026-03-26
[Docs] Fix regression on whats-new page

Sulaymon333:
[Deploy Preview](https://deploy-preview-5977--keen-mayer-a86c8b.netlify.app/whats-new)

#### What does this PR do?

- Fix regression issue on whats-new page


MikeKingdom (commented):
Why is this tied to `isLanding`? Ie. any hints on the regression cause?

MikeKingdom (commented):
So looking at this closer, DocsLayout has the PageContent. This one doesn't because it was told to be "plain"? To me that means nothing in Layout.js should change and What's new should have its own padding.

That said, it does seem like whats-new is dead code kinda except that the route is there?

PR #6105, opened 2026-04-15
Add syntax switcher to tokens page

### LINKED ISSUE #5983: DS Site Context Switcher - Update All Tokens page with Syntax Switcher
Update the all tokens page on the DS site to display each token type that we want to expose and the associated data.

Pre-requiste work:

- [x] https://github.com/grommet/hpe-design-system/issues/5980
- [x] https://github.com/grommet/hpe-design-system/issues/5981
- [x] https://github.com/grommet/hpe-design-system/issues/5982

Julia's design: https://www.figma.com/design/qKH1POCGUkhY2nD9jc9TtY/Theme-reference-syntax?node-id=214-252&t=QIkLRnVtuv1BUmPR-0

jcfilben:
https://deploy-preview-6105--keen-mayer-a86c8b.netlify.app/design-tokens/all-design-tokens?token=component%2Felement&tokenTypes=docs&mode=default

#### What does this PR do?
Redesigns the "All design tokens" page in the docs site to support comparing multiple token types side-by-side, adds shareable URL state for the active token, selected types, and color mode, and integrates Grommet theme tokens using the same token source and nav structure as the other token types.

Closes https://github.com/grommet/hpe-design-system/issues/5983


apps/docs/src/components/content/docsTokenUtils.js
```diff
@@ -0,0 +1,72 @@
...
+      const currentToken = tokens[mode][token];
+
+      const parts = token.split('.');
+      const category = parts[1];
+      const level = getCollectionFromMode(mode);
+
+      const normalizedMode =
+          level === 'semantic' &&
+          mode !== 'global' &&
+          mode !== 'dimension'
```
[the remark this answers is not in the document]
jcfilben: Resolved

apps/docs/src/components/content/docsTokenUtils.js
```diff
@@ -0,0 +1,72 @@
...
+  'light',
+  'dark',
+  'small',
+]);
+
+const getCollectionFromMode = mode => {
+  if (modeToCollection[mode]) return modeToCollection[mode];
+  if (semanticModes.has(mode)) return 'semantic';
+  return mode;
+};
```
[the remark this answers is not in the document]
jcfilben: Resolved

apps/docs/src/components/content/design-tokens/DesignTokensTable.js
```diff
@@ -104,10 +123,49 @@ export const DesignTokensTable = ({ active, maxHeight, toolbar }) => {
...
   const mode = selectedMode || hookSelectedMode;
   const modeOptions = modes || hookModes;
-  const currentData = data || hookData;
+  const currentData = dataProp || contextData || hookData;
+
+  const columns = useMemo(() => {
+    const dynamicTokenColumns =
+      tokenTypeColumns && tokenTypeColumns.length > 0
+        ? tokenTypeColumns.map(({ id, label }) =>
+            getTokenColumn(`token__${id}`, label),
```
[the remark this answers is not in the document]
jcfilben: Resolved

halocline (commented):
This is looking quite nice, @jcfilben.

We will likely continuing refining the UI elements / presentation as we learn, however this gets us 90+% of the way there.

I haven't done a code review yet, primarily experience related at this point. A few notes on first pass:
- Should we consider co-locating DesignTokenContext, DesignTokensTable, and all utils under a common directory? (soft opinion)
- I like how we are now using URL query parameters to restore state.
  - However, it seems to always default me to the "element" component. I think semantic/color would be a better default.
  - When I select a token category such as borderWidth, navigate away to something like Templates, then return to the All Tokens page, it would be nice if my previous "view" was retained (i.e. borderWidth should be the active category, not reverting to the default). Bonus points for honoring my previous search string applied for filtering the token set. These view settings should persist for the duration of my session. These can both be handled as follow on work.
  - The "tokenTypes" selection seems to be retained and working well. This should persist across sessions - which it appears to be doing.
- focusIndicator page seems to result in an error

apps/docs/src/components/content/design-tokens/DesignTokensTable.js
```diff
@@ -24,73 +24,98 @@ import {
 } from './designTokenUtils';
 import { DesignTokenContext } from './DesignTokenContext';
```
halocline, line 25: What is the relationship between DesignTokenContext, useLocalStorage, and useSessionStorage? To what degree do they provide overlapping functionality / serve duplicative purposes, if at all?

halocline (commented):
- Where do I find the messaging for "N/A. Handled by the theme." and "N/A. Handled by Figma."?
- Where are the Grommet utils?
- Typescript?
- Tests accompanying the util functions?

apps/docs/src/pages/design-tokens/all-design-tokens.js
```diff
@@ -99,8 +119,162 @@ const NavLayer = ({ children, onClose }) => {
...
+const docsTokens = Object.fromEntries(
+  Object.entries(structuredTokens).filter(
+    ([collection]) =>
+      collection !== 'hpeTheme' &&
+      collection !== 'figma' &&
+      collection !== 'css',
+  ),
+);
+
+const tokenTypeOrder = ['docs', 'css', 'hpeTheme', 'figma'];
```
halocline: All of these are distributions of the "HPE Theme."

```suggestion
const tokenTypeOrder = ['docs', 'css', 'grommetThemeHpe', 'figma'];
```

apps/docs/src/pages/design-tokens/all-design-tokens.js
```diff
@@ -99,8 +119,162 @@ const NavLayer = ({ children, onClose }) => {
...
+  { label: 'Grommet', value: 'hpeTheme' },
+  { label: 'Figma', value: 'figma' },
+];
+
+const tokenTypeLabels = {
+  docs: 'HPE design tokens',
+  css: 'CSS',
+  hpeTheme: 'Grommet',
+  figma: 'Figma',
+};
```
halocline: These seem duplicative defining the same labels in two different spots which are prone to falling out of alignment over time.

apps/docs/src/pages/design-tokens/all-design-tokens.js
```diff
@@ -99,8 +119,162 @@ const NavLayer = ({ children, onClose }) => {
...
+      collection !== 'hpeTheme' &&
+      collection !== 'figma' &&
+      collection !== 'css',
+  ),
+);
+
+const tokenTypeOrder = ['docs', 'css', 'hpeTheme', 'figma'];
+
+const tokenTypeOptions = [
+  { label: 'HPE design tokens', value: 'docs' },
```
halocline: Proper noun.

```suggestion
  { label: 'HPE Design Tokens', value: 'docs' },
```

apps/docs/src/pages/design-tokens/all-design-tokens.js
```diff
@@ -99,8 +119,162 @@ const NavLayer = ({ children, onClose }) => {
...
+  { label: 'CSS', value: 'css' },
+  { label: 'Grommet', value: 'hpeTheme' },
+  { label: 'Figma', value: 'figma' },
+];
+
+const tokenTypeLabels = {
+  docs: 'HPE design tokens',
```
halocline:
```suggestion
  docs: 'HPE Design Tokens',
```

apps/docs/src/pages/design-tokens/all-design-tokens.js
```diff
@@ -99,8 +119,162 @@ const NavLayer = ({ children, onClose }) => {
...
+      collection !== 'figma' &&
+      collection !== 'css',
+  ),
+);
+
+const tokenTypeOrder = ['docs', 'css', 'hpeTheme', 'figma'];
+
+const tokenTypeOptions = [
+  { label: 'HPE design tokens', value: 'docs' },
+  { label: 'CSS', value: 'css' },
```
halocline: Soft/medium opinion.

```suggestion
  { label: 'CSS vars', value: 'css' },
```

apps/docs/src/pages/design-tokens/all-design-tokens.js
```diff
@@ -99,8 +119,162 @@ const NavLayer = ({ children, onClose }) => {
...
+
+const tokenTypeLabels = {
+  docs: 'HPE design tokens',
+  css: 'CSS',
+  hpeTheme: 'Grommet',
+  figma: 'Figma',
+};
+
+const getEmptyStateMessage = type => {
+  if (type === 'hpeTheme') return 'N/A. Handled by the theme.';
```
halocline: Found it! Was hidden by the large diff.

```suggestion
  if (type === 'hpeTheme') return 'N/A. Handled by `grommet-theme-hpe`.';
```

apps/docs/src/pages/design-tokens/all-design-tokens.js
```diff
@@ -99,8 +119,162 @@ const NavLayer = ({ children, onClose }) => {
...
+  if (type === 'figma') return 'N/A. Handled by Figma.';
```
halocline:
```suggestion
  if (type === 'figma') return 'N/A. Handled by Figma library.';
```

Sulaymon333:
@jcfilben, great progress so far.  **Would it make sense to move the `Description` column to the end of the table?**

Right now, the column is quite wide but often empty, which pushes the resolved value further out of view. As a result, users have to horizontally scroll to see one of the most important pieces of information on the table.

Reordering the columns could improve usability by allowing users to more easily compare the tokens (Grommet, Figma, CSS, etc) alongside its resolved value without needing to scroll. The description would still be accessible, just less disruptive to the primary workflow.

@halocline @julauxen 

image

julauxen:
> @jcfilben, great progress so far. **Would it make sense to move the `Description` column to the end of the table?**
> 
> Right now, the column is quite wide but often empty, which pushes the resolved value further out of view. As a result, users have to horizontally scroll to see one of the most important pieces of information on the table.
> 
> Reordering the columns could improve usability by allowing users to more easily compare the tokens (Grommet, Figma, CSS, etc) alongside its resolved value without needing to scroll. The description would still be accessible, just less disruptive to the primary workflow.
> 
> @halocline @julauxen
> 
> image

Hello @Sulaymon333, output value is the least valuable column on the table, since the semantic names of tokens should give the user enough indication of where to use them, so output value is more of an extra info. Description, specially on the semantic set, is used to explain where to use each token value. I would keep the order as is, but maybe we can do something about the width of the column @jcfilben?

julauxen (approved):
From the design side, this follows our agreements. 

Like Matt mentioned we can continue on refining the UI, specially the scroll behaviour (having a scroll on table and on page seem confusing to navigate). I created a ticket to take care of that here: https://github.com/grommet/hpe-design-system/issues/6121
@jcfilben where do you think this ticket should be linked to? the syntax switcher parent ticket? or be on its own?

apps/docs/src/pages/design-tokens/all-design-tokens.js
jcfilben: Fixed

apps/docs/src/pages/design-tokens/all-design-tokens.js
jcfilben: Fixed

apps/docs/src/pages/design-tokens/all-design-tokens.js
jcfilben: Fixed

apps/docs/src/pages/design-tokens/all-design-tokens.js
jcfilben: Fixed

apps/docs/src/pages/design-tokens/all-design-tokens.js
jcfilben: Aligned to this suggestion

apps/docs/src/pages/design-tokens/all-design-tokens.js
jcfilben: Fixed

apps/docs/src/pages/design-tokens/all-design-tokens.js
jcfilben: Fixed

jcfilben:
> Where are the Grommet utils?

Moved from all-design-tokens into it's own file, grommetThemeHpeUtils.js‎

> Typescript?

Currently we don't have typescript set up in the /docs directory, this PR didn't feel like the right place to handle adding that

> Tests accompanying the util functions?

Yes I think we could have tests. Right now we are only set up for e2e tests in `/docs`, didn't want to make this PR too big by trying to handle adding unit testing support to `/docs`

jcfilben:
> Should we consider co-locating DesignTokenContext, DesignTokensTable, and all utils under a common directory? (soft opinion)

Done. This caused some additional file changes so may have made the PR a little harder to review, let me know if you would like me to revert.

> I like how we are now using URL query parameters to restore state.
> However, it seems to always default me to the "element" component. I think semantic/color would be a better default.

semantic/color is now the default

> When I select a token category such as borderWidth, navigate away to something like Templates, then return to the All Tokens page, it would be nice if my previous "view" was retained (i.e. borderWidth should be the active category, not reverting to the default). Bonus points for honoring my previous search string applied for filtering the token set. These view settings should persist for the duration of my session. These can both be handled as follow on work.

Will treat this as follow on work. Was running into issues, particularly with trying to preserve the search string.

> focusIndicator page seems to result in an error

Fixed

apps/docs/src/components/content/design-tokens/DesignTokensTable.js
jcfilben: DesignTokensContext is used to pass the token table state, for example between the page and table components in all-design-tokens.js and DesignTokensTable.js.

all-design-tokens.js currently does manual localStorage handling via getInitialTokenTypes and localStorage access. I was running into issues with the useLocalStorage hook on the all-design-tokens page where it was being too aggressive trying to preserve the url params and I couldn't change pages after going to the all design tokens table. Let me know if you would like me to remove the useLocalStorage hook or modify the code to try and use it.

jcfilben:
> I would keep the order as is, but maybe we can do something about the width of the column @jcfilben?

I could make the column width smaller, but I think when we last chatted you wanted it a little bigger so the text didn't wrap as much (if I am remembering correctly). Let me know if you want to adjust or leave as is.

jcfilben:
> @jcfilben where do you think this ticket should be linked to? the syntax switcher parent ticket? or be on its own?

No strong preference. We could link to the parent ticket to keep all the work related to this easier to trace.

julauxen:
> > I would keep the order as is, but maybe we can do something about the width of the column @jcfilben?
> 
> I could make the column width smaller, but I think when we last chatted you wanted it a little bigger so the text didn't wrap as much (if I am remembering correctly). Let me know if you want to adjust or leave as is.

Yes, I though we might change the width depending on context, but if that's not possible we can leave as is

apps/docs/src/pages/design-tokens/all-design-tokens.js
```diff
@@ -127,31 +311,335 @@ const AllTokens = () => {
...
+    showGrommetThemeHpeManagedEmptyState ||
+    showGrommetThemeHpeCategoryEmptyState;
+  const isFigmaManagedEmptyState =
+    showFigmaManagedEmptyState || showFigmaCategoryEmptyState;
+  const showSourceEmptyState =
+    showGrommetThemeHpeEmptyState || isFigmaManagedEmptyState;
+  const sourceEmptyStateText = isFigmaManagedEmptyState
+    ? 'Handled by Figma library.'
+    : 'Handled by grommet-theme-hpe.';
+  const showManagedEmptyState =
```
britt6612, line 394: why do we need both `showSourceEmptyState` and `showManagedEmptyState` logically are they not the same?

apps/docs/src/pages/design-tokens/all-design-tokens.js
```diff
@@ -127,31 +311,335 @@ const AllTokens = () => {
...
+        rowsForType = rowsForType.filter(datum =>
+          datum.token.endsWith('.fontSize'),
+        );
+      }
+
+      rowsForType.forEach(datum => {
+        const canonicalToken = Array.isArray(datum.path)
+          ? datum.path.join('.')
+          : datum.sourceToken || datum.token;
+        let tokenValue = datum.displayToken || datum.token;
```
britt6612: wont `datum.displayToken` be undefined? does `displayToken` get returned in the `getToken` func

apps/docs/src/pages/design-tokens/all-design-tokens.js
```diff
@@ -127,31 +311,335 @@ const AllTokens = () => {
...
+        query: nextQuery,
+      },
+      undefined,
+      { shallow: true, scroll: false },
     );
+  }, [
+    active,
+    isUrlSyncReady,
+    primaryTokenType,
+    router,
```
britt6612: is the `router` correct will this cause re-runs whenever the Next.js replaces the router reference.

apps/docs/src/pages/design-tokens/all-design-tokens.js
```diff
@@ -111,12 +254,53 @@ const AllTokens = () => {
     selectedMode,
     setSelectedMode,
     modes,
-  } = useDesignTokens('semantic.color');
+  } = useDesignTokens(getDefaultActiveTokenPath());
```
britt6612: should `getDefaultActiveTokenPath()` be wrapped in useMemo.

apps/docs/src/components/content/design-tokens/tokenUtilsShared.js
```diff
@@ -0,0 +1,85 @@
...
+      const transformedTokenName = transformTokenName
+        ? transformTokenName(token)
+        : token;
+
+      // Transform token data if a transformer is provided
+      const transformedTokenData = transformTokenData
+        ? transformTokenData(token, currentToken)
+        : currentToken;
+
+      const nextTokens = structuredTokens;
```
britt6612: do we need to define the `nextTokens` or just use the structuredTokens below?

apps/docs/src/components/content/design-tokens/tokenUtilsShared.js
```diff
@@ -0,0 +1,85 @@
...
+ */
+const addStructuredTokens = (
+  structuredTokens,
+  { rootKey, transformTokenName, transformTokenData },
+) => {
+  Object.keys(tokens).forEach(mode => {
+    Object.keys(tokens[mode]).forEach(token => {
+      const currentToken = tokens[mode][token];
+      const parts = token.split('.');
+      const category = parts[1];
```
britt6612: simplify and then remove the const parts
```suggestion
     const category = token.split('.')[1];
```

apps/docs/src/components/content/design-tokens/figmaTokenUtils.js
```diff
@@ -0,0 +1,18 @@
...
+
+const toFigmaTokenName = tokenName => tokenName.replaceAll('.', '/');
+
+const addFigmaTokens = structuredTokens => {
+  addStructuredTokens(structuredTokens, {
+    rootKey: 'figma',
+    transformTokenName: toFigmaTokenName,
+    transformTokenData: (token, currentToken) => ({
+      ...currentToken,
+      $name: currentToken.$name
```
britt6612:
```suggestion
      $name: toFigmaTokenName(currentToken.$name || token),
```

britt6612: simplify

apps/docs/src/components/content/design-tokens/designTokenUtils.js
```diff
@@ -87,21 +64,30 @@ const getTokens = (tokenObj, mode) =>
     return {
       id: key,
       token: key,
+      sourceToken: tokenObj[key]?.modes[mode]?.$name,
+      path: tokenObj[key]?.modes[mode]?.path,
       type: tokenObj[key]?.modes[mode]?.$type,
       description: tokenObj[key]?.modes[mode]?.$description,
       value: tokenObj[key]?.modes[mode].$value,
```
britt6612:
```suggestion
      value: tokenObj[key]?.modes[mode]?.$value,
```

apps/docs/src/components/content/design-tokens/DesignTokensTable.js
```diff
@@ -0,0 +1,241 @@
...
+const getTokenColumn = (property, header) => ({
+  property,
+  header,
+  render: datum => {
+    // Default token column should prefer source token name when present.
+    const displayValue =
+      property === 'token' ? datum.sourceToken || datum.token : datum[property];
+
+    return (
+      <Box direction="row">
```
britt6612: do we need this `direction='row'`

apps/docs/src/components/content/design-tokens/DesignTokensTable.js
```diff
@@ -0,0 +1,241 @@
...
+      </Box>
+    );
+  },
+});
+
+const descriptionColumn = {
+  property: 'description',
+  header: 'Description',
+  size: 'large',
+  render: datum => <Text>{datum.description ? datum.description : '--'}</Text>,
```
britt6612:
```suggestion
  render: datum => <Text>{datum.description || '--'}</Text>,
```

apps/docs/src/pages/design-tokens/all-design-tokens.js
jcfilben: Good catch, fixed

apps/docs/src/pages/design-tokens/all-design-tokens.js
jcfilben: Yes datum.displayToken is undefined, changed to remove it

apps/docs/src/pages/design-tokens/all-design-tokens.js
jcfilben: Tightened this up a bit so that it isn't as broad

apps/docs/src/pages/design-tokens/all-design-tokens.js
jcfilben: Yes, fixed

apps/docs/src/components/content/design-tokens/tokenUtilsShared.js
jcfilben: Removed nextTokens and just used structuredTokens

apps/docs/src/components/content/design-tokens/tokenUtilsShared.js
jcfilben: Done

apps/docs/src/components/content/design-tokens/figmaTokenUtils.js
jcfilben: Done

apps/docs/src/components/content/design-tokens/designTokenUtils.js
jcfilben: Done

apps/docs/src/components/content/design-tokens/DesignTokensTable.js
jcfilben: No, removed this box

apps/docs/src/components/content/design-tokens/DesignTokensTable.js
jcfilben: Done

britt6612 (approved):
looks good!

MikeKingdom (commented):
The tokens are specifying Menlo for a font-family but on my system it's just coming out Times (or some other serif font.) I don't think you have a font-face for it and it may not be on my system by default. 

Why are we using Menlo (I saw it in Julia's Figma)?

jcfilben:
> The tokens are specifying Menlo for a font-family but on my system it's just coming out Times (or some other serif font.) I don't think you have a font-face for it and it may not be on my system by default.
> 
> Why are we using Menlo (I saw it in Julia's Figma)?

Hmm I'm not seeing that the tokens are specifying Menlo for the font-family, can you share where you are seeing that?
Screenshot 2026-05-06 at 6 09 37 PM

Screenshot 2026-05-06 at 6 09 47 PM

apps/docs/src/components/content/design-tokens/DesignTokensTable.js
```diff
@@ -0,0 +1,241 @@
...
+    // Default token column should prefer source token name when present.
+    const displayValue =
+      property === 'token' ? datum.sourceToken || datum.token : datum[property];
+
+    return (
+      <Box
+        background="background-contrast"
+        pad="3xsmall"
+        round="xsmall"
+        style={{ fontFamily: 'Menlo' }}
```
Sulaymon333: @jcfilben this PR has code change that specifies `Menlo` font family here. 
Are we supposed to be using a new `Menlo` Font family? Menlo is macOS-only. On Windows/Linux it falls back to the browser default (often serif) I believe that is why @MikeKingdom sees the serif font. 

My understanding is that only `HPE Graphik`  with `Aria, san-serif` fallback should be used throughout the site.

apps/docs/src/pages/design-tokens/all-design-tokens.js
```diff
@@ -99,10 +121,166 @@ const NavLayer = ({ children, onClose }) => {
...
+};
+
+const isManagedCollectionForSource = (source, collection) =>
+  managedCollectionsBySource[source]?.includes(collection);
+
+const isManagedCategoryForSource = (source, category) =>
+  managedCategoriesBySource[source]?.includes(category);
+
+const getEmptyStateMessage = type => {
+  if (type === 'grommetThemeHpe') return 'N/A. Handled by `grommet-theme-hpe`.';
```
Sulaymon333: Nitpick: If the intention is to render an inline code, the backticks around `grommet-theme-hpe` does not render as inline code here - they appear as literal ` characters in the UI since this is inside a Grommet <Text> component, not markdown. Can we consider double quotes instead for easier readability:

@julauxen, I did not see any guidance around usage of "double quotation" or "backtick"  with text on the website.

```js
  if (type === 'grommetThemeHpe') return 'N/A. Handled by "grommet-theme-hpe".';
```

Image

apps/docs/src/components/content/design-tokens/DesignTokensTable.js
jcfilben: Ah thank you, this is an artifact from the prior DesignTokensTable file. Taylor has used menlo a few other places in the repo. I'm not sure why this was used. It does seem a bit odd. I can go ahead and remove it from the table.

Screenshot 2026-05-07 at 8 32 46 AM

jcfilben: I tried removing this style but the table looks odd without it. Since this was a pre-existing style I'm going to leave as-is for now. We can remove in a future PR if desired.

Image

MikeKingdom: I don't think this is OK. It doesn't render correctly on windows.

apps/docs/src/pages/design-tokens/all-design-tokens.js
julauxen: let's just remove the '' and not use anything else

MikeKingdom:
> > The tokens are specifying Menlo for a font-family but on my system it's just coming out Times (or some other serif font.) I don't think you have a font-face for it and it may not be on my system by default.
> > Why are we using Menlo (I saw it in Julia's Figma)?
> 
> Hmm I'm not seeing that the tokens are specifying Menlo for the font-family, can you share where you are seeing that? 

That's actually my point. It doesn't so on windows it can't do anything with that family so it defaults to times roman which looks horrible

It ends up with a serif font(Times New Roman instead of Menlo) along side a sans-serif font (Graphik)
image

apps/docs/src/components/content/design-tokens/DesignTokensTable.js
MikeKingdom: Word from BrandCentral is use Consolas for monospace. See [Brand Central](https://brandcentral.hpe.com/brand-central/content/typography) under "mono font".

MikeKingdom: With `font-family: Consolas` 
image

britt6612: @MikeKingdom does it just not render for you on windows? so is that also the case today since that font was already there?

MikeKingdom: Any case in the site using Menlo (in this PR or pre-existing) renders as Times New Roman on Windows. Windows doesn't have the Menlo font by default so it uses the default browser font which is always Times. So we decided the  best answer is probably `font-family: Consolas, Menlo, monospace`. This should work for both windows and mac and at least fallback to some monospace font rather than a serif font like Times.

apps/docs/src/pages/design-tokens/all-design-tokens.js
jcfilben: Removed the backticks

jcfilben:
> That's actually my point. It doesn't so on windows it can't do anything with that family so it defaults to times roman which looks horrible
> 
> It ends up with a serif font(Times New Roman instead of Menlo) along side a sans-serif font (Graphik) image

Gotcha, makes sense. Changed fontfamily to `font-family: Consolas, Menlo, monospace`

apps/docs/src/components/content/design-tokens/DesignTokensTable.js
jcfilben: Changed to use `font-family: Consolas, Menlo, monospace`

MikeKingdom (approved):
Thanks! Looks great.

Sulaymon333 (approved):
LGTM

apps/docs/src/pages/design-tokens/all-design-tokens.js
```diff
@@ -165,29 +659,62 @@ const AllTokens = () => {
...
-                    icon={<Menu />}
-                    a11yTitle="Open design tokens menu"
-                    margin={{ top: '3xsmall' }}
-                    onClick={() => setOpenLayer(true)}
-                  />
-                ) : undefined}
-                <Heading level={2} margin="none" id="token-table-heading">
-                  {activeCollection}
-                </Heading>
+              <Box margin={{ bottom: 'medium' }}>
```
halocline: We should be avoiding margin for solving layout related problems. Children should be selfish and not be concerned with what is outside them. Layout / organizing children should be the responsibility of the parent.

```suggestion
```

cc: @MikeKingdom , @britt6612 , @Sulaymon333

apps/docs/src/pages/design-tokens/all-design-tokens.js
```diff
@@ -165,29 +659,62 @@ const AllTokens = () => {
...
+                      normalizeQuerySyncState({
+                        token: currentTokenParam,
+                        tokenTypes: normalizedValue.join(','),
+                        mode: getSingleQueryValue(queryMode) || selectedMode,
+                      }),
+                    );
+
+                    setSelectedTokenTypes(normalizedValue);
+                  }}
+                />
```
halocline: Worth moving this to an `onToggle` method so that logic isn't cluttering the JSX? Medium opinion.

apps/docs/src/pages/design-tokens/all-design-tokens.js
```diff
@@ -165,29 +659,62 @@ const AllTokens = () => {
...
+              </Box>
+              <Box gap="xsmall" margin={{ bottom: 'xsmall' }}>
```
halocline: I don't think this Box is needed. What does the gap affect? Margin should be avoided - prefer parent to organize the layout.

```suggestion
```

cc: @MikeKingdom , @britt6612 , @Sulaymon333

apps/docs/src/pages/design-tokens/all-design-tokens.js
```diff
@@ -165,29 +659,62 @@ const AllTokens = () => {
...
+                <Box direction="row" align="center" gap="xsmall">
+                  {!isDesktop ? (
+                    <Button
+                      icon={<Menu />}
+                      a11yTitle="Open design tokens menu"
+                      margin={{ top: '3xsmall' }}
```
halocline: Avoid margin.

What are we trying to align this control to? Why won't just align="center" work?

Now that I think about this. This control seems like it should belong to the content container, not to the heading. It's affinity should be anchored in the upper left of the content container.

```suggestion
```

cc: @MikeKingdom , @britt6612 , @Sulaymon333

apps/docs/src/pages/design-tokens/all-design-tokens.js
```diff
@@ -203,12 +730,28 @@ const AllTokens = () => {
...
+                  data={showManagedEmptyState ? [] : tableData}
+                  toolbar
+                  tokenTypeColumns={tokenTypeColumns}
+                />
+                {showManagedEmptyState ? (
+                  <Box
+                    alignSelf="center"
+                    align="center"
+                    gap="xxsmall"
+                    margin={{ bottom: 'medium' }}
```
halocline: Avoid margin.

```suggestion
```

cc: @MikeKingdom , @britt6612 , @Sulaymon333

apps/docs/src/pages/design-tokens/all-design-tokens.js
```diff
@@ -203,12 +730,28 @@ const AllTokens = () => {
...
-              />
+              <Box gap="small">
+                <DesignTokensTable
+                  active={active}
+                  data={showManagedEmptyState ? [] : tableData}
+                  toolbar
+                  tokenTypeColumns={tokenTypeColumns}
+                />
+                {showManagedEmptyState ? (
+                  <Box
```
halocline: Use `<EmptyState />` from `aries-core`? Re-use. No need to recreate.

cc: @MikeKingdom , @britt6612 , @Sulaymon333

apps/docs/src/components/content/design-tokens/DesignTokensTable.js
```diff
@@ -0,0 +1,241 @@
...
+                    }}
+                  />
+                </FormField>
+              </Box>
+            ) : undefined}
+          </Toolbar>
+          <DataSummary />
+        </>
+      ) : undefined}
+      <Box height={{ max: 'xlarge' }} overflow="auto">
```
halocline: This introduces double vertical scroll bars. Is that what we want? Can we avoid?

halocline: Is the intent to keep the toolbar within view/reach when scrolling? If so, should we be considering a sticky header/toolbar instead? Or remove?

apps/docs/src/components/content/design-tokens/DesignTokensTable.js
```diff
@@ -0,0 +1,241 @@
...
+        <DataTable
+          aria-describedby="token-table-heading"
+          verticalAlign="top"
+          primaryKey="id"
+          columns={columns}
+          style={{ minWidth: 'max-content' }}
```
halocline: Why is inline style needed? Can the parent Box have `width="max-content"`? and achieve the same?

```suggestion
```

cc: @MikeKingdom , @britt6612 , @Sulaymon333

halocline:
> > When I select a token category such as borderWidth, navigate away to something like Templates, then return to the All Tokens page, it would be nice if my previous "view" was retained (i.e. borderWidth should be the active category, not reverting to the default). Bonus points for honoring my previous search string applied for filtering the token set. These view settings should persist for the duration of my session. These can both be handled as follow on work.
> 
> Will treat this as follow on work. Was running into issues, particularly with trying to preserve the search string.
> 

Has an issue to track this been created?

apps/docs/src/pages/design-tokens/all-design-tokens.js
britt6612: I thought this code was just copied over so assumed that the `margin` wasn't an issue since most of this code was already in place. 

https://github.com/grommet/hpe-design-system/blob/master/apps/docs/src/pages/design-tokens/all-design-tokens.js#L173C1-L174C1

apps/docs/src/components/content/design-tokens/DesignTokensTable.js
jcfilben: When I chatted with Julia one of the things she wanted to adjust was preventing excessive line wraps. This ensures the height of the rows doesn't get too tall and she felt like it improved readability. here is a before and after of having this change. The description on these rows is long and in the before the line wraps cause the row height to increase.

Setting `width="max-content"` doesn't have the same effect.

Screenshot 2026-05-11 at 4 48 30 PM

Screenshot 2026-05-11 at 4 48 58 PM

apps/docs/src/components/content/design-tokens/DesignTokensTable.js
jcfilben: I believe the intent was just making sure they table didn't get too long, you can still access sidebar and header/nav. Julia created a follow on ticket to look at alternatives to the double scroll bar. https://github.com/grommet/hpe-design-system/issues/6121

jcfilben:
> > > When I select a token category such as borderWidth, navigate away to something like Templates, then return to the All Tokens page, it would be nice if my previous "view" was retained (i.e. borderWidth should be the active category, not reverting to the default). Bonus points for honoring my previous search string applied for filtering the token set. These view settings should persist for the duration of my session. These can both be handled as follow on work.
> > 
> > 
> > Will treat this as follow on work. Was running into issues, particularly with trying to preserve the search string.
> 
> Has an issue to track this been created?

Follow on ticket created https://github.com/grommet/hpe-design-system/issues/6171

apps/docs/src/pages/design-tokens/all-design-tokens.js
jcfilben: Yeah a decent amount of this was pre-existing, github is just flagging it as lines changed since things in some of these files moved around enough.

I did go through and change these margin references though

apps/docs/src/pages/design-tokens/all-design-tokens.js
jcfilben: Fixed

apps/docs/src/pages/design-tokens/all-design-tokens.js
jcfilben: I kept this as is for now. Maybe a personal opinion but in a larger file like this I like to be able to read this stuff inline instead of having to scroll to a different spot, unless the code is used in multiple places then I'll pull it out. Open to other perspectives and if we want to align to a norm.

apps/docs/src/pages/design-tokens/all-design-tokens.js
jcfilben: Removed

apps/docs/src/pages/design-tokens/all-design-tokens.js
jcfilben: Removed

apps/docs/src/pages/design-tokens/all-design-tokens.js
jcfilben: Changed to use EmptyState

apps/docs/src/components/content/design-tokens/DesignTokensTable.js
halocline: > 
> Setting `width="max-content"` doesn't have the same effect.
> 

On which element? The DataTable or parent Box?

jcfilben: I had set it on the DataTable, just tried setting it on the Box and that had the desired effect, so I will push that up.

apps/docs/src/components/content/design-tokens/DesignTokensTable.js
```diff
@@ -0,0 +1,240 @@
...
+      nextProperties.token = { search: true };
+    }
+
+    return nextProperties;
+  }, [tokenTypeColumns]);
+
+  return (
+    <Data
+      data={currentData}
+      pad={{ vertical: 'medium' }}
```
halocline:
```suggestion
```

halocline (approved):
Looks great, @jcfilben! Thank you for pushing this through.

Approving.

Can you commit my suggested removal of `pad` from `<Data />` prior to merging?

Issue #6107, opened 2026-04-17
Site navigation - Add HPE Logo to Left Navigation and hide Global Header/Footer


julauxen:
### Context
For the left navigation MVP, we need the product logo to live within the page/navigation itself. This allows us to hide the global header and footer service, which is a key requirement for the MVP and has been previously aligned with stakeholders. 

### Goal
Move the logo into the left navigation (when open) or page body so the global header and footer can be safely hidden without losing branding clarity. [See design here.](https://www.figma.com/design/lV4lc9Kbk0giOS20k9sbrJ/DS-Website-UI-updates?m=auto&node-id=2722-8295&t=rLxtQk6DYGOWCDtg-1)

### Success Criteria
- The logo is visible within the left navigation when open.
- The logo is visible on page body when navigation is closed.
- *The global header and footer are hidden (depends on Edwin's instructions)
- Branding remains consistent and accessible

Open this PR against DOCS-NEXT branch. See Matt's deploy preview here: https://docs-next--keen-mayer-a86c8b.netlify.app/

julauxen:
Unknown: where the analitycs take place, should we keep the footer

Sulaymon333:
The assumption is that we might be able to still have analytics when global header and footer are hidden. There is a current discussion to validate this assumption with Edwin Amirian from Chris' team.

Sulaymon333:
**Out of scope**
This ticket will not change the text content/layout on the homepage as seen on the attached [Figma design](https://www.figma.com/design/lV4lc9Kbk0giOS20k9sbrJ/DS-Website-UI-updates?m=auto&node-id=2722-8295&t=rLxtQk6DYGOWCDtg-1). It is specifically about the site navigation/global header and footer. Julia will make another follow on ticket for that.

julauxen:
Link to matt's poc: https://docs-next--keen-mayer-a86c8b.netlify.app/

julauxen:
This ticket is blocked because we are waiting on instructions from Artur Pietruk (dev) and Edwin Amirian (PO) on how to handle the global header/footer so we can hide it and still have the analytics. In the last e-mail we shared both the figma link and the URL that they requested (those were sent on April 30th) and we are now waiting on a response.

Sulaymon333:
### Summary of latest updates from Artur. 
More details from the email thread. @julauxen please let me in case I miss anything from his response.

[...]

PR #6144, opened 2026-04-28
[Docs App] Add logo to side site nav (conditionally rendering)

Sulaymon333:
[Deploy Preview](https://deploy-preview-6144--keen-mayer-a86c8b.netlify.app/)

#### What does this PR do?

- The logo is visible within the left navigation when open.
- The logo is visible on page body when navigation is closed.
- Branding remains consistent and accessible
- hide global header by commenting out line `<div id="header" className="header" />` inside `_doucment.js`

related to #6107


apps/docs/src/layouts/navigation/SideNavHeader.js
```diff
@@ -0,0 +1,18 @@
+import PropTypes from 'prop-types';
+import { NavHeader } from '@shared/aries-core';
+import { BrandIdentity } from '../../components';
+
+
+export const SideNavHeader = ({ open, setOpen, ...rest }) => (
```
Sulaymon333: I added a new component and expose NavHeader from `aries-core` to reuse already baked in styling, accessibility and logic.

Alternative will be to create a new component. for example

```js
// CustomNavHeader
import { useEffect, useState } from 'react';
import { Box, Button } from 'grommet';
import { Sidebar as SidebarIcon } from '@hpe-design/icons-grommet';
import { BrandIdentity } from '../../../components';


export const NavHeader = ({ open, setOpen, ...rest }) => {
    const [autoFocus, setAutoFocus] = useState(false);

    // After mount set autoFocus to true
    useEffect(() => {
        setAutoFocus(true);
    }, []);

    return (
        <Box
            direction="row"
            align="center"
            background="background-front"
            justify={open ? 'between' : 'center'}
            gap="xxsmall"
            pad={{
                top: 'xsmall',
                bottom: 'xxsmall',
                horizontal: open ? 'small' : undefined,
            }}
            style={{ position: 'sticky', top: 0 }}
            flex={false}
            {...rest}
        >
            {open && (
                <BrandIdentity logoOnly />
            )}
            <Button
                a11yTitle={open ? 'Close navigation' : 'Open navigation'}
                aria-expanded={open}
                active={open}
                alignSelf="start"
                autoFocus={autoFocus}
                icon={<SidebarIcon aria-hidden />}
                onClick={() => setOpen(!open)}
                tip={open ? 'Close navigation' : 'Open navigation'}
            />
        </Box>
    );
};

NavHeader.propTypes = {
    open: PropTypes.bool.isRequired,
    setOpen: PropTypes.func.isRequired,
};
```

What option do you suggest we keep?

apps/docs/src/layouts/navigation/NavContext.js
```diff
@@ -0,0 +1,21 @@
+import { createContext, useContext, useState, useMemo } from 'react';
+import PropTypes from 'prop-types';
+
+export const NavContext = createContext({
```
Sulaymon333, line 4: NavContext is added to lift state up so that app header can be aware of side nav toggle state

apps/docs/src/components/content/BrandIdentity.js
```diff
@@ -0,0 +1,64 @@
+import React, { useContext } from 'react';
+import { useRouter } from 'next/router';
+import PropTypes from 'prop-types';
+import { Box, Button, Image, Text } from 'grommet';
+import { ThemeContext } from 'styled-components';
+
+
+export const BrandIdentity = (
```
Sulaymon333: `BrandIdentity` is added since it involves more logic. They are listed below
1.  theme driven logo - light/dark HPE brand logo.
2. @SeamusLeonardHPE said, the HPE brand logo will not be added to the hpe icons grommet package. So, it should be loaded locally
3. no need of brand unlike AppIdentity

apps/docs/src/layouts/navigation/SideNavHeader.js
Sulaymon333: this is no longer required. I added `logo` prop to `NavigationMenu` component and it is much easier and cleaner in my opinion. see this [commit](https://github.com/grommet/hpe-design-system/pull/6144/changes/5132cd1fd73be6530c4bbbca56a44bbacf1bcc7b)

shared/aries-core/src/js/components/core/NavigationMenu/index.ts
```diff
@@ -1,2 +1,3 @@
 export * from './NavigationMenu';
+export * from './NavContainer';
 export * from './NavItem';
```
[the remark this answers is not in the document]
Sulaymon333: this is no longer needed and no exposure of `NavContainer` or `NavHeader` via barrel export. logo prop is now added to `NavigationMenu` component and it is rendered in place of default heading when provided.

shared/aries-core/src/js/components/core/NavigationMenu/NavigationMenu.tsx
```diff
@@ -80,20 +81,26 @@ export const NavigationMenu = ({
...
   const navigationId = 'navigation-menu';
   const menuTitle = title ? `${title}` : 'Navigation Menu';
 
   // Add unique Id to each item for aria and key purposes
   const items = useMemo(() => assignUniqueIds(itemsProp), [itemsProp]);
 
+  const setOpen = (nextOpen: boolean) => {
+    setOpenState(nextOpen);
+    onToggle?.(nextOpen);
+  };
```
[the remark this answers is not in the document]
Sulaymon333: Needs test coverage.

shared/aries-core/src/js/components/core/NavigationMenu/NavContainer/NavHeader.tsx
```diff
@@ -7,6 +7,7 @@ interface NavHeaderProps {
   title?: string;
   navigationId?: string;
   open: boolean;
+  logo?: React.ReactNode;
   setOpen: (open: boolean) => void;
 }
```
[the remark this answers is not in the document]
Sulaymon333: no longer relvant.

shared/aries-core/src/js/components/core/NavigationMenu/NavigationMenu.tsx
```diff
@@ -65,6 +65,7 @@ interface NavigationMenuProps extends BoxProps {
   header?: React.ReactNode;
   items: NavItemType[];
   open?: boolean;
+  onToggle?: (open: boolean) => void;
   title?: string;
```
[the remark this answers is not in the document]
Sulaymon333: docs need update

apps/docs/src/layouts/navigation/Navigation.js
```diff
@@ -20,6 +23,9 @@ export const Navigation = ({ ...rest }) => {
     <NavigationMenu
       activeItem={activeItem}
       items={navItems}
+      header={<SideNavHeader open={navOpen} setOpen={setNavOpen} />}
```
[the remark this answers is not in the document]
Sulaymon333: no longer relevant since logo prop is now added to `NavigationMenu` component and `NavHeader` is no longer exposed

apps/docs/src/components/content/BrandIdentity.js
Sulaymon333: I explore updating `AppIdentity`  and usage example as shown below.

```js
// example usage
<AppIdentity brand="hpe2" logoOnly /> // or
 <AppIdentity
              brand="hpe2"
              title="Design System"
              logo={!navOpen}
            />

```js
// AppIdentity.js updates
import React, { forwardRef, useContext } from 'react';
import { Box, Button, Image, Text } from 'grommet';
import { ThemeContext } from 'styled-components';

// TODO replace with DS icon package when available
import { Aruba } from 'grommet-icons';
import { Element } from '@hpe-design/icons-grommet';



const Hpe2Logo = () => {
    return (
        <Box width="4xsmall">
            <Image
                alt="HPE"
                fit="contain"
                src={theme?.dark
                    ? '/HPE_logo_full-clr_rev_rgb.svg'
                    : '/HPE_logo_full-clr_pos_rgb.svg'}
            />
        </Box>
    );
};

const brands = {
    hpe: {
        name: 'HPE',
        logo: <Element color="brand" height="medium" />,
    },
    hpe2: {
        logo: <Hpe2Logo />,
    },
    aruba: {
        name: 'Aruba',
        logo: <Aruba color="plain" height="medium" />,
    },
};

export const AppIdentity = forwardRef(
    (
        {
            brand,
            logo = true,
            logoOnly = false,
            href = '/',
            title,
            ...rest
        },
        ref,
    ) => {
        const textSize = 'medium';

        console.log('AppIdentity render', { brand, logo, logoOnly, href, title });

        return (
            <Button
                href={href}
                ref={ref}
                {...rest}
            >
                <Box
                    direction="row"
                    align="center"
                    gap="small"
                    // pad maintains accessible hit target
                    // non-responsive maintains same dimensions for mobile
                    pad={{ vertical: 'xsmall' }}
                    responsive={false}
                >
                    {brand && logo && brands[brand].logo}
                    {!logoOnly && (
                        <Box direction="row" gap="3xsmall" wrap>
                            <Text weight="bold" size={textSize} color="text-strong">
                                {brands[brand]?.name}
                            </Text>
                            <Text size={textSize} color="text-strong">
                                {title}
                            </Text>
                        </Box>
                    )}
                </Box>
            </Button>
        );
    },
);

AppIdentity.propTypes = {
    brand: PropTypes.string.isRequired,
    logo: PropTypes.bool,
    logoOnly: PropTypes.bool,
    href: PropTypes.string,
    title: PropTypes.string,
};
```

apps/docs/src/components/content/BrandIdentity.js
```diff
@@ -0,0 +1,62 @@
...
+                    direction="row"
+                    align="center"
+                    gap="small"
+                    // pad maintains accessible hit target
+                    // non-responsive maintains same dimensions for mobile
+                    pad={{ vertical: 'xsmall' }}
+                    responsive={false}
+                >
+                    {logo && (<Box width="4xsmall">
+                        <Image
```
jcfilben: In the design the hpe logo is a larger size, is this divergence from the design intentional?

apps/docs/src/components/content/BrandIdentity.js
```diff
@@ -0,0 +1,62 @@
...
+        href = '/',
+        title,
+        ...rest
+    }, ref) => {
+        const textSize = 'medium';
+        const theme = useContext(ThemeContext);
+
+
+        return (
+            <Button
```
jcfilben: I think it would be helpful to explain to screenreader users where this link will take them. They might need more information than just 'hpe logo' especially since we have more than one hpe logo on the page.

apps/docs/src/layouts/navigation/Navigation.js
```diff
@@ -20,6 +24,13 @@ export const Navigation = ({ ...rest }) => {
     <NavigationMenu
       activeItem={activeItem}
       items={navItems}
+      logo={
+        <Link href="/" passHref legacyBehavior>
```
jcfilben: What is legacyBehavior and should we be using it?

jcfilben (commented):
Is this the correct ticket for this work? https://github.com/grommet/hpe-design-system/issues/6107
If so it also mentioned removing the global header and footer but I don't see that change in this PR.

Also it looks like you have some copilot comments about adding tests and adjusting the documentation. Do those changes still need to go in?

shared/aries-core/src/js/components/core/NavigationMenu/NavigationMenu.tsx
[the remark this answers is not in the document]
Sulaymon333: Test will be added after @halocline validates the new prop can be used moving forward.

shared/aries-core/src/js/components/core/NavigationMenu/NavigationMenu.tsx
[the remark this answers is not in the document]
Sulaymon333: `Documentation.md` will be updated after @halocline validates the new prop introduction.

apps/docs/src/components/content/BrandIdentity.js
Sulaymon333: @julauxen confirms offline that the logo looks good. The next t-short size up (`3xsmall`) looks quite big and the smaller size(`5xsmall`) is too small.

apps/docs/src/layouts/navigation/Navigation.js
Sulaymon333: `legacyBehavior` restores the pre-Next.js 13 <Link> behavior where <Link> does not render its own <a> element - instead it passes href/onClick to its child via `cloneElement` and attaches a ref via React's `ref` forwarding. More info here https://nextjs.org/docs/13/pages/api-reference/components/link#legacybehavior.

> Should be be using it?

Yes, with `legacyBehavior`, Link doesn't render `<a>` - it passes href/onClick/ref directly to `BrandIdentity/AppIdentity`, which applies them to its <Button>. No nesting.

Alternative is that we can drop it but we will ned to refactor `BrandIdentity/AppIdentity` to not render its own anchor (remove href from `<Button>` and use onClick + router.push instead), but that's a bigger change across the codebase. Given we're on Next.js 15 with the Pages Router, `legacyBehavior` looks like the correct pattern for wrapping custom components that already render anchors.

apps/docs/src/components/content/BrandIdentity.js
Sulaymon333: this is addressed via this [commit](https://github.com/grommet/hpe-design-system/pull/6144/changes/7ae257eb835d6053a504ebc976335c25fb67ca26)

shared/aries-core/src/js/components/core/NavigationMenu/NavigationMenu.tsx
```diff
@@ -64,7 +64,9 @@ interface NavigationMenuProps extends BoxProps {
   activeItem?: string;
   header?: React.ReactNode;
   items: NavItemType[];
+  logo?: React.ReactNode;
```
halocline: @Sulaymon333 - You shouldn't need to alter any properties in `aries-core` components. Try to think of them as an external library you are consuming but don't have direct access to. How do you use the components with the capabilities they make available? Eventually I can see a version of `aries-core` graduating to be an HPE Design System opinionated component library used alongside Grommet. With this mental model, changes to the APIs of these components should be heavily considered.

The existing `header` and `open` props should provide all functionality required to meet the requirements. Neither `logo` nor `onToggle` are needed.

cc: @jcfilben @britt6612 @SOjaHPE

shared/aries-core/src/js/components/core/NavigationMenu/NavContainer/NavContainer.tsx
```diff
@@ -4,6 +4,7 @@ import { NavHeader } from './NavHeader';
 interface NavContainerProps {
   children: React.ReactNode;
   header?: React.ReactNode;
+  logo?: React.ReactNode;
```
halocline: Not needed. Don't reach into `aries-core` components.

apps/docs/src/layouts/navigation/Navigation.js
halocline: ☝️ 

`legacyBehavior` is needed and proper here.

apps/docs/src/layouts/navigation/NavContext.js
halocline: NavContext seems a bit heavy. Can we get away with just a state variable within Layout.js?

It might be necessary if it avoids deep prop drilling, but I think we might be able to get away with a state variable if we update AppIdentity.

halocline (commented):
It would be nice if we could just use the existing AppIdentity instead of BrandIdentity and not need to create what is essentially the same thing. I would support just modifying it.

There are a few site examples that also consume this, however the site examples can be updated to `import { AppIdentity } from '@shared/aries-core'` instead so that they have parity with their existing state and then should feel free to alter the AppIdentity that lives in /components/content.

shared/aries-core/src/js/components/core/NavigationMenu/NavigationMenu.tsx
Sulaymon333: The changes to `aries-core` components are now reverted via this [commit](https://github.com/grommet/hpe-design-system/pull/6144/changes/f2b2caf0be360060eece9450b13539e9029e70ae). And `header` and `open` props are used to meet the requirements.

sorry, seems to be that I misunderstood that we can update the `aries-core` for this issue. my understanding was that you gave a green light offline when I showed the `onToggle` prop that `aries-core` component can be updated based on this issue need.

Yes sure, `header` and `open` props can be used to meet the same requirement like it was implemented in PR #6141

shared/aries-core/src/js/components/core/NavigationMenu/NavContainer/NavContainer.tsx
Sulaymon333: This is reverted via this [commit](https://github.com/grommet/hpe-design-system/pull/6144/changes/f2b2caf0be360060eece9450b13539e9029e70ae) and provided `header` and `open` props are used to meet the requirements.

Sulaymon333:
> It would be nice if we could just use the existing AppIdentity instead of BrandIdentity and not need to create what is essentially the same thing. I would support just modifying it.
> 
> There are a few site examples that also consume this, however the site examples can be updated to `import { AppIdentity } from '@shared/aries-core'` instead so that they have parity with their existing state and then should feel free to alter the AppIdentity that lives in /components/content.

**AppIdentity vs BrandIdentity - Done.** 
Dropped `BrandIdentity` and modified the local `AppIdentity` in `/components/content` to support a new `hpe2` brand (theme-aware HPE logo via Image). Site examples were updated to import from `@shared/aries-core` so they remain unaffected. This is addressed via this [commit](https://github.com/grommet/hpe-design-system/pull/6144/changes/06331d0e4ce1e8986da7252b7b6dabdf375b641e)

Sulaymon333:
> NavContext seems a bit heavy. Can we get away with just a state variable within Layout.js?
> It might be necessary if it avoids deep prop drilling, but I think we might be able to get away with a state variable if we update AppIdentity.

A dedicated NavContext looks like overkill - agreed. But pure `useState` + props in `Layout.js` doesn't work because the homepage's `<AppHeader />` sits inside page content (a Layout child), not as a direct sibling. Leveraging the existing `ViewContext` is the minimal solution that bridges both cases without introducing new abstractions. This is implemented via this [commit](https://github.com/grommet/hpe-design-system/pull/6144/changes/06331d0e4ce1e8986da7252b7b6dabdf375b641e). More context below. 

I initially agreed and tried just `useState` in `Layout.js` with props. However, `useState` alone in `Layout.js` doesn't cover the home page case.

**The issue:** `pages/index.js` renders its own `<AppHeader />` inside the page content (as a child of <Layout>) because the homepage header shares the hero background image. Layout conditionally skips its own AppHeader for the home route. This means the homepage's AppHeader lives below Layout in the tree and cannot receive navOpen via props without threading it through children - which would require either:

- A render prop / cloneElement pattern (fragile)
- Deep prop drilling through `<Main>` → page content → `<AppHeader />`
- Or a context

**The solution:** since ViewContext already exists in `_app.js` and is consumed by `Layout.js` for contentHistory/pageUpdateReady, I added navOpen/setNavOpen there. This avoids creating a new context (NavContext is now dropped), avoids prop drilling, and lets both the Layout-rendered AppHeader and the homepage's standalone AppHeader read the same state.

apps/docs/src/components/content/BrandIdentity.js
```diff
@@ -0,0 +1,63 @@
+import React, { forwardRef, useContext } from 'react';
```
halocline: Is this being consumed anywhere? I'd expect this file to not exist if we are using AppIdentity.

apps/docs/src/components/content/AppIdentity.js
```diff
@@ -37,9 +57,11 @@ export const AppIdentity = forwardRef(
...
-                {brands[brand].name}
-              </Text>
+              {brands[brand]?.name && (
+                <Text weight="bold" size={textSize} color="text-strong">
+                  {brands[brand].name}
+                </Text>
+              )}
               <Text size={textSize} color="text-strong">
                 {title}
               </Text>
```
halocline, line 56: Since this is now only for the DS site, we can eliminate the extra dynamic brand stuff.

```suggestion
import React, { forwardRef, useContext } from 'react';
import PropTypes from 'prop-types';
import { Box, Button, Image, Text } from 'grommet';
import { ThemeContext } from 'styled-components';

const HpeLogo = () => {
  const theme = useContext(ThemeContext);
  return (
    <Box width="4xsmall">
      <Image
        alt="HPE logo"
        fit="contain"
        src={theme?.dark
          ? '/HPE_logo_full-clr_rev_rgb.svg'
          : '/HPE_logo_full-clr_pos_rgb.svg'}
      />
    </Box>
  );
};

export const AppIdentity = forwardRef(
  (
    { 
      logo = true, 
      logoOnly = false,
      title = 'HPE Design System', 
      ...rest 
    },
    ref,
  ) => {
    const textSize = 'medium';
    const href = '/';
    const linkLabel = title ? `Go to ${title} home` : 'Go to home';

    return (
      <Button href={href} ref={ref} a11yTitle={linkLabel} {...rest}>
        <Box
          direction="row"
          align="center"
          gap="xsmall"
          // pad maintains accessible hit target
          // non-responsive maintains same dimensions for mobile
          pad={{ vertical: 'xsmall' }}
          responsive={false}
        >
          {logo && <HpeLogo />}
          {!logoOnly && (
            <Box direction="row" gap="3xsmall" wrap>
              {!logo && 
                <Text weight="bold" size={textSize} color="text-strong">
                  HPE
                </Text>
              }
              <Text size={textSize} color="text-strong">
                {title}
              </Text>
            </Box>
          )}
        </Box>
      </Button>
    );
  },
);

AppIdentity.propTypes = {
  logo: PropTypes.bool,
  logoOnly: PropTypes.bool,
  title: PropTypes.string,
};
```

apps/docs/src/layouts/navigation/Navigation.js
jcfilben: We may want to add a comment as support for legacyBehavior has been dropped in the latest version of next.js (v16)

apps/docs/src/components/content/BrandIdentity.js
jcfilben: When we get discrepancies between design and code it may signal that something isn't working correctly either in code or in figma. Might be worth looking into more. Ideally our component sizes should be the same in figma and code. 
cc: @julauxen

apps/docs/src/layouts/main/AppHeader.js
```diff
@@ -1,18 +1,20 @@
...
 } from 'grommet';
+import Link from 'next/link';
 import { Search as SearchIcon } from '@hpe-design/icons-grommet';
 import { ThemeModeToggle, AppIdentity } from '../../components';
 import { Search } from '../navigation';
+import { ViewContext } from '../../pages/_app';
 
 const StyledHeader = ({ ...rest }) => {
   const [showSearch, setShowSearch] = useState(false);
+  const { navOpen } = useContext(ViewContext);
```
halocline: I now see why you are reaching for context. AppHeader is consumed in Layout.js and in /page/index.js -- which is not ideal. It would be nice if we hadn't introduced a unique app layout for the home page which creates bifurcation... but alas, that is where we are. Leave as is for now.

Prior to realizing this, I was hoping we could just have the state in Layout.js.

RE: ViewContext - Now that I understand why you need context -- I prefer your NavContext approach for clarity of purpose. I'm not really a fan of this ViewContext -- its too difficult to figure out what it is and what its intended purpose is -- too ambiguous.

apps/docs/src/components/content/BrandIdentity.js
Sulaymon333: it is cleaned up via this [commit](https://github.com/grommet/hpe-design-system/pull/6144/changes/ad6dd05b6a6f6eb388f336d5cf29667cde31e42c)

apps/docs/src/layouts/main/AppHeader.js
Sulaymon333: NavContext is now restored via this [commit](https://github.com/grommet/hpe-design-system/pull/6144/changes/b8455cc2f10499813370b3e280e3259daaff17d7)

apps/docs/src/components/content/AppIdentity.js
Sulaymon333: @halocline is it intentional to keep the text brand of `**HPE**` along side the text `Design System`? This is different from the design spec.

image

<br/>

**Design spec**
image

apps/docs/src/layouts/navigation/Navigation.js
Sulaymon333: I added a [commit](https://github.com/grommet/hpe-design-system/pull/6144/changes/7b3629b5accd784a2640527705d34b566005a840) to add comment two relevant files using `legacyBehavior` in this PR. Adding comment to other instances of `legacyBehavior` usage across the codebase can be updated via a follow on ticket to keep this ticket scoped.

jcfilben: Sounds good, thanks

apps/docs/src/components/content/AppIdentity.js
halocline: I did that intentionally. Need to discuss with @julauxen, but I think we should maintain the "HPE Design System" branding no matter what.

Also, I noticed that the HPE logo in the menu's header is clickable and navigable. Is that what we want @julauxen? We have the "home" link which is equivalent and more obvious in this context.

Sulaymon333: Yes, the logo being clickable and navigable seems to be a common approach for many websites nowadays. I understand the concern that we have "Home" link as well but it seems to me that it serves another purpose to indicate current route the user is currently on in addition to being navigable/clickable.

For mobile, that means users do not need to open side panel to go home. They can simply click the logo and go to home page.

halocline: **"the logo being clickable and navigable"** in the AppHeader - yes.

In the NavMenu's header... ?

halocline (approved):

apps/docs/src/components/content/AppIdentity.js
julauxen: @halocline I do not see the logo being clickable as a usability issue. Since we are making the logo clickable in the header we might as well keep it clickable on the nav header to not give the same item different behaviours depending on circumstance (that could generate confusion). 

On the branding, I removed the HPE in front of the text in the design because with the logo we keep that context all the time (the logo is either in nav or header within our website) - seemed redundant. Also, if we close the nav and the logo is moved to the header, then the expectations is to have HPE (logo) + HPE Design System (text)? I'm open to your thoughts here though, not a super firm opinion.

julauxen: I just saw the current implementation and it looks like a bug to have the logo substitute the HPE text when nav is closed. It changes and it calls way to much attention. I would either:

- Remove the HPE text like the design
- Keep the logo + text, so when nav is open we see HPE Design System (text) and when it's closed we see HPE (logo) + HPE Design System (text)

I still think option one is slightly better.

https://github.com/user-attachments/assets/525e0b39-5a7f-42d7-a045-f1dcf1672512

julauxen: @SOjaHPE any thoughts here about our branding?

halocline: What about an animation to ease the transition?

- Having two HPE logos in the same layout seems excessive. I believe we want to avoid that.
- Not having HPE as part of the HPE Design System branding turns HPE Design System into a generic design system and loses all brand association. Maintaining consistent branding for the HPE Design System is important for establishing the system as HPE's endorsed and supported design system.

julauxen (approved):
looks good to me

shared/aries-core/src/js/components/core/NavigationMenu/NavigationMenu.tsx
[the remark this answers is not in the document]
Sulaymon333: no longer required. prop was removed

shared/aries-core/src/js/components/core/NavigationMenu/NavigationMenu.tsx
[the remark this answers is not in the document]
Sulaymon333: no longer required. prop was removed

apps/docs/src/components/content/BrandIdentity.js
Sulaymon333: resolved

apps/docs/src/components/content/AppIdentity.js
Sulaymon333: animation is now added to the HPE (text) to HPE (logo) and vice versa.

Issue #6078, opened 2026-04-09
[Docs] Site Navigation - MVP success criteria


halocline:
### MVP criteria:

- [x] NavigationMenu presented in site's app layout
- [x] Navigation menu presents top-level hub pages and all of their children
  - [x] Children should be grouped by their category and display the category as a subheading
  - [x] Category order can be specified
- [x] Top-level hub page menu items removed from app header
- [ ] HPE branding moved and HPE web common header/footer service removed #6107 
- [ ] Scroll and sticky header behaviour #6108 
- [ ] Resolve issue where sticky header obscures focused nav items (https://github.com/grommet/hpe-design-system/issues/5658)
- [ ] Free of any critical A11y issues #6079 

### Future enhancements:
- [ ] Spacing and alignment polishing
- [ ] Typography polishing
- [ ] Collapsible icon fill
- [ ] Collapsible icon animation
- [ ] Expand / collapse animations for menu container
- [ ] Expand / collapse animations for expandable menu items
- [ ] Navigation menu title (?)

Wish list enhancements:
- [ ] [your enhancement here]

julauxen:
Things to consider:

Chris mentioned that if we add the HPE logo to the nav or our own header, we might be able to get rid of the marketing global header, like [this.](https://www.figma.com/design/lV4lc9Kbk0giOS20k9sbrJ/DS-Website-UI-updates?m=auto&node-id=2722-8295&t=K1hZGd0L5r2VJIWj-1)

The side nav should scroll and keep it to the viewport, instead of filling the height.

CillianHPE:
**Overflow & scroll behavior**
When opening the Components menu, the navigation grows in height and forces the entire page to scroll to see lower items.
**Suggestion:** Constrain the navigation to the viewport height and apply overflow scrolling to the navigation body instead. This might have been alluded to already in this tracking issue, but calling it out again here.

**Nav Header**
Remove right‑side padding in the nav header so the toggle button aligns visually with the right edge of the nav items.
<img width="486" height="909" alt="Image" src="https://github.com/user-attachments/assets/47c93cd2-6f34-4fab-aacf-79dddae376b7" />

**Animation speed**
The open/close animation feels a bit slow. There’s an opportunity to polish this further.
**Suggestion:** Reuse or build on the animation exoloration work explored during recent innovation sprints. I’d love to help with this. If we can set up the navigation in a CodePen, I can tweak easing, speed, and opacity tokens to dial it in.

[...]

PR #6178, opened 2026-05-14
[Docs App] Side nav mobile

### LINKED ISSUE #6177: Site navigation - Add mobile responsiveness to Side Navigation Layout
### Context
This ticket builds on PR #6107. So new PR should be opened against PR #6107 or `DOCS-NEXT` branch once #6107 is merged into it.

The current left navigation layout is not working as expected. It is not responsive on mobile screens. See preview for #6107  https://deploy-preview-6144--keen-mayer-a86c8b.netlify.app/

### Goal
Mimic the responsive behaviour of the left navigation for storybook example. [See here.](https://hpe-design-storybook.netlify.app/?path=/story/patterns-navigation--navigation)

### Success Criteria
- Behavior is consistent across screen sizes (mobile and desktop)
- HPE logo is visible on mobile

Sulaymon333:
[Deploy Preview](https://deploy-preview-6178--keen-mayer-a86c8b.netlify.app/)

#### What does this PR do?

- Behavior is consistent across screen sizes (mobile and desktop)
- HPE logo is visible on mobile

Related PR - #6190 

closes #6177


apps/docs/src/layouts/main/AppHeader.js
```diff
@@ -22,8 +29,21 @@ const StyledHeader = ({ ...rest }) => {
...
         >
+          {/* Show HPE logo + title in header when nav is collapsed */}
+          {/* legacyBehavior restores pre-Next.js 13 behavior where Link
+              does not render its own <a>, instead passing href/onClick to
+              the child via cloneElement and attaches a ref via React's ref 
+              forwarding. Required here because AppIdentity already renders 
+              an anchor via its <Button href>. This prop has been dropped 
+              in Next.js 13.15.11+, so this will need to be  refactored when 
+              upgrading. More details:
+              https://nextjs.org/docs/13/pages/api-reference/components/link#legacybehavior */}
```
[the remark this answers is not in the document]
Sulaymon333: sorted in previous PR #6144

apps/docs/src/layouts/navigation/SideNavHeader.js
```diff
@@ -0,0 +1,64 @@
...
+            flex={false}
+        >
+            {/* legacyBehavior restores pre-Next.js 13 behavior where Link
+              does not render its own <a>, instead passing href/onClick to
+              the child via cloneElement and attaches a ref via React's ref 
+              forwarding. Required here because AppIdentity already renders 
+              an anchor via its <Button href>. This prop has been dropped 
+              in Next.js 13.15.11+, so this will need to be  refactored when 
+              upgrading. More details:
+              https://nextjs.org/docs/13/pages/api-reference/components/link#legacybehavior */}
```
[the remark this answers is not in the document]
Sulaymon333: sorted in previous PR #6144

apps/docs/src/components/home/Community.js
```diff
@@ -79,7 +79,7 @@ export const Community = ({ ...rest }) => {
           columns="medium"
           rows={[['auto', 'full']]}
           gap={{ row: 'xlarge', column: 'xsmall' }}
-          fill
+          fill="horizontal"
```
Sulaymon333, line 81: After side nav is introduced, there was an issue because the outer Layout Grid's `rows={['auto', 'flex']}` makes the main area stretch to fill viewport height. That height cascaded into Community's <Grid fill>, stretching it vertically and displacing the Feedback icon. 

Changing `fill → fill="horizontal"` prevents the vertical stretch while keeping the two-column layout spanning full width.

**Before fix**

Image

**After** fix

Image

apps/docs/src/layouts/main/AppHeader.js
```diff
@@ -24,22 +27,32 @@ const StyledHeader = ({ ...rest }) => {
...
+              />
+            )}
+            {/* Show HPE logo + title in header when nav is collapsed */}
+            {/* legacyBehavior restores pre-Next.js 13 behavior where Link
+                does not render its own <a>, instead passing href/onClick to
+                the child via cloneElement and attaches a ref via React's ref 
+                forwarding. Required here because AppIdentity already renders 
+                an anchor via its <Button href>. This prop has been dropped 
+                in Next.js 13.15.11+, so this will need to be  refactored when 
+                upgrading. More details:
```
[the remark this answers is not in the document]
Sulaymon333: addressed in previous PR

britt6612:
If a user opens the mobile drawer, then resizes to desktop, and back to mobile, the drawer would re-open immediately I think we would want this to be reset.



https://github.com/user-attachments/assets/0ba7bdad-6cbe-485b-a4b6-bed0ffbba77b

britt6612:
is there a reason the `tip` is not also used in the mobile version 


https://github.com/user-attachments/assets/9e59aa7e-f3f6-4b80-89b1-14e71309dbfa

apps/docs/src/layouts/navigation/Navigation.js
```diff
@@ -9,24 +9,48 @@ import { SideNavHeader } from './SideNavHeader';
...
   const handleSelect = ({ item, event }) => {
     event.preventDefault();
     router.push(item.url);
     sendAnalytics({ type: 'navItemClick', href: item.url });
+    if (isMobile) {
+      setMobileNavOpen(false);
+    }
   };
 
-  return (
```
britt6612, line 22: // On mobile, render nothing inline; show a full-screen Layer instead
  if (isMobile && !mobileNavOpen) return null;


should we move this check above here

apps/docs/src/layouts/navigation/Navigation.js
```diff
@@ -9,24 +9,48 @@ import { SideNavHeader } from './SideNavHeader';
...
-      header={<SideNavHeader open={navOpen} setOpen={setNavOpen} />}
-      open={navOpen}
+      header={<SideNavHeader open={isOpen} setOpen={setNavOpen} />}
+      open={isOpen}
       onSelect={handleSelect}
       {...rest}
     />
   );
+
+  // On mobile, render nothing inline; show a full-screen Layer instead
```
britt6612, line 26: then here we can get rid of the 

```
    if (!mobileNavOpen) return null;
```

apps/docs/src/layouts/navigation/NavContext.js
```diff
@@ -1,16 +1,49 @@
...
     const [navOpen, setNavOpen] = useState(true);
-    const value = useMemo(() => ({ navOpen, setNavOpen }), [navOpen]);
+    // Mobile drawer visibility state (default: closed)
+    const [mobileNavOpen, setMobileNavOpen] = useState(false);
+
+    const toggleNav = useCallback((val) => {
+        setNavOpen(val);
+    }, []);
+
+    const toggleMobileNav = useCallback((val) => {
```
britt6612: why do we need the `useCallback` ?

can we remove and just have the `useMemo`?

```
    const value = useMemo(
        () => ({
            navOpen,
            setNavOpen,
            mobileNavOpen,
            setMobileNavOpen,
            isMobile,
        }),
        [navOpen, mobileNavOpen, isMobile],
    );
```

julauxen (approved):
i'm approving assuming brittany's comment [here](https://github.com/grommet/hpe-design-system/pull/6178#issuecomment-4480985104) will be worked on. No other comments on my side.

apps/docs/src/layouts/navigation/NavContext.js
Sulaymon333: Good catch, the `useCallback` is not adding value in this case since it is only wrapping setState which react guarantee to have stable identity across re-render.

> The set function has a stable identity, so you will often see it omitted from Effect dependencies, but including it will not cause the Effect to fire.

More here on [react docs](https://react.dev/reference/react/useState#setstate-caveats)

Sulaymon333:
> i'm approving assuming brittany's comment [here](https://github.com/grommet/hpe-design-system/pull/6178#issuecomment-4480985104) will be worked on. No other comments on my side.

@julauxen all @britt6612 comments are addressed via this [commit](https://github.com/grommet/hpe-design-system/pull/6178/changes/b45acc69518a15bc649b92f37add8cbc9f48d8f8)

britt6612:
https://github.com/user-attachments/assets/91e44e74-9972-48cb-ae93-3aa509944ade

@Sulaymon333 should the focus in mobile view also stay on the close button when the nav is open? see desktop behavior

shared/aries-core/src/js/components/core/NavigationMenu/NavContainer/NavContainer.tsx
```diff
@@ -29,7 +29,7 @@ export const NavContainer = ({
     global: { edgeSize: Record<string, string> };
   };
   return (
-    <Box {...rest} width={open ? 'small' : undefined}>
+    <Box  width={open ? 'small' : undefined} {...rest}>
```
Sulaymon333: this fixes a regression bug where the layer width is not full because the `...rest` prop is being spread before the `width`. Moving the `...rest` prop after the width prop fixed the width to be full again. see attached images below.

**Before fix**

Image

**After fix**

Image

Sulaymon333:
> Screen.Recording.2026-05-19.at.10.27.13.AM.mov 
> @Sulaymon333 should the focus in mobile view also stay on the close button when the nav is open? see desktop behavior

this is addressed via this [commit](https://github.com/grommet/hpe-design-system/pull/6178/changes/4354b522187dbc43deb81c6667362da85b014296) with the possibility to set it as active but the first item (logo) is focusable as well. So, I did not see much difference aside visual ring around it. 

Unlike in the [storybook nav story](https://hpe-design-storybook.netlify.app/?path=/story/patterns-navigation--navigation-with-subheadings), the first item on the sidenav is a text so that issue is not present.

**Update**
`useRef`/ `setTimeOut` is used to achieve the focus on sidenav close button on mobile. since Grommet Layer handles focus in such a way that first focusable item in the layer receives the focus on mount. `tabInxex=-1` on the first focusable item inside the layer, the logo or `autoFocus` on the close buton itself won't work.

apps/docs/src/layouts/navigation/MobileNavHeader.js
britt6612: @Sulaymon333 some of the spacing here looks weird

apps/docs/src/layouts/navigation/MobileNavHeader.js
```diff
@@ -6,6 +7,18 @@ import { AppIdentity } from '../../components';
 
 export const MobileNavHeader = ({ ...rest }) => {
     const { setMobileNavOpen } = useNavState();
+    const buttonRef = useRef(null);
+
+    // Grommet L
```
britt6612: lets fix this format

```
  // Grommet Layer's FocusedContainer focuses the first focusable
  // descendant on mount. setTimeout defers our .focus() call to run
  // after Layer's focus management completes.

```

apps/docs/src/layouts/navigation/MobileNavHeader.js
Sulaymon333: image

apps/docs/src/layouts/navigation/MobileNavHeader.js
Sulaymon333: fixed

apps/docs/src/components/home/Community.js
britt6612, line 81: @Sulaymon333 this feels like were not solving the actual problem? can we revert this back?

britt6612: in community.js in the grid can we not change the `rows` there I think we can just remove that line? I have this locally can you please try and test 

Image

apps/docs/src/layouts/main/Layout.js
```diff
@@ -121,9 +122,10 @@ export const Layout = ({
               as="aside"
               a11yTitle="Site navigation"
               background="background-front"
-              height="100vh"
+              height={isMobile ? undefined : '100vh'}
+              fill={isMobile || undefined}
               overflow="auto"
-              style={{
+              style={isMobile ? undefined : {
```
britt6612: are we trying to make the header "non sticky" in mobile?

apps/docs/src/layouts/navigation/MobileNavHeader.js
```diff
@@ -0,0 +1,63 @@
...
+                <AppIdentity logoOnly />
+            </Link>
+            <Button
+                ref={buttonRef}
+                active
+                aria-controls="navigation-menu"
+                aria-expanded
+                alignSelf="start"
+                a11yTitle="Close navigation"
+                icon={
```
britt6612: @sulaymon-tajudeen-hpe 

I feel like this file still is not formatted correctly?

```


import { useEffect, useRef } from 'react';
import Link from 'next/link';
import { Button, Header } from 'grommet';
import { Sidebar } from '@hpe-design/icons-grommet';
import { useNavState } from './NavContext';
import { AppIdentity } from '../../components';

export const MobileNavHeader = ({ ...rest }) => {
  const { setMobileNavOpen } = useNavState();
  const buttonRef = useRef(null);

  // Grommet Layer's FocusedContainer focuses the first focusable
  // descendant on mount. setTimeout defers our .focus() call to run
  // after Layer's focus management completes.
  useEffect(() => {
    const timer = setTimeout(() => {
      buttonRef.current?.focus();
    }, 0);
    return () => clearTimeout(timer);
  }, []);

  return (
    <Header
      pad={{
        left: 'medium',
        right: 'small',
        vertical: 'xsmall',
      }}
      direction="row"
      align="center"
      justify="between"
      {...rest}
    >
      {/* legacyBehavior restores pre-Next.js 13 behavior where Link
              does not render its own <a>, instead passing href/onClick to
              the child via cloneElement and attaches a ref via React's ref 
              forwarding. Required here because AppIdentity already renders 
              an anchor via its <Button href>. This prop has been dropped 
              in Next.js 13.15.11+, so this will need to be  refactored when 
              upgrading. More details:
              https://nextjs.org/docs/13/pages/api-reference/components/link#legacybehavior */}
      <Link href="/" passHref legacyBehavior>
        <AppIdentity logoOnly />
      </Link>
      <Button
        ref={buttonRef}
        active
        aria-controls="navigation-menu"
        aria-expanded
        alignSelf="start"
        a11yTitle="Close navigation"
        icon={<Sidebar aria-hidden style={{ rotate: '180deg' }} />}
        onClick={() => setMobileNavOpen(false)}
        tip="Close navigation"
      />
    </Header>
  );
};

```

apps/docs/src/layouts/main/Layout.js
Sulaymon333: yes correct.

apps/docs/src/layouts/navigation/MobileNavHeader.js
Sulaymon333: I did not see this formatting issue locally. see below attached
image

apps/docs/src/components/home/Community.js
Sulaymon333: I removed the `rows` prop for the grid via this [commit](https://github.com/grommet/hpe-design-system/pull/6178/changes/235bc0abecfc0b74ed3d6eabe9776886bf99fe2b)

apps/docs/src/layouts/main/Layout.js
britt6612: it still looks sticky to me not sure if this is a issue though or if its fine to be sticky on both mobile and desktop

https://github.com/user-attachments/assets/6c055b57-c0f7-4a03-b35e-fb15689e53fc

apps/docs/src/layouts/navigation/MobileNavHeader.js
britt6612: look at the `icon` with the `Sidebar` it looks like its indented 4 spaces vs what I had locally 

```
 icon={<Sidebar aria-hidden style={{ rotate: '180deg' }} />}
```

apps/docs/src/layouts/navigation/Navigation.js
```diff
@@ -1,24 +1,52 @@
...
+    }
   };
 
+  // On mobile, render nothing inline; show a full-screen Layer instead
+  if (isMobile && !mobileNavOpen) return null;
+
+  if (isMobile) {
+    return (
+      <Layer
+        onClickOutside={() => setMobileNavOpen(false)}
```
britt6612, line 32: should we add a `aria-label` to better announce to the user. 
```
  aria-label="Site navigation"
```

so it announces "Site navigation, dialog"

apps/docs/src/layouts/main/Layout.js
julauxen: why are we trying to make the header non-sticky? what's the rationale here? I would expect it to be sticky, working like the implementation on desktop

britt6612: ☝️ thats what I wasn't sure about why this check for mobile is here

Sulaymon333: `position: sticky` is a desktop-specific concern - it keeps the sidebar pinned while the main content scrolls. On mobile, the nav renders inside a Layer (fixed-position portal), where sticky has no scroll ancestor and serves no purpose. The mobile nav header stays fixed naturally via NavContainer's layout structure (header is a sibling above the scrollable `<Box overflow="auto">`, not inside it). We intentionally don't pass desktop layout props to mobile to keep the separation clean and avoid leaking irrelevant styles into the Layer context.

so bottom line we don't need the `position: sticky` on mobile hence it is NOT needed to be passed down.

I added a comment for clarity

cc: @britt6612 @julauxen

apps/docs/src/layouts/navigation/MobileNavHeader.js
Sulaymon333: hmm, I did not notice the same

apps/docs/src/layouts/navigation/Navigation.js
Sulaymon333: `a11yTitle` is already covered in `Layout.js` via [Line 127](https://github.com/grommet/hpe-design-system/pull/6178/changes#diff-70a4413a4700d95e1fb8ce95f827c456c8def1b9ba8150908ffcf98aeb3448d4L127) and it should be passed down via `...rest` prop right? Do we need to add another one?

apps/docs/src/layouts/main/Layout.js
britt6612: @Sulaymon333 thank you for that makes sense!

apps/docs/src/layouts/navigation/MobileNavHeader.js
britt6612: not sure Im following? You can see in the code you pasted the Sidebar icon is indented 4 spaces from the `icon` prop

apps/docs/src/layouts/navigation/Navigation.js
britt6612: Good catch, but doesnt the `a11yTitle` via `...rest` reaches NavigationMenu (the inner <nav>), not the `Layer` itself they're separate elements. For the Layer to announce "Site navigation, dialog" on open, `aria-label` would need to be added directly to Layer?

apps/docs/src/layouts/navigation/Navigation.js
Sulaymon333: addressed via this [commit](https://github.com/grommet/hpe-design-system/pull/6178/changes/b45acc69518a15bc649b92f37add8cbc9f48d8f8#diff-44f9818e1740abdfd69d4105c9e450654fe8da70a3a473a7c5904c8e28fb0208)

apps/docs/src/layouts/navigation/Navigation.js
Sulaymon333: addressed via this [commit](https://github.com/grommet/hpe-design-system/pull/6178/changes/b45acc69518a15bc649b92f37add8cbc9f48d8f8#diff-44f9818e1740abdfd69d4105c9e450654fe8da70a3a473a7c5904c8e28fb0208)

apps/docs/src/layouts/navigation/MobileNavHeader.js
Sulaymon333: this is fixed now via this [commit](https://github.com/grommet/hpe-design-system/pull/6178/changes/afd3b4ffdc4686a5661d5558ecf7f58f28288a2f).

apps/docs/src/layouts/navigation/Navigation.js
Sulaymon333: this is addressed via this [commit](https://github.com/grommet/hpe-design-system/pull/6178/changes/5dcd31b9d26a7f62268e9f290022c95c02db8977)

apps/docs/src/layouts/navigation/MobileNavHeader.js
```diff
@@ -0,0 +1,58 @@
...
+              an anchor via its <Button href>. This prop has been dropped 
+              in Next.js 13.15.11+, so this will need to be  refactored when 
+              upgrading. More details:
+              https://nextjs.org/docs/13/pages/api-reference/components/link#legacybehavior */}
+      <Link href="/" passHref legacyBehavior>
+        <AppIdentity logoOnly />
+      </Link>
+      <Button
+        ref={buttonRef}
+        active
```
britt6612, line 47: just want to clarify we want this button to always be `active` ? Is this for styling reasons?

@sulaymon-tajudeen-hpe @julauxen

apps/docs/src/layouts/main/Layout.js
```diff
@@ -105,29 +106,34 @@ export const Layout = ({
...
+                Layer (fixed-position portal) where sticky has no
+                scroll ancestor and serves no purpose. */}
             <Navigation
               gridArea="nav"
               as="aside"
               a11yTitle="Site navigation"
               background="background-front"
-              height="100vh"
+              height={isMobile ? undefined : '100vh'}
+              fill={isMobile || undefined}
```
britt6612:
```suggestion
              fill={isMobile}
```

shared/aries-core/src/js/components/core/NavigationMenu/NavContainer/NavContainer.tsx
britt6612: this current fix also silently changes prop precedence for all the callers of NavContainer. Can we in Navigation.js just remove the `...rest` from the NavigationMenu and add `fill` for the mobile

```
import { useAnalytics, Layer } from 'grommet';
import { useRouter } from 'next/router';
import { NavigationMenu } from '@shared/aries-core';
import { navItems } from './navItems.ts';
import { useNavState } from './NavContext';
import { MobileNavHeader } from './MobileNavHeader';
import { SideNavHeader } from './SideNavHeader';

export const Navigation = ({ ...rest }) => {
  const router = useRouter();
  const sendAnalytics = useAnalytics();
  const { navOpen, setNavOpen, isMobile, mobileNavOpen, setMobileNavOpen } =
    useNavState();

  const activeItem = router.pathname;

  const handleSelect = ({ item, event }) => {
    event.preventDefault();
    router.push(item.url);
    sendAnalytics({ type: 'navItemClick', href: item.url });
    if (isMobile) {
      setMobileNavOpen(false);
    }
  };

  // On mobile, render nothing inline; show a full-screen Layer instead
  if (isMobile && !mobileNavOpen) return null;

  if (isMobile) {
    return (
      <Layer
        onClickOutside={() => setMobileNavOpen(false)}
        aria-label="Site navigation"
        position="left"
        onEsc={() => setMobileNavOpen(false)}
      >
        <NavigationMenu
          activeItem={activeItem}
          items={navItems}
          header={<MobileNavHeader />}
          open
          onSelect={handleSelect}
          fill
        />
      </Layer>
    );
  }

  return (
    <NavigationMenu
      activeItem={activeItem}
      items={navItems}
      header={<SideNavHeader open={navOpen} setOpen={setNavOpen} />}
      open={navOpen}
      onSelect={handleSelect}
      {...rest}
    />
  );
};

```

apps/docs/src/layouts/navigation/MobileNavHeader.js
Sulaymon333: The implementation follows hpe design storybook [implementation](https://hpe-design-storybook.netlify.app/?path=/story/patterns-navigation--navigation&globals=viewport.value:mobile1). We can remove the active prop here. no strong opinion 

cc: @julauxen

shared/aries-core/src/js/components/core/NavigationMenu/NavContainer/NavContainer.tsx
Sulaymon333: Having `...rest` before other props on `NavContainer` makes the side panel of [hpe design navigation story ](https://hpe-design-storybook.netlify.app/?path=/story/patterns-navigation--navigation&globals=viewport.value:mobile1)panel not to have full width again. see attached image below. 

```js
    <Box {...rest}  width={open ? 'small' : undefined}> // causes story mobile width layer NOT to be full
```

```js
    <Box  width={open ? 'small' : undefined} {...rest}> // allows story mobile width layer to be FULL
```

can you share the issue you observed with "all" callers of NavContainer. I noticed it is only NaviagtionMenu.tsx that is only using NavContainer throughout the codebase and NavContainer is not meant to be used outside `aries-core`.  It's not part of the public API surface of the package.

image

britt6612: Yes the "all callers" was a overstated statement. However in the `NavigationPanel.tsx` we are passing `  width={undefined}` to explicitly opt out of the fixed width. This pattern only works when the `...rest` comes **after** the internal default.

britt6612: > Having ...rest before other props on NavContainer makes the side panel of [hpe design navigation story ](https://hpe-design-storybook.netlify.app/?path=/story/patterns-navigation--navigation&globals=viewport.value:mobile1)panel not to have full width again. see attached image below.
> 
>     <Box width={open ? 'small' : undefined} {...rest}> // causes story mobile width layer NOT to be full
>     <Box  width={open ? 'small' : undefined}> // allows story mobile width layer to be FULL


@sulaymon-tajudeen-hpe this did not work even when putting `fill` on the NavigationMenu?

Sulaymon333: sorry for the typo. I updated the comment and share the corrected code I meant to share earlier below


```js
    <Box {...rest}  width={open ? 'small' : undefined}> // causes story mobile width layer NOT to be full
```

```js
    <Box  width={open ? 'small' : undefined} {...rest}> // allows story mobile width layer to be FULL
```

britt6612 (approved):

apps/docs/src/layouts/navigation/MobileNavHeader.js
britt6612: I approved just want to verify this before merge

britt6612: So I think this is actually the correct behavior since we want to avoid confusion to the user it helps to distinguish between temporary press and a persistent active state.

MikeKingdom (commented):
Still reviewing, but I have a couple general comments.

1. The PR description needs more about what was done, how I can test what was done and what to expect. Currently the copilot description isn't really enough and I have to dig into a lot of the files and figure out what pages/areas are affected.

2. The original responsiveness was designed to work with the space available from the window (`@media`) breakpoints. We should consider wrapping the main area in `<Box responsive="container">` to basically preserve this behavior via container queries whether or not the nav is there or open.

MikeKingdom (approved):
In general, I think it looks good and am fine approving it.

I do have a few things I think don't make sense from a design perspective:

Regarding `active` when the nav is open in mobile mode, really in this case basically being a layer/dialog I would think X actually makes sense. You're no longer expanding/collapsing a side panel.

And the tip/aria terminology is a bit off IMHO. Non-mobile the icon expands/collapses the sidebar rather than open/close. For mobile it does open/close

The icon also seems backwards to me in every case, but that may just be me. To me it shows opening a sidebar to the right of a main area. And rotating it never makes sense since it never looks like that

apps/docs/src/layouts/navigation/MobileNavHeader.js
julauxen: Hey @britt6612 I think the active state makes sense here, specially because this specific tickets asks to mimic the storybook implementation. I do think we can consider other options for this, but I'll add as a future enhancement under this ticket #6078. This enhancement is not part of mvp.

julauxen:
> egarding `active` when the nav is open in mobile mode, really in this case basically being a layer/dialog I would think X actually makes sense. You're no longer expanding/collapsing a side panel.
> 
> And the tip/aria terminology is a bit off IMHO. Non-mobile the icon expands/collapses the sidebar rather than open/close. For mobile it does open/close
> 
> The icon also seems backwards to me in every case, but that may just be me. To me it shows opening a sidebar to the right of a main area. And rotating it never makes sense since it never looks like that

Hey @MikeKingdom added your notes to #6078 under "future enhancements" so we don't loose track of this info.

PR #6181, opened 2026-05-15
[docs] button page refactor 

### LINKED ISSUE #6164: Button - component restructure
Apply the refactor workflow to button.

julauxen:
[Deploy Preview](https://deploy-preview-6181--keen-mayer-a86c8b.netlify.app/)
button page: https://deploy-preview-6181--keen-mayer-a86c8b.netlify.app/components/button

#### What does this PR do?
Generated button yaml.
Restructures the button documentation to fit our template.

Closes #6164


apps/docs/src/examples/components/button/DosDonts/ButtonBadActionLabelsPreview.js
```diff
@@ -0,0 +1,34 @@
...
+      >
+        <CheckBox id="active-bad" name="active" label="Active" />
+      </FormField>
+      <FormField
+        htmlFor="inactive-bad"
+        name="inactive"
+        contentProps={{ border: false }}
+      >
+        <CheckBox id="inactive-bad" name="inactive" label="Inactive" />
+      </FormField>
```
halocline: Use CheckBoxGroup?

apps/docs/src/examples/components/button/ButtonActionLabelsExample.js
```diff
@@ -0,0 +1,10 @@
+import React from 'react';
+import { Box, Button } from 'grommet';
+
+export const ButtonActionLabelsExample = () => (
+  <Box direction="row" gap="xsmall" pad="medium" wrap>
```
halocline:
```suggestion
  <Box direction="row" gap="xsmall" wrap>
```

apps/docs/src/examples/components/button/ButtonBeginNewTaskExample.js
```diff
@@ -0,0 +1,132 @@
...
+export const ButtonBeginNewTaskExample = () => {
+  const [open, setOpen] = useState(false);
+
+  return (
+    <Box gap="medium" pad="medium" width="large">
+      <Box align="center" direction="row" justify="between">
+        <Box gap="xsmall">
+          <Text id="devices-heading" size="large" weight="bold">
+            Devices
+          </Text>
```
halocline: Use Heading component

apps/docs/src/examples/components/button/ButtonBeginNewTaskExample.js
```diff
@@ -0,0 +1,132 @@
...
+            {devices.length} items
+          </Text>
+        </Box>
+        <Button
+          label="Create device"
+          onClick={() => setOpen(true)}
+          primary
+        />
+      </Box>
+      <DataTable
```
halocline, line 48: @julauxen, @jcfilben , @britt6612 , @MikeKingdom , @Sulaymon333  -- What do you all think about this example implementation? I kind of feel like it should be implemented with Data & Friends. Thoughts?

If so, how might we enhance/instruct the examples creator agent to use Data and friends when dealing with any sort of data collection. This should likely be a /knowledge/core/skills that gets added.

apps/docs/src/pages/components/button.mdx
```diff
@@ -38,189 +46,105 @@ import {
...
+  height={{ min: 'small' }}
+>
+  <ButtonBusySimpleExample />
 </Example>
 
-### Buttons vs. anchors
+### Beginning a new task
 
-The HTML elements for buttons and links (a.k.a. anchors) describe a very specific type of action that is going to be taken when they are used. It is important you know when to use which, as the distinction matters:
+The user initiates a workflow for a resource that doesn't yet exist — for example, clicking "Create device" on a device management page or "Deploy application" in a pipeline view. This is the primary action on the page; use it once per view with a verb–noun label (e.g., "Create device", "Schedule backup"). Activating it navigates to the next step or opens the first stage of the workflow.
```
halocline, line 65: I'm not sure how well this text and the example compliment each other. 

Do we want to showcase button as the primary action on a page? If so, its probably more of an empty state scenario. Once we have records in the collection, "create" is probably no longer the primary. Or do want to show a more general "begin a new task" an use secondary instead?

halocline: Since the next use case example is "providin a primary call to action", then we should do the latter and use secondary in this example.

apps/docs/src/pages/components/button.mdx
```diff
@@ -38,189 +46,105 @@ import {
...
+  height={{ min: 'small' }}
+>
+  <ButtonBeginNewTaskExample />
+</Example>
 
-Icon buttons have an `a11yTitle`, `aria-label` or button `label` that describes the action that the user will be performing when using screen readers.
+### Providing a primary call to action
 
-**Best Practices:**
+A single primary button stands out visually among all other controls, signaling the most important or recommended action on the page. For example, "Activate service" on a dashboard draws the user's eye to the recommended next step. Use at most one primary button per view; all other actions use secondary or default styling.
```
halocline: Something other than dashboard.

```suggestion
A single primary button stands out visually among all other controls, signaling the most important or recommended action on the page. For example, "Activate service" on a page draws the user's eye to the recommended next step. Use at most one primary button per view; all other actions use secondary or default styling.
```

apps/docs/src/examples/components/button/ButtonCancelExample.js
```diff
@@ -0,0 +1,136 @@
...
+            <Button
+              label="Save changes"
+              primary
+              type="submit"
+            />
+            <Button
+              label="Cancel"
+              onClick={handleCancel}
+            />
+          </ButtonGroup>
```
halocline: Only use padding for internal spacing, not placement in a layout. 

@julauxen , @jcfilben , @britt6612 , @Sulaymon333 , @MikeKingdom ☝️ this should become a /knowledge/core/skills

```suggestion
<Box gap="medium">
  <>
          <FormField
            htmlFor="name"
            label="Device name"
            name="name"
            required
          >
            <TextInput
              id="name"
              name="name"
              onChange={e =>
                setDraft(prev => ({
                  ...prev,
                  name: e.target.value,
                }))
              }
              value={draft.name}
            />
          </FormField>
          <FormField
            htmlFor="description"
            label="Description"
            name="description"
          >
            <TextArea
              id="description"
              name="description"
              onChange={e =>
                setDraft(prev => ({
                  ...prev,
                  description: e.target.value,
                }))
              }
              value={draft.description}
            />
          </FormField>
 </>
          <ButtonGroup>
            <Button
              label="Save changes"
              primary
              type="submit"
            />
            <Button
              label="Cancel"
              onClick={handleCancel}
            />
          </ButtonGroup>
</Box>
```

halocline (commented):
I need to come back and do more reviewing...

apps/docs/src/examples/components/button/ButtonBeginNewTaskExample.js
julauxen: Yes, it should be created using data and friend imo. 
Also agree that we need to add a skill for code generation (when to use what, examples of best practices, etc). 
What we have right now for the generate-example-agents that kinda fits that model is shown below, but it's rather thin.

Image

I created a ticket on the backlog for us to map all improvements we wish to make on the workflow based on the first batch of restructured pages: https://github.com/grommet/hpe-design-system/issues/6229
I added this comment as a bullet point but will create a proper ticket in the future. Feel free to add other items to this ticket.

apps/docs/src/pages/components/button.mdx
```diff
@@ -38,189 +46,106 @@ import {
...
+  height={{ min: 'small' }}
+>
+  <ButtonCancelExample />
+</Example>
 
-### When to use a primary action button
+## Anatomy
 
-Primary action buttons are used within enterprise applications. They are meant to assist and are designed to maximize a user’s experience by highlighting the most common or important action on a page and allowing the user to accomplish a goal or task.
+{/* TODO: Add <Example plain> anatomy diagram once ButtonAnatomy component is created. */}
```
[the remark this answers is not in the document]
julauxen: this will be solved by #6217

apps/docs/src/examples/components/button/DosDonts/ButtonBadActionLabelsPreview.js
julauxen: fixed by commit ec49f389

apps/docs/src/examples/components/button/ButtonActionLabelsExample.js
julauxen: fixed by commit 2a9716f

apps/docs/src/examples/components/button/ButtonBeginNewTaskExample.js
julauxen: fixed by commit 5b6af2e

apps/docs/src/examples/components/button/ButtonCancelExample.js
julauxen: fixed by commit bc73b1a and knowledge gap added to #6229

apps/docs/src/pages/components/button.mdx
julauxen: I believe you are correct @halocline thanks for pointing it out and changing the example. I'll resolve this thread since this was altered already via commit c6955c8

apps/docs/src/examples/components/button/index.js
```diff
@@ -1,13 +1,20 @@
...
+export * from './ButtonCancelExample';
 export * from './ButtonIconExample';
 export * from './ButtonLeftAlignExample';
+export * from './ButtonPrimaryCtaExample';
 export * from './ButtonRightAlignExample';
+export * from './ButtonSecondaryActionsExample';
 export * from './ButtonSizingExample';
 export * from './ButtonStatesExample';
+export * from './ButtonToggleExample';
+export * from './ButtonWizardStepExample';
```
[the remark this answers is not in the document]
julauxen: fixed by commit 5d5a89492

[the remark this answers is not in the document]
julauxen: fixed by 5d5a89492

apps/docs/src/pages/components/button.mdx
```diff
@@ -279,128 +200,228 @@ The `badge` prop adds a small indicator to a Button to convey extra information,
...
+- **Disabled** — Prevents user interaction and applies reduced-opacity styling. Prefer leaving buttons active and surfacing an explanation on click over hiding or disabling them; do not show a disabled button for an action that is simply unavailable in the current context — remove it or show it active with an explanatory message.
+- **Busy** — Displays a spinner indicator to signal that an action is in progress. Use for short async operations (a few seconds or less); if the action may take longer, let the user leave and notify them separately when the task completes.
+- **Success** — Displays a check mark after an action has completed, providing immediate in-place confirmation before the UI transitions.
+
+<Example
+  docs="https://v2.grommet.io/button?theme=hpe#props"
+  code="https://raw.githubusercontent.com/grommet/hpe-design-system/master/apps/docs/src/examples/components/button/ButtonBusySimpleExample.js"
+  height={{ min: 'small' }}
+>
+  <ButtonBusySimpleExample />
```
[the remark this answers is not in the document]
julauxen: not needed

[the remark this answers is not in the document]
julauxen: not needed

apps/docs/src/examples/components/button/ButtonActionLabelsExample.js
```diff
@@ -0,0 +1,10 @@
+import React from 'react';
+import { Box, Button } from 'grommet';
+
+export const ButtonActionLabelsExample = () => (
+  <Box direction="row" gap="xsmall" wrap>
+    <Button label="Apply filters" onClick={() => {}} primary />
+    <Button label="Reset" onClick={() => {}} secondary />
```
britt6612, line 7: more of a question why do we make this button `secondary` with the Reset and the others in the DO Dont example were just default buttons?

apps/docs/src/examples/components/button/ButtonBeginNewTaskExample.js
britt6612: agree here it should definitely be implemented with data and friends!

apps/docs/src/examples/components/button/ButtonBusySimpleExample.js
```diff
@@ -1,26 +1,75 @@
 import React, { useState } from 'react';
-import { Button } from 'grommet';
+import { Box, Button, Form, FormField, TextInput } from 'grommet';
+import { ButtonGroup } from '@shared/aries-core';
 
 export const ButtonBusySimpleExample = () => {
   const [busy, setBusy] = useState(false);
   const [success, setSuccess] = useState(false);
 
+  const handleSubmit = () => {
```
britt6612, line 9: a little cleaner 


```
  const handleSubmit = () => {
    if (busy || success) return;
    setBusy(true);
    setTimeout(() => { setBusy(false); setSuccess(true); }, 2000);
    setTimeout(() => setSuccess(false), 5500);
  };
```

apps/docs/src/examples/components/button/ButtonSecondaryActionsExample.js
```diff
@@ -0,0 +1,41 @@
...
+const stats = [
+  { label: 'Capacity', value: '48 TiB' },
+  { label: 'Used', value: '31.2 TiB' },
+  { label: 'Status', value: 'Online' },
+];
+
+export const ButtonSecondaryActionsExample = () => (
+  <Box background="background-front" gap="large" pad="medium" round="small">
+    <Box gap="small">
+      <>
```
britt6612, line 14:
```suggestion

```


I dont think we need this fragment

apps/docs/src/pages/components/button.mdx
```diff
@@ -1,26 +1,34 @@
...
 import {
+  ButtonActionLabelsExample,
+  ButtonAlignmentTable,
+  ButtonBeginNewTaskExample,
+  ButtonBadActionLabelsPreview,
   ButtonBadCancelPreview,
   ButtonBadGroupPreview,
   ButtonBadSignUpPreview,
   ButtonBadgeExample,
   ButtonBusyExample,
```
britt6612: I dont think the ButtonBusyExample is being used

apps/docs/src/pages/components/button.mdx
```diff
@@ -38,189 +46,105 @@ import {
...
-### Toggle buttons
-
-Toggle buttons are used either to switch between two states, such as checked and unchecked, or to trigger two actions like on or off.
-
-#### Toggle buttons with icons
+<Example
+  docs="https://v2.grommet.io/button?theme=hpe#props"
+  height={{ min: 'small' }}
+>
+  <ButtonBeginNewTaskExample />
```
britt6612, line 71: these may need the code prop and the added links

apps/docs/src/examples/components/button/ButtonBeginNewTaskExample.js
Sulaymon333: Yes, I agreed. We can enhance the agents to take account for this case.

apps/docs/src/examples/components/button/ButtonBusySimpleExample.js
Sulaymon333: Yes read cleaner but in terms of outcome, I think both provide very similar outcomes but with subtle differences, so they are not exactly identical.

```js
// 1. first version (original)
 const handleSubmit = () => {
    if (busy || success) return;
    setBusy(true);
    setTimeout(() => {
      setBusy(false);
      setSuccess(true);
      setTimeout(() => {
        setSuccess(false);
      }, 3500);
    }, 2000);
  };

// 2. second version (suggested)
const handleSubmit = () => {
    if (busy || success) return;
    setBusy(true);
    setTimeout(() => { setBusy(false); setSuccess(true); }, 2000);
    setTimeout(() => setSuccess(false), 5500);
  };
```

**1. Timing behavior**

- First version: success is cleared 3500ms after the 2000ms busy delay finishes, so nominally 5500ms from submit.
- Second version: success clear is scheduled immediately for 5500ms from submit.
- In normal conditions they appear the same.
    - Under timer delays, they can differ slightly:
    - First version preserves a 3500ms success window after busy ends.
    - Second version keeps a fixed end time at 5500ms from click.

**2. Practical difference**

- First version is usually better if you care about “show success for 3.5s after request completes”.
- Second version is fine if you want a fixed total lifecycle from click.


So for For this use case, I would recommend we keep using the first pattern since it is closer to real-life implementation.

**Reason:**

1. It ties success timing to completion of the busy state.
2. Users consistently see success for the intended duration after the save finishes.
3. It is safer if the busy phase ever changes from a fixed 2000ms to a real async request.

So let's keep the behavior model below:

1. Start busy
2. Finish busy
3. Show success
4. Clear success after 3500ms

apps/docs/src/examples/components/button/ButtonBeginNewTaskExample.js
julauxen: @Sulaymon333 could you update this example to use data and friends in your clean up PR if you haven't done so already?

apps/docs/src/examples/components/button/ButtonActionLabelsExample.js
julauxen: the "reset" here could be both secondary or default. Because we are already showcasing cancel on default we can have reset as secondary to show a bit more of the hierarchy we offer, but neither are wrong - it all depends on the context for the user.

apps/docs/src/pages/components/button.mdx
julauxen: Removed by commit 5d5a89492

apps/docs/src/examples/components/button/ButtonSecondaryActionsExample.js
halocline: If the fragment is removed, won't a 'small' gap be applied between each Text? Is that what we want?

apps/docs/src/examples/components/button/ButtonBusySimpleExample.js
halocline: ☝️ I agree that we should stick with the current implementation as it resembles real-life implementations.

apps/docs/src/examples/components/button/ButtonBeginNewTaskExample.js
Sulaymon333: Yes, this is updated via this [commit](https://github.com/grommet/hpe-design-system/pull/6231/changes/0896561e1ea543f8aecc582b2a893ade2f70d661) on the clean up [PR](https://github.com/grommet/hpe-design-system/pull/6231)

Sulaymon333:
@britt6612 I have addressed most of the comments on this PR on the [clean-up PR](https://github.com/grommet/hpe-design-system/pull/6231). You can have a review when you plate is more free. This review request should not be a P1, from my understanding @julauxen.

PR #6212, opened 2026-05-20
[docs] Checkbox refactor

### LINKED ISSUE #6183: Checkbox - component restructure
Apply the restructure workflow to checkbox

julauxen:
[Deploy Preview](https://deploy-preview-6212--keen-mayer-a86c8b.netlify.app/)
checkbox: https://deploy-preview-6212--keen-mayer-a86c8b.netlify.app/components/checkbox

#### What does this PR do?
Generated checkbox yaml.
Restructures the checkbox documentation to fit our template. 

Closes #6183 

Anatomy and the coded examples generated by AI were already reviewed and approved.


apps/docs/src/pages/components/checkbox.mdx
```diff
@@ -67,37 +101,134 @@ Toggle is a great choice to use when the user is given a single option that they
...
+| Space       | Toggles the checked state of a focused checkbox. |
+| Tab         | Moves focus to the next focusable element.       |
+| Shift + Tab | Moves focus to the previous focusable element.   |
+
+### ARIA
+
+| Attribute         | Value                    | Condition                                                  |
+| ----------------- | ------------------------ | ---------------------------------------------------------- |
+| `aria-checked`    | `true \| false \| mixed` | Reflects checked, unchecked, or indeterminate state.       |
+| `aria-label`      | descriptive string       | When no visible `label` prop is provided.                  |
```
jcfilben: In the aria section can we change the guidance for using aria-label from: "When no visible label prop is provided." to "When no visible label prop is provided or when additional context is needed for a screenreader user."
```suggestion
| `aria-label`      | descriptive string       | When no visible `label` prop is provided or when additional context is needed for a screenreader user.                  |
```

apps/docs/src/pages/components/checkbox.mdx
```diff
@@ -67,37 +101,134 @@ Toggle is a great choice to use when the user is given a single option that they
...
+**Validation error** — appears via the enclosing `FormField` when the checkbox fails form validation.
 
 <Example
   docs="https://v2.grommet.io/checkbox?theme=hpe#props"
-  code="https://raw.githubusercontent.com/grommet/hpe-design-system/master/apps/docs/src/examples/components/checkbox/CheckBoxDisabledExample.js"
+  code="https://raw.githubusercontent.com/grommet/hpe-design-system/master/apps/docs/src/examples/components/checkbox/CheckBoxValidationExample.js"
   height={{ min: 'xsmall' }}
 >
-  <CheckBoxDisabledExample />
+  <CheckBoxValidationExample />
```
jcfilben, line 205: We advise against using a toggle in a FormField but then later on in the page show a toggle in a FormField. Seems like we should remove the toggle validation example.

Image

Image

apps/docs/src/pages/components/checkbox.mdx
britt6612: @julauxen is this still relevant

apps/docs/src/pages/components/checkbox.mdx
britt6612: @julauxen is this still relevant

apps/docs/src/pages/components/checkbox.mdx
julauxen: fixed via commit 0011ebfc1

apps/docs/src/pages/components/checkbox.mdx
julauxen: corrected via this commit 59cce5edb

apps/docs/src/pages/components/checkbox.mdx
```diff
@@ -1,63 +1,97 @@
...
 </Example>
 
-## Guidance
+## Use Cases
 
-CheckBox can be used to present an individual option or group of options to the user. When presented in a group, the user can select a single or multiple options. Refer to CheckBoxGroup for examples.
+### Accepting an agreement
 
-### Usage
+The user reads a statement — such as a privacy policy, terms of service, or disclaimer — and confirms acceptance by selecting a checkbox. Place the checkbox standalone, outside a `Form` with a submit action, in account setup steps or modal confirmations. Write the label as a first-person declaration, for example, "I agree to the terms and conditions". Selecting the checkbox enables a continue or submit button.
```
jcfilben: Should this say within a Form instead of outside a Form.

Also removed the backticks, usually we only use those for prop names or a snippet of inline code.

The last sentence says "Selecting the checkbox enables a continue or submit button" but usually we advise against disabling the button and instead advise for people to rely on form validation. So I feel like this last sentence goes against our guidance.
```suggestion
The user reads a statement — such as a privacy policy, terms of service, or disclaimer — and confirms acceptance by selecting a checkbox. Place the checkbox standalone, within a Form with a submit action, in account setup steps or modal confirmations. Write the label as a first-person declaration, for example, "I agree to the terms and conditions". Selecting the checkbox enables a continue or submit button.
```

apps/docs/src/pages/components/checkbox.mdx
```diff
@@ -1,63 +1,97 @@
...
+  <CheckBoxSolelyExample />
+</Example>
 
-There are three states which CheckBox provides:
+### Revealing content conditionally
 
-- Unchecked - Value has not been selected.
-- Checked - Value has been selected.
-- Indeterminate - Used in situations where the CheckBox is a parent to a list of child options. Use the indeterminate state to indicate that only a subset of the child options have been checked.
+Use a checkbox to show or hide additional fields or sections based on the user's choice. This keeps the form compact by surfacing detail only when relevant. The label should make the consequence of checking the box clear, for example, "Ship to a different address".
```
jcfilben, line 53: Would we recommend people use the 'toggle' prop here like is shown in the example? Or is it okay to use either a regular checkbox or a toggle in this case?

apps/docs/src/pages/components/checkbox.mdx
```diff
@@ -67,37 +101,134 @@ Toggle is a great choice to use when the user is given a single option that they
...
+</BestPracticeGroup>
+
+## Behaviors and States
+
+### Interactive states
+
+- **Rest** — default, unfocused appearance.
+- **Hover** — a visible hover indicator appears when the pointer is over the control or label.
+- **Focus** — a focus ring appears around the control when navigating via keyboard.
+- **Active** — the control appears pressed while holding the mouse button or Space key.
```
jcfilben: When I hold the mouse button down or hold the space key down I'm not sure that I see the 'pressed' appearance

apps/docs/src/pages/components/checkbox.mdx
```diff
@@ -67,37 +101,134 @@ Toggle is a great choice to use when the user is given a single option that they
...
-  code="https://raw.githubusercontent.com/grommet/hpe-design-system/master/apps/docs/src/examples/components/checkbox/CheckBoxValidationExample.js"
+  code="https://raw.githubusercontent.com/grommet/hpe-design-system/master/apps/docs/src/examples/components/checkbox/CheckBoxDisabledExample.js"
   height={{ min: 'xsmall' }}
 >
-  <CheckBoxValidationExample />
+  <CheckBoxDisabledExample />
 </Example>
 
-### Disabled
+**Indeterminate** — indicates that only a subset of child items in a list are selected. Use indeterminate only on a parent checkbox that controls a list of child checkboxes.
```
jcfilben: Maybe mention that indeterminate is not valid for toggles

apps/docs/src/pages/components/checkbox.mdx
julauxen: hi @jcfilben I think the form point is a different interpretation of what the description is trying to say. It's trying to convery that in this use cases, checkbox should be displayed as a standalone instead of insite a FormField wrapper. I'll update the sentence to make this clearer. 
Good point about the backstricks, this was something we had in instructions before but I checked and it's not there anymore 👀 I'll refine the sillks to contain this info.
You are correct about the last phrase. I'll remove it. I'll also update the example to not have a disabled button.

julauxen: Content updates via commit 59cce5edb

apps/docs/src/pages/components/checkbox.mdx
julauxen: I'd say both are ok. Do you feel like the example is misleading or need more info?

apps/docs/src/pages/components/checkbox.mdx
julauxen: that's a good point, the "active" state exists but it's not visual. Should it be visual? Or are you saying we should review the text to something like "Active — no distinct visual change; the control responds to mouse and keyboard press events."
There's also the possibility of making this section more generic, like in the menu PR: "Checkbox supports the following interactive states: rest, hover, focus, and active. These states are handled automatically by the component and follow the HPE Design System's interactive state styles."

I would go for the latter. Let me know what you think.

apps/docs/src/pages/components/checkbox.mdx
julauxen: Addressed in commit 38f190bbc

apps/docs/src/pages/components/checkbox.mdx
julauxen: Refined skills and agents for the backsticks on this PR: https://github.com/grommet/hpe-design-system/pull/6453

apps/docs/src/pages/components/checkbox.mdx
julauxen: Actually, we say in other parts of the guidance that this should not happen. I'll do some digging about what might be approapriate

julauxen: So I did some digging around other DS and this is a summary of what i found:

Toggles should generally not be used within forms because they communicate immediate action, whereas form controls typically collect input that is only applied after submission. Users expect a toggle to take effect instantly, similar to a physical on/off switch. When a toggle is placed inside a form that requires a submit, save, or apply action, it creates a mismatch between the control's behavior and the user's expectations.

For form-related choices such as "Use different address," "Remember my preference," or other contextual options, checkboxes provide a more appropriate mental model. Checkboxes represent selections that can be reviewed and submitted together, while toggles are better suited to persistent settings, modes, or features that change state immediately when activated.

In addition, checkboxes support common form patterns, scale better when multiple options are presented together, and avoid ambiguity about whether a change has already been applied or is still pending submission.

**So, I'll review the guidance to make sure that's reflected and that we don't have inconcistebcies on what we're saying.**

julauxen: fixed via commit 5e1bf7ff5

apps/docs/src/pages/components/checkbox.mdx
julauxen: Updated via commit 96efdad61

PR #6273, opened 2026-06-08
Docs - update forms example

julauxen:
[Deploy Preview](https://deploy-preview-6273--keen-mayer-a86c8b.netlify.app/)

#### What does this PR do?
Updates the **"Single required FormField"** example to match the rules presented under the [first Dos and Donts ](https://design-system.hpe.design/templates/double-confirmation?q=confirm#dos-and-do-nots) for double confirmation.
It states "Don't add unnecessary instructions or phrasing that causes doubt about what will happen or questions the user’s intent.", which is contradictory to the current live example under forms.

This was brought up by Marina, from GLP team.

BEFORE
Screenshot 2026-06-08 at 16 40 41
AFTER
Screenshot 2026-06-08 at 16 41 06

### DIFF
--- a/apps/docs/src/examples/templates/forms/SingleFormFieldExample.js
+++ b/apps/docs/src/examples/templates/forms/SingleFormFieldExample.js
@@ -45,8 +45,8 @@ export const SingleFormFieldExample = () => {
           </Heading>
         </Header>
         <Text>
-          Are you sure you want to delete this cluster? Doing so will erase all
-          the data.
+          This will permanently delete the cluster 
+          and all its associated data.
         </Text>
         <Box
           // Padding used to prevent focus from being cutoff
@@ -76,7 +76,7 @@ export const SingleFormFieldExample = () => {
               direction="row"
             >
               <Button label="Cancel" />
-              <Button label="Delete cluster" primary type="submit" />
+              <Button label="Delete" primary type="submit" />
             </Box>
           </Form>
         </Box>


Sulaymon333:
@julauxen thanks for the PR and providing a clearer subtitle text that follows our guidance. 

One quick observation about the primary button, it is renamed to "Delete" from "Delete cluster". Is this intentional?, can we double check it against the guidance on button [action-label conventions](https://design-system.hpe.design/components/button?q=button#action-label-conventions)

MikeKingdom:
> @julauxen thanks for the PR and providing a clearer subtitle text that follows our guidance.
> 
> One quick observation about the primary button, it is renamed to "Delete" from "Delete cluster". Is this intentional?, can we double check it against the guidance on button [action-label conventions](https://design-system.hpe.design/components/button?q=button#action-label-conventions)

IMHO I can go either way on this. Our guidance text on that is pretty wishy washy "When in doubt, follow the verb + noun convention to ensure clarity." Do I have doubt? :)

julauxen:
> @julauxen thanks for the PR and providing a clearer subtitle text that follows our guidance.
> 
> One quick observation about the primary button, it is renamed to "Delete" from "Delete cluster". Is this intentional?, can we double check it against the guidance on button [action-label conventions](https://design-system.hpe.design/components/button?q=button#action-label-conventions)

This is a good question. I deleted the noun to bring this example closer to the ones we have under "double confirmation". This labels guidance was part of the layers page previously and was moved to button, but neves discussed if still true.
I think let's leave this button without the noun, to get it closer to the double confirmation examples, and revaluate once we talk about the button content.

Sulaymon333 (approved):
LGTM

MikeKingdom (approved):

Issue #6121, opened 2026-04-22
All tokens page - improve table browsing experience (scroll or pagination)


julauxen:
## Background / Context
While adding the syntax switcher to the All Tokens page, a maximum width was introduced to the tokens table to expose horizontal scrolling. This resulted in the table becoming horizontally scrollable while the page itself already has vertical scrolling.

Currently:
- The page scrolls vertically
- The table scrolls both horizontally and vertically

This creates a confusing and awkward browsing experience, especially when interacting with large datasets.

---

## Problem Statement
The current table layout introduces nested scroll areas, which negatively impacts usability and makes it harder for users to browse, compare, and explore token data efficiently.

Users may:
- Lose context when scrolling
- Struggle to scan or compare values
- Experience friction when navigating large datasets

---

## Goal
Improve the browsing experience of the All Tokens table so users can comfortably and efficiently explore large datasets without confusion or unnecessary scrolling.

---

## Scope / Areas to Explore (Non-Prescriptive)
Potential approaches to investigate include, but are not limited to:

- Replacing vertical scrolling inside the table with pagination
- Introducing sticky elements (e.g. table headers or left navigation with a max height)
- Removing or reducing nested scroll behavior
- Exploring alternative table layouts or responsive strategies
- Improving keyboard and trackpad navigation

---

## Out of Scope
- Backend or API changes
- Token data accuracy or content updates

---

## Acceptance Criteria
- Users can browse large token lists using a single, clear scrolling behavior
- Table interactions feel intuitive and predictable
- Horizontal navigation (if required) does not interfere with vertical page scrolling

PR #6286, opened 2026-06-11
[Docs] Made the description column smaller and swapped description and value…

### LINKED ISSUE #6200: All tokens table scrolls in 2 directions and difficult to use
The "All design tokens" page/table is difficult to use since it always scrolls in 2 directions, even when not necessary.

There's no way to get the area the table is in to be wider. If you're trying to see tokens and what their value is you have to keep scrolling the page to the bottom to access the scrollbar, scroll horizontally, scroll back up to find the tokens you wanted to see (which may be difficult since the token may now be scrolled off to the left, etc.

Image

Image

There's no reason the description field needs to be there. 

Recommendation: Besides allowing the table more horizontal space, the column widths shouldn't be fixed and the description column shouldn't dominate. It would be more useful to swap the position of the value and description columns

MikeKingdom:
[Deploy Preview](https://deploy-preview-6286--keen-mayer-a86c8b.netlify.app/design-tokens/all-design-tokens?token=semantic%2Fcolor&tokenTypes=docs&mode=light)

#### What does this PR do?
Currently the "All tokens" page has a table that often scrolls in 2 directions. When working with some tokens it's very useful to see the value (right-most column). However the description column comes before it and has a fixed size of `large`. It almost always forces the value column to not be visible and making the window larger is no help (the page content is not full so it doesn't get wider). In many cases there is no description or a short description so you end up with a bunch of empty whitespace and still have to scroll horizontally to see the content. This makes it very difficult to use in practice. As a general rule, we try to avoid scrolling in 2 directions.

This change moves the "Output value" column left of the "Description", The value column is typically pretty short and is often what I'd rather see since the token name itself is often descriptive enough (or the only description.) I also changed the description to render as a max width rather than always being "large". See the example below of how this helps.

Before:
image

After:
image

Simpler set of tokens that have descriptions (after change):
image

Fixes #6200 

DesignTokensTable.js

All design tokens page

Not necessary

Backwards compatible

### DIFF
--- a/apps/docs/src/components/content/design-tokens/DesignTokensTable.js
+++ b/apps/docs/src/components/content/design-tokens/DesignTokensTable.js
@@ -89,8 +89,11 @@ const getTokenColumn = (property, header) => ({
 const descriptionColumn = {
   property: 'description',
   header: 'Description',
-  size: 'large',
-  render: datum => <Text>{datum.description || '--'}</Text>,
+  render: datum => (
+    <Box width={{ max: 'medium' }}>
+      <Text>{datum.description || '--'}</Text>
+    </Box>
+  ),
 };
 
 const formatTokenValue = value => {
@@ -153,8 +156,8 @@ export const DesignTokensTable = ({
     return [
       previewColumn,
       ...dynamicTokenColumns,
-      descriptionColumn,
       valueColumn,
+      descriptionColumn,
     ];
   }, [tokenTypeColumns]);
 


Sulaymon333:
@MikeKingdom thank you for surfacing this issue and working on possible solution. 

I raised exact concern on the original [PR](https://github.com/grommet/hpe-design-system/pull/6105) via this [thread](https://github.com/grommet/hpe-design-system/pull/6105#issuecomment-4295575583) and @julauxen created a follow-on ticket to have a close look. It looks like we might be having a duplicate ticket here #6121. Once it is confirmed as duplicate, I can close #6121 that out as a duplicate of #6200.

From dev workflow, it is essential to see the resolved value for a token, and I find myself to constantly scroll left and right as well.

With the current fix, it is quite helpful and make the experience way better. However, we quickly get to the same issue of "value column" out of view when multiple token modes (css vars, figma, grommet, etc) is enabled. Will pinning a column like "value column" in this case be a possible solution?

MikeKingdom:
> With the current fix, it is quite helpful and make the experience way better. However, we quickly get to the same issue of "value column" out of view when multiple token modes (css vars, figma, grommet, etc) is enabled. Will pinning a column like "value column" in this case be a possible solution?

Thanks @Sulaymon333 . Pinning columns is pretty heavy handed and takes control away from the user. Given that turning on that many token formats is not a common enough use case, I'd rather not try to optimize for it somehow. I have considered changing this page to a `full` format rather than just `wide` so if I really have the monitor space, I could use it more effectively here.

Sulaymon333:
@MikeKingdom agreed considering this only become an issue when one enables more than two modes. So, I am happy we have your current implementation.

@julauxen can we close #6121 as a duplicate or does the ticket cover other areas not touched in this PR?

Sulaymon333 (approved):
LGTM

britt6612 (approved):

julauxen:
> @MikeKingdom agreed considering this only become an issue when one enables more than two modes. So, I am happy we have your current implementation.
> 
> @julauxen can we close #6121 as a duplicate or does the ticket cover other areas not touched in this PR?

Let's not close it, as Mike's PR is more connected to the description being moved and #6121 is broader, to think about how we can improve the browsing experience as a whole. Even though #6200 talks about all the scroll areas, this PR doesn't seem to fix that issue, but rather focused on shifting things within the table to improve use.

PR #6390, opened 2026-07-16
Updated icon details

SeamusLeonardHPE:
https://deploy-preview-6390--keen-mayer-a86c8b.netlify.app/templates/status-indicator#icons-and-shapes

**Change 1**
Replaced  `Info` with  `StatusInfo` icon as the copy and other examples are all status related.

**Change 2**
Modified copy from "normal" to "ok" to match icon semantic naming convention

Previous example uses the [Info](https://hpe-design-icons-grommet.netlify.app/?path=/story/icons-hpe-icons-for-grommet--default&args=iconName:Info;size:xxlarge) icon incorrectly

image

Updated to use the correct semantic icon [StatusInfo](https://hpe-design-icons-grommet.netlify.app/?path=/story/icons-hpe-icons-for-grommet--default&args=iconName:StatusInfo;size:xxlarge)

image

### DIFF
--- a/apps/docs/src/examples/cardPreviews/status-indicator.js
+++ b/apps/docs/src/examples/cardPreviews/status-indicator.js
@@ -6,7 +6,7 @@ import {
   StatusWarning,
   StatusCritical,
   StatusUnknown,
-  Info,
+  StatusInfo,
 } from '@hpe-design/icons-grommet';
 
 import { Box } from 'grommet';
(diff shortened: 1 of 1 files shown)


julauxen (approved):
These look correct

Sulaymon333 (approved):
LGTM

jcfilben:
@SeamusLeonardHPE 
We were using the old info icon in our examples for accessibility reasons. The StatusInfo icon isn't distinguishable from the icon used for the Unknown state.

The longer term plan was to explore options for StatusInfo and make it accessible and then switch to recommending using StatusInfo instead of Info. But the temporary solution when we upgraded the site to use the landmark theme was to just use the Info icon.

PR #6416, opened 2026-07-30
[aries-core] Zerto - Provide recommendation and/or reference regarding customizing Select options

### LINKED ISSUE #6414: [Support] Zerto - Provide recommendation and/or reference regarding customizing Select options
Slack thread: https://hpe-internal.slack.com/archives/C0BG1AUSX4H/p1785330471028979

Since we get this type of question quite often, as alternative to CodeSandbox, perhaps creating a story for aries-core Storybook would be useful and something we could reuse and point people to instead?

britt6612:
[Deploy Preview](https://deploy-preview-6416--hpe-design-storybook.netlify.app/?path=/story/patterns-forms-input-customization--loading)

#### What does this PR do?
Adds a new **Loading, icons & grouping** pattern story to the Select component that demonstrates how to implement:
- **Loading state** — shows a spinner in the placeholder while options are being fetched
- **Grouped options** — organizes options by category (e.g., Production, Development) with visual group headings and dividers
- **Custom rendering** — displays option-level icons and a left-side selection marker that matches HPE design tokens

Closes #6414
Select story
1. Open Storybook: `pnpm storybook`
2. Navigate to `Patterns > Select > Loading, icons & grouping`
3. Test the following:
   - Loading spinner appears in placeholder for 1.5s, then options render
   - Grouped options display with category headings (Production, Development) and dividers between groups
   - Clicking an option shows the selected value with icon indicators
   - Keyboard and mouse navigation work (focus should land on the first selectable option)
   - Selected options show the left-side marker (aligned with HPE theme styling)
This story demonstrates that the Select component is flexible enough to support complex patterns without requiring wrapper components. It uses:
- **`valueLabel` prop** — for custom rendering of the selected value
- **`children` prop** — for custom option rendering (group headings + option rows)
- **`disabledKey` prop** — to disable group label rows from selection
- **`onOpen` callback** — to handle focus management
Yes. This adds a new pattern story that addresses a common request from teams adopting the HPE Design System. It should be noted in the changelog as a new pattern example for the Select component.
yes


shared/aries-core/src/stories/select.stories.js
```diff
@@ -0,0 +1,274 @@
...
+    </StyledMarkerBox>
+  );
+};
+
+const meta = {
+  title: 'Patterns/Select',
+  parameters: {
+    controls: { disable: true },
+  },
+};
```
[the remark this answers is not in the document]
britt6612: oh yes Ill need to do this ☝️

shared/aries-core/src/stories/select.stories.js
```diff
@@ -0,0 +1,274 @@
...
+      const isFirstInGroup =
+        index === 0 || options[index - 1].group !== option.group;
+
+      if (isFirstInGroup) {
+        acc.push({
+          label: option.group,
+          value: `__group__${option.group}`,
+          isGroupLabel: true,
+          disabled: true,
+        });
```
halocline: Nice.

halocline:
Screenshot 2026-07-31 at 3 44 06 PM

@britt6612 
- keyboard and focus behaviors are looking good.
- option fontWeight does not match a typical Select option 
- selected option background / marker does not match typical selected option (is there an issue with the theme selector?)
- divider styling matching NavigationMenu is good (given the anatomy, its easier to implement than the full width Menu divider)
- Padding is decreasing below a responsive breakpoint. Probably should stay consistent, so `<Box responsive={false} ... />`.
- Group heading is indented further than option label. Probably should have the same justification. May be impacted by the Box responsive mentioned above.
- Separate into distinct stories demonstrating each capability ( 1. loading, 2. custom options, 3. grouped options). Not all in one.


Consider structuring so that the `renderOption` / `OptionRow` is something that can be reusable from `aries-core/source/js/components/core`. Maybe called `<SelectOption />`? Same with `<LoadingPlaceholder />` which could be reused by other input types if needed. Then import those reusable components into the stories similar to NavigationMenu.

shared/aries-core/src/stories/select/LoadingStory.js
jcfilben, line 1: If the options are still loading I would expect a spinner/skeleton within the drop instead of at the placeholder level

jcfilben (commented):
Looks like the code isn't loading for these stories

Image

shared/aries-core/src/stories/select/LoadingStory.js
britt6612: think the teams were wanting the loading option in the select see here for more conversation

https://hpe-internal.slack.com/archives/C0BG1AUSX4H/p1785330471028979
https://hpe.slack.com/archives/C04LMJWQT/p1785519876459069

jcfilben: I think that's fine if they want to do that but I would hesitate to put this pattern on our DS storybook, I don't know that we would endorse this pattern and want other teams to copy it

britt6612: I think from my understanding this is what we were wanting and having it in out DS storybook was the recommended I think also Victor was going to add it to Grommet as an enhancement so someone didnt have to do a custom loading 

cc: @halocline @oliverHPE

halocline: The consensus splits cleanly across two main use cases:

## 1. The Inline Spinner Wins for Dependent Form Fields
If the select dropdown is loading data automatically because of a previous choice the user made (e.g., selecting a "State" field causing the "City" dropdown to fetch data), the inline spinner is the industry standard. [3] 

* The Logic: The user did not click the second dropdown yet, so opening a menu with a loading message would be intrusive and unexpected.
* The Consensus: Keep the select field closed, display a small [[inline spinner](https://uxplanet.org/loading-states-ux-patterns-9e8d2d45fa3c)](https://uxplanet.org/loading-states-ux-patterns-9e8d2d45fa3c) inside it, and temporarily disable interaction until the data populates. [3, 4] 

## 2. The Loading List Wins for Type-Ahead & Searchable Dropdowns
If the user actively clicks into a combobox or type-ahead field and starts typing to filter a remote database (e.g., searching for a flight destination or tagging a user), the loading list item or a component skeleton wins. [5] 
 
* The Logic: The user has already initiated an interaction with that specific field. They expect the container to respond immediately.
* The Consensus: Open the menu overlay instantly to confirm the application acknowledged their keystroke. Displaying a "Loading..." message or a pulsing skeleton row inside the menu keeps the user's focus exactly where the results will appear. [5, 6] 

## Summary Decision Matrix

| Scenario | Consensual Pattern | Why? |
|---|---|---|
| Dependent fields (loading on page load) | Inline Spinner | Prevents empty dropdown clicks. |
| Async Search (user actively typing) | Loading List Item | Confirms keystroke input instantly. |
| Heavy Data (takes >3 seconds) | Skeleton Rows in List | Lowers perceived wait times. |

[1] [[https://m3.material.io](https://m3.material.io/components/loading-indicator/guidelines)](https://m3.material.io/components/loading-indicator/guidelines)
[2] [[https://www.nngroup.com](https://www.nngroup.com/videos/skeleton-screens-vs-progress-bars-vs-spinners/)](https://www.nngroup.com/videos/skeleton-screens-vs-progress-bars-vs-spinners/)
[3] [[https://www.linkedin.com](https://www.linkedin.com/posts/katerina-anna-smith_which-loading-option-would-you-choose-activity-7488644672219676672-gMyK)](https://www.linkedin.com/posts/katerina-anna-smith_which-loading-option-would-you-choose-activity-7488644672219676672-gMyK)
[4] [[https://ux.stackexchange.com](https://ux.stackexchange.com/questions/119184/what-are-the-best-ux-practices-to-show-that-an-autocomplete-data-for-an-input-fi)](https://ux.stackexchange.com/questions/119184/what-are-the-best-ux-practices-to-show-that-an-autocomplete-data-for-an-input-fi)
[5] [[https://www.onething.design](https://www.onething.design/post/skeleton-screens-vs-loading-spinners)](https://www.onething.design/post/skeleton-screens-vs-loading-spinners)
[6] [[https://www.linkedin.com](https://www.linkedin.com/posts/zubaeruxd_which-loading-ui-actually-feels-better-for-activity-7402687813093433344-CCnS)](https://www.linkedin.com/posts/zubaeruxd_which-loading-ui-actually-feels-better-for-activity-7402687813093433344-CCnS)

halocline: Major design systems handle this distinction deliberately. Out-of-the-box support heavily leans toward the Inline Spinner pattern because it maps easily to a simple Boolean attribute (like loading or isLoading). However, frameworks built for advanced, data-heavy applications prioritize the Loading List pattern.

------------------------------

## Systems Native to the "Inline Spinner" Pattern

These systems feature a built-in prop that instantly handles the inline spinner. When activated, it swaps or complements the dropdown chevron with an animated wheel and temporarily disables user interaction.
 
* [Ant Design (AntD)](https://ant.design/components/select/): Features native support via the Select `loading` prop. It injects a small, revolving spinner directly onto the right side of the closed input container, replacing the standard arrow icon. [1] 
* [Chakra UI](https://chakra-ui.com/docs/components/select): Relies heavily on the [[Chakra React Select](https://www.npmjs.com/package/chakra-react-select)](https://www.npmjs.com/package/chakra-react-select) wrapper for asynchronous operations. Passing isLoading to this element natively renders a loading spinner right inside the input indicator container. [2] 
* [IBM Carbon Design System](https://carbondesignsystem.com/): Uses a variant of this for its basic dropdown components. While it leans heavily on skeleton patterns for whole-page loads, a component-level loader wraps the specific field to keep layout shifts locked down. [3, 4, 5]  

------------------------------

## Systems Native to the "Loading List" Pattern

These libraries are optimized for highly dynamic type-aheads, comboboxes, and multi-select tags. They prioritize immediate overlay responses to user typing.

* React Select / Headless UI: Considered the benchmark for type-ahead mechanics. When fetching data remotely, these libraries automatically expand the menu overlay and swap the choices container with an option that reads "Loading...". [6] 
* [[Preline UI / FlyonUI](https://flyonui.com/docs/advanced-forms/advanced-select/)](https://flyonui.com/docs/advanced-forms/advanced-select/): Utilizes unstyled headless Tailwind plugins to support an "Advanced Select" component. It defaults to opening the dropdown container upon focus, displaying an animated placeholder row while the data queries. [7] 
* Microsoft Fluent UI: Prefers using an expanded loading list item for searchable comboboxes, pairing a small loading track or spinner with technical microcopy within the list row to reassure users typing fast on enterprise applications.

------------------------------

## Summary Checklist for Implementation
 
* Choose Inline Spinner classes if using a static custom dropdown component where input is restricted until options are fully resolved.
* Choose a Loading List Item layout if building an interactive search feature where keystrokes trigger instant asynchronous fetches. [6, 7, 8] 


[1] [[https://ant.design](https://ant.design/components/select/)](https://ant.design/components/select/)
[2] [[https://www.npmjs.com](https://www.npmjs.com/package/chakra-react-select)](https://www.npmjs.com/package/chakra-react-select)
[3] [[https://carbondesignsystem.com](https://carbondesignsystem.com/patterns/loading-pattern/)](https://carbondesignsystem.com/patterns/loading-pattern/)
[4] https://carbondesignsystem.com/
[5] [[https://carbondesignsystem.com](https://carbondesignsystem.com/components/loading/usage/)](https://carbondesignsystem.com/components/loading/usage/)
[6] [[https://www.youtube.com](https://www.youtube.com/watch?v=qJnIJa-cF2M)](https://www.youtube.com/watch?v=qJnIJa-cF2M)
[7] [[https://flyonui.com](https://flyonui.com/docs/advanced-forms/advanced-select/)](https://flyonui.com/docs/advanced-forms/advanced-select/)
[8] [[https://www.youtube.com](https://www.youtube.com/watch?v=Z0IaN0CHLm4&t=500)](https://www.youtube.com/watch?v=Z0IaN0CHLm4&t=500)
[9] [[https://tailwindcss.com](https://tailwindcss.com/plus/ui-blocks/application-ui/forms/select-menus)](https://tailwindcss.com/plus/ui-blocks/application-ui/forms/select-menus)
[10] [[https://refine.dev](https://refine.dev/blog/ant-design-vs-mui/)](https://refine.dev/blog/ant-design-vs-mui/)

halocline: Does the when to use which logic resonate?

**Primary use case:** dynamic form data using a progressive disclosure pattern. (e.g., select environment B --> subsequent inputs rely on environment B specific data --> need to show user next steps in the form --> need to fetch related options --> need to communicate to the user that data is being fetched and that fetching was triggered by the user's choice --> where does that communication happen?

What are the concerns about the inline pattern?

Worth a share out discussion with broader team?

halocline:
> * Padding is decreasing below a responsive breakpoint. Probably should stay consistent, so `<Box responsive={false} ... />`.

Screenshot 2026-08-06 at 2 27 07 PM

The padding is still shrinking

halocline:
Screenshot 2026-08-06 at 2 29 57 PM

Marker height and selected background height do not match

halocline:
input height and width when no selection:
Screenshot 2026-08-06 at 2 31 20 PM


input height and width after selection:
Screenshot 2026-08-06 at 2 31 32 PM

I would expect the input's dimensions to remain the same.

halocline:
Screenshot 2026-08-06 at 2 36 23 PM

Screenshot 2026-08-06 at 2 36 13 PM


- I would expect the spinner's left alignment to be the same alignment as the placeholder text alignment.
- Gap between spinner and "loading" should be tighter. Maybe down one t-shirt size?

halocline:
Screenshot 2026-08-06 at 2 41 51 PM

Icon on the right should be right-aligned.

```
const Identifier = {
  <Box direction="row" gap="3xsmall">
    <Icon />
    <Text />
  </Box>
}

<Box direction="row" justify="between">
  <Identifier />
  <StatusIcon />
</Box>
```

britt6612:
> > * Padding is decreasing below a responsive breakpoint. Probably should stay consistent, so `<Box responsive={false} ... />`.
> 
> Screenshot 2026-08-06 at 2 27 07 PM
> The padding is still shrinking

sorry about that not sure where that change went but I push up again

halocline:
GroupedOptions story looks good.

halocline:
Story names and placement.

I think we should group these under "Forms" as they are patterns to solve form related use cases / user problems. Or create a "Customizing inputs" / "Input customization" directory. Kind of like the idea of creating a new directory.

Pattern directories and stories should be organized and labeled by the problem being solved.

- Loading --> Loading options
- CustomOptions --> Custom selected value
- GroupedOption --> Grouped options

Sulaymon333:
> ## Pull request overview
> Copilot reviewed 8 out of 8 changed files in this pull request and generated no new comments.
> 
> Suppressed comments (5)
> **shared/aries-core/src/stories/select.stories.js:10**
> 
> * The PR description/manual test steps say this should appear under “Patterns > Select …”, but the Storybook title nests it under “Patterns > Forms > Input customization”. This makes the documented navigation path inaccurate and reduces discoverability for Select-specific guidance.
> 
> ```
> const meta = {
>   title: 'Patterns/Forms/Input customization',
>   parameters: {
>     controls: { disable: true },
>   },
> ```
> 
> **shared/aries-core/src/stories/select/GroupedOptionsStory.js:44**
> 
> * This story claims (in the PR description) to manage focus on open so mouse-open lands on the first selectable option, but there is no `onOpen` handler here. With disabled group label rows, focus can land on the listbox container instead of an option.
> 
> ```
>           labelKey={renderOptionLabel}
>           valueKey={{ key: 'value', reduce: true }}
>           disabledKey="disabled"
>           placeholder="Select a service"
>           onChange={({ value: nextValue }) => setValue(nextValue)}
> ```
> 
> **shared/aries-core/src/stories/select/CustomOptionsStory.js:23**
> 
> * The options include per-option `icon` + `iconColor` (from `allOptions`), but the option renderer always shows the same `<Cpu />` icon. This prevents the story from actually demonstrating option-level icon customization.
> 
> ```
>   const renderOptionLabel = option => {
>     return (
>       <Box direction="row" align="center" gap="small">
>         <Cpu size="small" />
>         <Text>{option.label}</Text>
> ```
> 
> **shared/aries-core/src/stories/select/LoadingStory.js:16**
> 
> * `fetchOptions()` resolves asynchronously, but the effect doesn’t guard against the component unmounting before the promise resolves. Switching stories quickly can cause a React “setState on unmounted component” warning.
> 
> ```
>   useEffect(() => {
>     fetchOptions().then(data => {
>       setOptions(data.map(({ label }) => label));
>       setLoading(false);
>     });
> ```
> 
> **shared/aries-core/src/js/components/core/SelectOption/focusFirstSelectableOption.js:32**
> 
> * `requestAnimationFrame` is called unconditionally. If this utility is used in a non-browser environment (or a test environment without `requestAnimationFrame`), it will throw.
> 
> ```
>   if (attempt < maxAttempts) {
>     requestAnimationFrame(() =>
>       focusFirstSelectableOption({
>         root: searchRoot,
>         listboxSelector,
> ```

@britt6612 can you have a look at the copilot comments?

Sulaymon333:
Should we tighten up the group header row to be closer to the items under it, the current group headers look like they are (smaller) select options as well.

image

Sulaymon333:
@britt6612 I noticed the group headers are implemented as disabled Select options via `disabledKey`. In the rendered accessibility tree, they remain `role="option"` with `aria-disabled="true"` rather than being exposed as named option groups.

This raises a few accessibility concerns:

1. Arrow-key navigation skips the disabled header rows, although this is a already pattern for disabled options, applying to this use case means screen-reader users may miss the group context.  
2. There is no programmatic relationship between each visual heading and the options below it.
3. The group headings are included in the listbox’s **option count**.
4. Because the grouping is conveyed visually but not programmatically, this presents a WCAG 1.3.1 Info and Relationships risk.

Should we document this as a visual interim workaround until Grommet Select supports native grouped-option semantics?

image
image

shared/aries-core/src/stories/select/shared.js
```diff
@@ -0,0 +1,64 @@
...
+        label: option.group,
+        value: `__group__${option.group}`,
+        isGroupLabel: true,
+        disabled: true,
+      });
+    }
+
+    acc.push(option);
+    return acc;
+  }, []);
```
Sulaymon333: `buildGroupedOptions` assumes options from the same group are adjacent, causing duplicate headings for unsorted data.

Image

```suggestion
const buildGroupedOptions = (options) => {
  const groups = new Map();

  options.forEach((option) => {
    const groupOptions = groups.get(option.group) || [];
    groupOptions.push(option);
    groups.set(option.group, groupOptions);
  });

  const result = [];

  groups.forEach((groupOptions, group) => {
    result.push(
      {
        label: group,
        value: `__group__${group}`,
        isGroupLabel: true,
        disabled: true,
      },
      ...groupOptions,
    );
  });

  return result;
};
```

britt6612:
> > ## Pull request overview
> > Copilot reviewed 8 out of 8 changed files in this pull request and generated no new comments.
> > Suppressed comments (5)
> > **shared/aries-core/src/stories/select.stories.js:10**
> > 
> > * The PR description/manual test steps say this should appear under “Patterns > Select …”, but the Storybook title nests it under “Patterns > Forms > Input customization”. This makes the documented navigation path inaccurate and reduces discoverability for Select-specific guidance.
> > 
> > ```
> > const meta = {
> >   title: 'Patterns/Forms/Input customization',
> >   parameters: {
> >     controls: { disable: true },
> >   },
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > **shared/aries-core/src/stories/select/GroupedOptionsStory.js:44**
> > 
> > * This story claims (in the PR description) to manage focus on open so mouse-open lands on the first selectable option, but there is no `onOpen` handler here. With disabled group label rows, focus can land on the listbox container instead of an option.
> > 
> > ```
> >           labelKey={renderOptionLabel}
> >           valueKey={{ key: 'value', reduce: true }}
> >           disabledKey="disabled"
> >           placeholder="Select a service"
> >           onChange={({ value: nextValue }) => setValue(nextValue)}
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > **shared/aries-core/src/stories/select/CustomOptionsStory.js:23**
> > 
> > * The options include per-option `icon` + `iconColor` (from `allOptions`), but the option renderer always shows the same `<Cpu />` icon. This prevents the story from actually demonstrating option-level icon customization.
> > 
> > ```
> >   const renderOptionLabel = option => {
> >     return (
> >       <Box direction="row" align="center" gap="small">
> >         <Cpu size="small" />
> >         <Text>{option.label}</Text>
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > **shared/aries-core/src/stories/select/LoadingStory.js:16**
> > 
> > * `fetchOptions()` resolves asynchronously, but the effect doesn’t guard against the component unmounting before the promise resolves. Switching stories quickly can cause a React “setState on unmounted component” warning.
> > 
> > ```
> >   useEffect(() => {
> >     fetchOptions().then(data => {
> >       setOptions(data.map(({ label }) => label));
> >       setLoading(false);
> >     });
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > **shared/aries-core/src/js/components/core/SelectOption/focusFirstSelectableOption.js:32**
> > 
> > * `requestAnimationFrame` is called unconditionally. If this utility is used in a non-browser environment (or a test environment without `requestAnimationFrame`), it will throw.
> > 
> > ```
> >   if (attempt < maxAttempts) {
> >     requestAnimationFrame(() =>
> >       focusFirstSelectableOption({
> >         root: searchRoot,
> >         listboxSelector,
> > ```
> 
> @britt6612 can you have a look at the copilot comments?

@Sulaymon333 
Honest assessment of each:

Title path mismatch: Not worth acting on. The Patterns/Forms/Input customization path is exactly what Matt requested in a real review comment. Copilot is comparing against an outdated PR description.

Missing onOpen handler: Fixed this.

Custom options always shows `<Cpu />` the point is just to show a custom renderer layout, it's fine as-is, unless anyone thinks otherwise

LoadingStory unmount guard: Technically correct but low priority. The "setState on unmounted component" warning was removed in React 18 and doesn't cause actual bugs. Storybook stories also rarely get unmounted mid-fetch in practice.

requestAnimationFrame guard: Noise for this context. It's a browser-only story utility; it'll never run in a non-browser environment.

britt6612:
> @britt6612 I noticed the group headers are implemented as disabled Select options via `disabledKey`. In the rendered accessibility tree, they remain `role="option"` with `aria-disabled="true"` rather than being exposed as named option groups.
> 
> This raises a few accessibility concerns:
> 
> 1. Arrow-key navigation skips the disabled header rows, although this is a already pattern for disabled options, applying to this use case means screen-reader users may miss the group context.
> 2. There is no programmatic relationship between each visual heading and the options below it.
> 3. The group headings are included in the listbox’s **option count**.
> 4. Because the grouping is conveyed visually but not programmatically, this presents a WCAG 1.3.1 Info and Relationships risk.
> 
> Should we document this as a visual interim workaround until Grommet Select supports native grouped-option semantics?
> 
> image image

@Sulaymon333 
Thanks for the thorough breakdown these are valid concerns and we've addressed them as far as the current Grommet Select API allows:

Arrow-key navigation skipping headers: I added `onOpen` to focus the first selectable option automatically, so keyboard users land on a real option rather than the listbox container. The skip behavior on disabled rows is consistent with how Grommet handles all disabled options.

No programmatic relationship:  I added `role="presentation"` to the header wrapper and `role="heading" aria-level={3})` to the label text, matching the same pattern used in the side navigation's grouped headings. Screen readers that process inner listbox content will now announce these as section headings. A full fix would require the listbox structure to look like this:

This isn't achievable with the current Grommet Select API Grommet owns the listbox and always renders a flat list of `role="option"` buttons. There's no hook to inject `role="group"` wrappers around a subset of options. We'd need a Grommet change for full native group semantics.

Headers inflating option count: This is also a constraint of the flat listbox structure. No workaround is available without that upstream change.

WCAG 1.3.1 risk:  Agreed. With `role="heading"`the grouping is partially programmatic, but the binding between heading and child options is still missing, so this would still be flagged in a strict audit. We'll document this as a known limitation and track the upstream Grommet enhancement needed to resolve it fully.

cc: @jcfilben

shared/aries-core/src/stories/select/shared.js
Sulaymon333: @britt6612 did you notice this comment yet?

Sulaymon333 (approved):
LGTM, thank you for the PR.

shared/aries-core/src/stories/select/shared.js
britt6612: yes it was implemented you can see the file is outdated

shared/aries-core/src/stories/select/CustomOptionsStory.js
```diff
@@ -0,0 +1,93 @@
...
+
+const CustomOptionsExample = () => {
+  const [value, setValue] = useState('');
+  const selected = allOptions.find(o => o.value === value) || null;
+  const theme = useContext(ThemeContext);
+  const optionPad = theme?.button?.size?.medium?.option?.pad || HPE_OPTION_PAD;
+
+  const renderOptionLabel = option => {
+    return (
+      <Box direction="row" align="center" gap="small">
```
halocline:
```suggestion
      <Box responsive={false} direction="row" align="center" gap="small">
```

shared/aries-core/src/js/components/core/Select/utils/focusFirstSelectableOption.js
```diff
@@ -0,0 +1,35 @@
+// Grommet auto-focuses an option on keyboard-open, but mouse-open can leave
+// focus on the listbox container, so we force focus to the first selectable option.
+const focusFirstSelectableOption = ({
```
halocline, line 5: Is this a component or a utility? Seems like a utility, but where is the component SelectOption?

I was expecting to see some components here (e.g. group options, custom option, etc.)

I fine with this for now. We can reorganize at a later date.

halocline (commented):
small fix and a question.

shared/aries-core/src/js/components/core/Select/utils/focusFirstSelectableOption.js
britt6612: > Is this a component or a utility? Seems like a utility, but where is the component SelectOption?
> 
> I was expecting to see some components here (e.g. group options, custom option, etc.)
> 
> I fine with this for now. We can reorganize at a later date.

Good call, I've since reorganized this as part of this PR. @halocline 

`focusFirstSelectableOption` was moved out of the old SelectOption/ folder into a unified Select feature folder with a clearer structure:

```
Select/
  components/
    SelectGroupLabel.js       ← accessible group label (role, aria-level, pad)
    SelectOptionLabel.js      ← custom option row with icon + label
    SelectValueLabel.js       ← custom selected value with icon + label
    SelectOptionTheme.js      ← theme patch for transparent disabled options
    index.js
  utils/
    focusFirstSelectableOption.js
    index.js
  index.js

```

The `Select/components/`folder is now where the reusable option-level components live, and [Select/utils/] holds behavioral helpers. The old SelectOption/ and SelectUtil/ folders have been removed entirely.

The stories `GroupedOptionsStory, CustomOptionsStory` now import from `../../js/components/core/Select` and each has a dedicated example file `GroupedOptionsExample.js`, `CustomOptionsExample.js` that Storybook Docs uses as its source so the code panel shows only the consumer-facing component code, not Storybook boilerplate.

shared/aries-core/src/js/components/core/Select/components/SelectOptionLabel.js
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
+import { Box, Text } from 'grommet';
+
+const SelectOptionLabel = ({ icon: Icon, label, iconColor = 'icon' }) => (
+  <Box direction="row" align="center" gap="small">
```
halocline:
```suggestion
  <Box responsive={false} direction="row" align="center" gap="small">
```

shared/aries-core/src/js/components/core/Select/components/SelectValueLabel.js
```diff
@@ -0,0 +1,40 @@
...
+        direction="row"
+        align="center"
+        gap="small"
+        pad={optionPad}
+      >
+        <Cpu size="small" color="text-weak" />
+        <Text color="text-weak">Select a service</Text>
+      </Box>
+    );
+  }
```
halocline: I'd think that the icon and placeholder would be configurable.

It's not clear to me why optionPad is needed. It seems like the component should be responsible for getting the pad from the theme, not the caller.

```suggestion
const SelectValueLabel = ({ option, placeholder, ...rest }) => {
  if (!option) {
    return (
      <Box
        responsive={false}
        direction="row"
        align="center"
        gap="small"
        pad={hpeTheme?.button?.size?.medium?.option?.pad || defaultPad} // need a fallback
        {...rest}
      >
        {option.icon}
        <Text color="text-weak">{placeholder}</Text>
      </Box>
    );
  }
```

shared/aries-core/src/js/components/core/Select/components/SelectValueLabel.js
```diff
@@ -0,0 +1,40 @@
...
+  const Icon = option.icon;
+
+  return (
+    <Box
+      responsive={false}
+      direction="row"
+      align="center"
+      gap="small"
+      pad={optionPad}
+    >
```
halocline, line 59: This seems to be the same thing as above, but being repeated.

shared/aries-core/src/js/components/core/Select/components/SelectValueLabel.js
```diff
@@ -0,0 +1,40 @@
...
+      <Box responsive={false} direction="row" align="center" gap="xsmall">
+        <Cpu size="small" />
```
halocline: This is always a Cpu icon. As a reusable component, I'd expect this to be configurable via prop.

shared/aries-core/src/js/components/core/Select/components/SelectValueLabel.js
```diff
@@ -0,0 +1,40 @@
...
+        <Text>{option.label}</Text>
+        <Icon size="small" color={option.iconColor} />
+      </Box>
+    </Box>
```
halocline, line 61: I believe I left this comment before about spacing between leading and trailing icons...

```suggestion
    <Box
      responsive={false}
      direction="row"
      align="center"
      justify="between"
      pad={optionPad}
    >
      <Box responsive={false} direction="row" gap="xsmall">
        <Cpu size="small" />
        <Text>{option.label}</Text>
      </Box>
      <Icon size="small" color={option.iconColor} />
    </Box>
```

shared/aries-core/src/js/components/core/Select/utils/focusFirstSelectableOption.js
halocline: Thank you, I think these will be useful patterns to have and reuse.

shared/aries-core/src/stories/select/CustomOptionsExample.js
```diff
@@ -0,0 +1,43 @@
...
+} from '../../js/components/core/Select';
+import { allOptions } from './shared';
+
+const HPE_OPTION_PAD = hpeTheme?.button?.size?.medium?.option?.pad;
+
+const CustomOptionsExample = () => {
+  const [value, setValue] = useState('');
+  const selected = allOptions.find(o => o.value === value) || null;
+  const theme = useContext(ThemeContext);
+  const optionPad = theme?.button?.size?.medium?.option?.pad || HPE_OPTION_PAD;
```
halocline: Can this responsibility be moved to be the responsibility of the SelectOptionLabel and SelectValueLabel? Seems like too much overhead for the consumer of these reusable components

shared/aries-core/src/stories/select/GroupedOptionsExample.js
```diff
@@ -0,0 +1,46 @@
...
+  focusFirstSelectableOption,
+} from '../../js/components/core/Select';
+import { allOptions, buildGroupedOptions } from './shared';
+
+const HPE_OPTION_PAD = hpeTheme?.button?.size?.medium?.option?.pad;
+
+const GroupedOptionsExample = () => {
+  const [value, setValue] = useState('');
+  const theme = useContext(ThemeContext);
+  const optionPad = theme?.button?.size?.medium?.option?.pad || HPE_OPTION_PAD;
```
halocline: Make responsibility of the reusable component?

shared/aries-core/src/stories/select/GroupedOptionsExample.js
```diff
@@ -0,0 +1,46 @@
...
+  const renderOptionLabel = option => {
+    if (option.isGroupLabel)
+      return <SelectGroupLabel label={option.label} optionPad={optionPad} />;
+
+    return option.label;
+  };
+
+  return (
+    <Box fill align="center" justify="start" pad="large">
+      <ThemeContext.Extend value={transparentDisabledOptionTheme}>
```
halocline: Can this be part of SelectGroupLabel? Seems like the caller shouldn't have to worry about this styling.

shared/aries-core/src/stories/select/LoadingExample.js
```diff
@@ -0,0 +1,42 @@
...
+        id="select-loading"
+        name="select-loading"
+        options={loading ? [] : options}
+        value={value}
+        placeholder={
+          loading ? (
+            <Box direction="row" align="center" gap="xxsmall">
+              <Spinner size="xsmall" />
+              <Text color="text-weak">Loading...</Text>
+            </Box>
```
halocline: Reusable component?

halocline (commented):
Looking good.

You didn't have to move the items to the reusable components. We could of taken care of that at a later date. That said, I think this is what we want longer term and will be valuable to have. So, thank you.

Now that they've been moved, we should finish the move so that they act like a component we would include in our libraries. Content should be prop driven and configurable, use `...rest` for overrides, etc.

shared/aries-core/src/js/components/core/Select/components/SelectValueLabel.js
britt6612: it was I re-factored

shared/aries-core/src/js/components/core/Select/components/SelectValueLabel.js
britt6612: made reusable component

shared/aries-core/src/js/components/core/Select/components/SelectValueLabel.js
britt6612: yes done

shared/aries-core/src/stories/select/GroupedOptionsExample.js
halocline: done

shared/aries-core/src/js/components/core/Select/components/SelectGroupLabel.js
```diff
@@ -0,0 +1,31 @@
+import React from 'react';
+import { Box, Text, ThemeContext } from 'grommet';
+import { resolveOptionPad } from '../utils';
+
+const SelectGroupLabel = ({ label }) => {
+  const theme = React.useContext(ThemeContext);
+  const pad = resolveOptionPad(theme);
+
+  return (
+    <Box
```
halocline, line 12: Is it possible to consume the theme scope here and remove it from GroupedOptionsExample.js? Or does it have to live at the example level? Trying to get it so the caller wouldn't be thinking about theming at all....

```suggestion
  return (
  <SelectOptionThemeScope>
    <Box
```

halocline (approved):
So much cleaner!!!! Looking very nice.

Approving.

- One question about the theme scope location.
- components in `/Select/components` should have `...rest`

Please go ahead and merge.

@Sulaymon333 , please familiarize yourself for how this PR has been refactored, moving the reusability and configurability to be the concerns of the core components, while the examples minimize if not eliminate any customization required by the caller. Result is super clean separation of concerns, self contained, and reusable.

cc: @SOjaHPE

shared/aries-core/src/stories/select/LoadingExample.js
```diff
@@ -0,0 +1,34 @@
...
+  useEffect(() => {
+    fetchOptions().then(data => {
+      setOptions(data.map(({ label }) => label));
+      setLoading(false);
+    });
+  }, []);
+
+  return (
+    <Box fill align="center" justify="start" pad="large">
+      <Select
```
jcfilben, line 22: When the input is in the loading state it should be clear to a screenreader user what is happening

shared/aries-core/src/stories/select/GroupedOptionsExample.js
```diff
@@ -0,0 +1,40 @@
...
+    return option.label;
+  };
+
+  return (
+    <Box fill align="center" justify="start" pad="large">
+      <SelectOptionThemeScope>
+        <Select
+          id="select-grouped-options"
+          name="select-grouped-options"
+          options={groupedOptions}
```
jcfilben, line 30: The screenreader readout makes it sounds like the group name is a disabled option instead of what it actually is, a group header

britt6612: This was chatted through up here I dont think I can fix this here:
https://github.com/grommet/hpe-design-system/pull/6416#issuecomment-5269155517

shared/aries-core/src/stories/select/LoadingExample.js
britt6612: I added an `aria-label` I get the following now 

Screenshot 2026-08-17 at 5 22 40 PM

Screenshot 2026-08-17 at 5 26 21 PM

Screenshot 2026-08-17 at 5 26 49 PM

shared/aries-core/src/js/components/core/Select/components/SelectGroupLabel.js
britt6612: It's possible to remove it from the example it just can't live inside `SelectGroupLabel`. `SelectGroupLabel` is a descendant of the button that needs the theme not an ancestor.

shared/aries-core/src/stories/select/GroupedOptionsExample.js
britt6612: I added a inline comment about this

shared/aries-core/src/stories/select/LoadingExample.js
jcfilben: Thanks!

shared/aries-core/src/stories/select/GroupedOptionsExample.js
jcfilben: Gotcha, sorry I missed the above conversation

shared/aries-core/src/stories/select/LoadingStory.js
julauxen: hey @halocline thanks for sharing this, I know I'm 3 weeks late, but I agree with the rationale here.

Issue #5963, opened 2026-03-24
Reestablish design token naming schema


oliverHPE:
## Motivation
The current design token documentation describes semantic tokens using the term "subtype" for items like `background`, `text`, `icon`, and `border`, which is inconsistent with how those tokens are used and interpreted by our code and design guidance. In practice, these correspond to target CSS properties, and our code pipeline expects the semantic structure:

```
[prefix] . [type] . [property] . [variant] . [scale] . [state]
```

For example: `hpe.color.background.primary.strong`.

## What needs to change
- Update the guidance on [how to read design tokens](https://design-system.hpe.design/design-tokens/how-to-read-design-tokens#semantic) and all related docs to refer to `background`, `text`, `icon`, `border` as **property**, not subtype, in the context of semantic tokens.
- Clearly distinguish this from component tokens, where property comes at the end (e.g. `button.primary.rest.background`), but **functionally means the same thing**.
- Align terminology between color documentation, component token tables, and actual code schemas.
- Include explanation of why the `[prefix] . [type] . [property] . [variant] . [scale] . [state]` order is necessary for code, Figma, and Grommet theme compatibility.
- Add a historical note explaining why terms diverged, and clarify why the property applies to both semantic and component tokens (different position in path, same conceptual role).

## Acceptance Criteria
- Doc changes clarify that `[property]` denotes the target CSS property in semantic tokens
- Component and semantic paths explained with side-by-side examples
- Design team and engineering reviewed and agree on updated terminology

PR #6431, opened 2026-08-07
[hpe-design-tokens] Adding semantic color definitions and parsing instructions

### LINKED ISSUE #6277: Create semantic color token schema type definition in hpe-design-tokens
Follow-on from #6238, which formalised the naming schema and metadata format for `hpe.color.*` tokens.

### Task

Implement the schema as TypeScript type definitions in the `hpe-design-tokens` package.

### Types to define

- **Token name slots**: `target`, `role` (`family` + `variant`), `scale`, `state` — as union/literal types matching the agreed vocabularies
- **Metadata record type**: typed shape for the structured metadata object:

```ts
type SemanticColorTokenMetadata = {
  type: 'color';
  target: 'background' | 'text' | 'icon' | 'border' | 'foreground' | null;
  role: { family: string | null; variant: string } | null;
  scale: 'xweak' | 'weak' | 'default' | 'strong' | 'xstrong' | null;
  state: 'hover' | 'focus' | 'active' | null;
};
```

- **Token map type**: `Record<string, SemanticColorTokenMetadata>` or a branded/validated variant if appropriate for the package conventions

### Open question: `role` type shape

Should `role` always be an object, or `null` for foundation tokens?

**Option A — Always object:**
```ts
role: { family: string | null; variant: string | null }
```
- Foundation tokens: `{ family: null, variant: null }`
- Consumers always have consistent property access
- Downside: `{ null, null }` is a meaningless object that still requires null checks

**Option B — Nullable object (recommended):**
```ts
role: { family: string | null; variant: string } | null
```
- Foundation tokens: `null`
- When present, always the same shape — one code path
- Querying: `token.role?.variant === "primary"` — optional chaining handles null naturally
- No pointless empty object

**Option C — String or object union:**
```ts
role: string | { family: string; variant: string } | null
```
- Simple roles as bare strings, compound roles as objects
- Downside: every filter/query needs a type guard (`typeof role === 'string'` vs object)

**Decision pending.**

### References

- Schema definition: #6238
- Parent: #5963

halocline:
#### What does this PR do?

This PR introduces a semantic color path contract and aligns parsing, normalization, adapter behavior, metadata export, and tests around that contract.

What this PR is trying to achieve:

1. Define and centralize semantic color vocabulary and path structures.
1. Normalize Figma color variable names to canonical token paths and back.
1. Parse semantic color token metadata with deterministic role/scale/state handling.
1. Preserve parity between import/export behavior with golden fixtures and CI checks.
1. Improve contributor onboarding with architecture + path contract documentation.

https://github.com/grommet/hpe-design-system/issues/5963
Closes: https://github.com/grommet/hpe-design-system/issues/6277

1. Read the path contract first: SEMANTIC_COLOR_PATH_CONTRACT.md
1. Then read architecture context: FIGMA_TOKEN_EXPORT_ARCHITECTURE.md
1. Then inspect core semantic vocabulary: semantic_color_core.ts

##### Recommended review path (in order)

1. Core concepts and naming: semantic_color_core.ts
    - Reviewers should verify:
      - Role vocabulary by target
      - Subrole validation map semantics
      - Figma family segmentation semantics
      - Clear separation of validation vs serialization concerns
1. Normalization and parser behavior: semantic_color_normalization.ts, semantic_color_parser.ts
    - Reviewers should verify:
      - Canonical segment cleaning and expansion logic
      - Deterministic handling of state/scale suffixes
      - Family/subrole parsing and non-canonical handling
      - Exception handling for focus and transparent paths
1. Adapter boundary and integration points: semantic_color_figma_adapter.ts, token_export.ts, token_import.ts, build-style-dictionary.js
    - Reviewers should verify:
      - Figma name <-> canonical alias round-trip behavior
      - Import alias resolution compatibility
      - Metadata/report generation behavior and failure boundaries
1. Tests and fixtures (evidence layer): semantic_color_normalization.test.ts, semantic_color_parser.test.ts, semantic_color_figma_adapter.test.ts, semantic_color_figma_adapter_parity.test.ts, semantic_color_payload_golden.test.ts, export-output.golden.json, import-output.golden.json
    - Reviewers should verify:
      - Assertions encode intended contract, not implementation accidents
      - Fixtures reflect intentional behavior changes only
      - Parity coverage includes accent and non-canonical boundary cases
1. CI and contributor docs: hpe-design-tokens-parity.yml, README.md, DOCUMENTATION.md

##### Reviewer focus areas

1. Backward-compatibility boundaries
    - Figma path shape expectations
    - Build-style-dictionary metadata/report expectations
1. Non-canonical role handling
    - Ensure soft-exception paths do not mask true parser errors
1. Round-trip integrity
    - canonical -> figma -> canonical should remain stable for covered paths

##### Out of scope for this PR review

1. Broad token taxonomy redesign beyond semantic color path handling.
1. Non semantic-color token systems unless directly touched by integration changes.

##### Validation evidence run on branch

1. Semantic parser + normalization test command passes in current branch context.
1. Parity and fixture-based tests were added/expanded for boundary coverage.


packages/hpe-design-tokens/src/semantic_color_normalization.ts
```diff
@@ -0,0 +1,249 @@
...
+      target as keyof typeof SEMANTIC_COLOR_SUBROLES_BY_TARGET_FAMILY
+    ] as Record<string, readonly string[]> | undefined;
+    const compactFamily = partTokens[0];
+    const compactRoleName = partTokens.slice(1).join('-');
+    const compactRoleNameOptions = roleNamesByFamily?.[compactFamily];
+
+    if (compactRoleNameOptions?.includes(compactRoleName)) {
+      return [compactFamily, compactRoleName];
+    }
+  }
```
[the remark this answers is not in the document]
halocline: This is working as intended. Expanded regression test cases for this concern.

packages/hpe-design-tokens/src/semantic_color_parser.ts
```diff
@@ -0,0 +1,455 @@
...
+
+    return [];
+  });
+}
+
+export function collectSemanticColorTokenLeafPathsFromTokenTree(
+  tokenTree: TokenTree,
+) {
+  return collectTokenLeafPaths(tokenTree);
+}
```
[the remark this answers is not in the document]
halocline: done.

packages/hpe-design-tokens/src/semantic_color_parser.ts
```diff
@@ -0,0 +1,456 @@
...
+        allowNonCanonicalRoleName: true,
+      });
+
+      if (softParsed.ok) {
+        metadataMap[tokenName] = softParsed.metadata;
+        exceptions[tokenName] = {
+          code: 'NON_CANONICAL_ROLE_EXCEPTION',
+          message:
+            'Token role is non-canonical and requires explicit downstream' +
+            'handling:' +
```
Sulaymon333: Nitpick: can we use template literal instead?

```js
`Token role is non-canonical and requires explicit downstream handling: ${tokenName}`
```

Current message -> `Token role is non-canonical and requires explicit downstreamhandling: color/background/accent/purple-custom`

.github/workflows/hpe-design-tokens-parity.yml
```diff
@@ -0,0 +1,40 @@
+name: HPE Design Tokens Parity Gate
+
+on:
+  pull_request:
+    branches:
+      - master
+  push:
```
Sulaymon333, line 7: Can we add `paths: ['packages/hpe-design-tokens/**']` so that workflow does not run on all PRs not related to design-tokens? Similar approach in `update-design-tokens-stable.yml` and `sync-token-to-figma.yml` workflows

Some few benefit of adding the paths filter:

- Saves CI minutes on every unrelated PR
- Reduces noise (a failing parity test won't block a docs-only PR)
- Makes the CI gate semantically correct — it only fires when the code it guards actually changes

packages/hpe-design-tokens/src/tests/semantic_color_figma_adapter_parity.test.ts
```diff
@@ -0,0 +1,161 @@
...
+      name = parts.slice(1).join('-');
+    }
+
+    adjustedName = `${section}${name ? `/${name}` : ''}`;
+  }
+
+  return adjustedName;
+}
+
+describe('semantic_color_figma_adapter parity', () => {
```
Sulaymon333, line 114: Can we use `it.each` in place of `forEach` loops inside a single `it()` block because if one normalization case breaks, Vitest inform exactly which input failed rather than just reporting the overall test as failed and subsequent cases are silently skipped.

```ts
describe('semantic_color_figma_adapter parity', () => {
  it.each([
    'color/background/critical-weak',
     // ...
  ])('matches legacy Figma → token normalization for "%s"', name => {
    expect(normalizeColorVariableNameFromFigma(name)).toBe(
      legacyNormalizeColorVariableNameFromFigma(name),
    );
  });

  it.each([
    'color/background/critical/weak/DEFAULT/REST',
     // ...
  ])('matches legacy token → Figma normalization for "%s"', alias => {
    expect(tokenAliasToFigmaAlias(alias)).toBe(
      legacyTokenAliasToFigmaAlias(alias),
    );
  });
});
});

**Review by Sulaymon333 — COMMENTED** [review]
Thank you for the PR. I left few comments.

File: packages/hpe-design-tokens/src/tests/fixtures/semantic-color-parity/export-input.jsonc [hunk 6]
```diff
@@ -0,0 +1,155 @@
+// Parity fixture input representing a Figma local variables payload used to
```
Sulaymon333, line 1: Nice addition of `JSONC` (a superset of `json`) support which makes it possible to document token files inline with comments

.github/workflows/hpe-design-tokens-parity.yml
halocline: Good call. Done. Please confirm this matches your expectations.

Also, for a change like this, using the inline suggestion in your comment makes it super easy for the author to accept the suggestion and immediately commit.

packages/hpe-design-tokens/src/semantic_color_parser.ts
halocline: Done. We have competing ESLint rules `max-line-len` vs. `strings-use-single-quotes`, so to use template literal, needed to disable `max-line-len` rule.

packages/hpe-design-tokens/src/tests/semantic_color_figma_adapter_parity.test.ts
halocline: Good call. Done.

Another opportunity for inline suggestion for easy commit.

Sulaymon333 (approved):
Thanks for the PR, LGTM.

packages/hpe-design-tokens/src/semantic_color_core.ts
```diff
@@ -0,0 +1,202 @@
...
+> = SemanticColorSubrolesByTargetFamily[T][F] extends readonly (infer V)[]
+  ? V
+  : never;
+
+// Serialization map (Figma alias shape):
+// target -> families that keep their own path segment in Figma names.
+// This is expected to be a subset of the family keys in
+// SEMANTIC_COLOR_SUBROLES_BY_TARGET_FAMILY[target].
+// Full path-shape reference: docs/SEMANTIC_COLOR_PATH_CONTRACT.md.
+export const SEMANTIC_COLOR_FIGMA_FAMILIES_BY_TARGET = {
```
britt6612, line 170: The contract doc says this map must stay a subset of `SEMANTIC_COLOR_SUBROLES_BY_TARGET_FAMILY` but there's no test enforcing it, could we look into adding one?

packages/hpe-design-tokens/src/semantic_color_core.ts
```diff
@@ -0,0 +1,202 @@
...
+export const SEMANTIC_COLOR_TARGETS = [
+  'background',
+  'border',
+  'dataVis',
+  'decorative',
+  'focus',
+  'foreground',
+  'icon',
+  'text',
+  'transparent', // transparent does not feel right as a target. Revisit.
```
britt6612, line 11: agree here but not blocking this PR just feels like it breaks the standard path.

packages/hpe-design-tokens/src/semantic_color_core.ts
halocline: Added `satisfies` to the type. This will catch the type error in the editor at the moment someone adds an invalid family and also keeps the error at the definition site. 

```
as const satisfies {
  // entries must be a subset of the family keys in the subroles map
  [T in keyof typeof SEMANTIC_COLOR_SUBROLES_BY_TARGET_FAMILY]?: ReadonlyArray<
    keyof (typeof SEMANTIC_COLOR_SUBROLES_BY_TARGET_FAMILY)[T]
  >;
```

Does this address your concern?

britt6612: yes thank you

britt6612 (approved):

packages/hpe-design-tokens/src/semantic_color_core.ts
halocline: Yes, there are a few legacy items which don't quite fit the semantic definitions. We will need to live with these as is for the time being, but the rest should provide a strong foundation for how colors should be constructed going forward.

PR #6453, opened 2026-08-13
[Docs] - Improvements to skills and agents - markdown backsticks

julauxen:

#### What does this PR do?
This PR refines and standardizes inline code formatting guidance across the docs workflow by clarifying when code-style formatting should be used (literal identifiers like props, ARIA, and token identifiers) and when it should not (plain prose). It also updates docs-refactor agent instructions so generation and copy review enforce that same rule consistently.

Comes from comment https://github.com/grommet/hpe-design-system/pull/6212#discussion_r3716204645


halocline (approved):

seanhillweb (approved):
Nice work @julauxen!

PR #6548, opened 2026-09-10
[monorepo] fixed release workflow CI

halocline:
#### What does this PR do?

This pull request makes a small change to the release workflow configuration. The `aries` package is no longer ignored during the versioning step of the `design-tokens-release-pr.yml` workflow.


britt6612 (approved):
🤞

Issue #6535, opened 2026-09-09
Align semantic colour terminology in the tokens package with the parser


oliverHPE:
## Problem

The semantic colour token path is described with different vocabularies in four places. The parser is the only executable source of truth, and none of the written sources fully match it.

| Concept | Contract doc | Code | Public docs | #6238 comment |
| ------- | ------------ | ---- | ----------- | ------------- |
| Where colour is applied | `target` | `target` | `property` | `target` |
| Semantic intent | `role` | `role` | `variant` | `role` |
| Namespace segment | `family` | `role.family` | not described | `role.family` |
| Leaf segment | `subrole` | `role.name` | not described | `role.variant` |
| Emphasis | `scale` | `scale` / `prominence` | `scale` | `scale` |
| Interaction | `state` | `state` / `interaction` | `state` | `state` |

This cost real time on #6483, where the team had to reconcile "role", "family", "subrole", "name" and "variant" before a naming decision could be made.

Worth noting the local variables inside the parser and normalizer are already named `roleNamesByFamily`, `roleName` and `roleFamily`. The implementation already thinks in the target vocabulary. Only the exported names and comments lag.

## Agreed terminology

```
color/<target>/<role>/<scale>/<state>
```

**Role is one or two segments: an optional family, followed by a name. When there is no family, role is just the name.**

| Slot | Term | Vocabulary source |
| ---- | ---- | ----------------- |
| Where colour is applied | `target` | `SEMANTIC_COLOR_TARGETS` |
| Semantic intent | `role` = `{ family, name }` | `SemanticColorRoleMetadata` |
| Emphasis | `scale` | `SEMANTIC_COLOR_SCALES` |
| Interaction | `state` | `SEMANTIC_COLOR_STATES` |

Metadata shape, already implemented in `semantic_color_core.ts`:

```ts
export type SemanticColorRoleMetadata = {
  family: string | null;
  name: string;
};
```

Examples:

| Path | Parsed role |
| ---- | ----------- |
| `color/background/primary/strong/REST` | `{ family: null, name: 'primary' }` |
| `color/background/accent/purple/strong/REST` | `{ family: 'accent', name: 'purple' }` |
| `color/background/input/DEFAULT/REST` | `{ family: null, name: 'input' }` |

A segment is promoted to family only when a second segment follows it (`shouldTreatFirstAsFamily` requires `roleParts.length > 1` in `semantic_color_parser.ts`). Depth is therefore additive and non breaking.

[...]

Issue #6483, opened 2026-08-25
Define and name new semantic color token for FormField background


oliverHPE:
### Background

Following #6239, FormField inputs are getting a background fill. The chosen value is **white at 50% opacity in light mode**. Dark mode is not yet decided and can remain transparent.

The same fill applies to checkbox and radio control backgrounds in their rest state. All three are currently `{color.transparent}`.

FormField, CheckBox and RadioButton consume **component tokens**. To feed those, a **semantic** colour token is needed to carry the light and dark mode values.

### Ask

1. Decide and name the semantic token, following the semantic colour token schema.
2. Define the light and dark mode values.
3. Confirm the name validates against the parser and that the Figma variable name round-trips.
4. Re-point the consuming component tokens.

### Naming schema

```
color/<target>/<role>/<scale>/<state>
```

Role is one or two segments: an optional family followed by a name. When there is no family, role is just the name. Terminology per #6535.

Reference implementation:

- `packages/hpe-design-tokens/src/semantic_color_core.ts` (vocabularies)
- `packages/hpe-design-tokens/src/semantic_color_normalization.ts`
- `packages/hpe-design-tokens/src/semantic_color_parser.ts`

## Decision

**`color.background.input.DEFAULT.REST`**

> Background fill for inputs in their rest state. Use for form input containers and for checkbox and radio control fills.

| Mode | Value |
| ---- | ----- |
| Light | White @ 50% opacity |
| Dark | TBD, transparent for now |

Full rationale, schema validation and serialisation forms are in the decision comment below.

### Consumers, rest state only

- `formField.default.input.container.rest.background`
- `checkbox.default.control.rest.background`
- `radioButton.default.control.rest.background`

`formField.default.input.group.container.rest.background` inherits from `formField.default.input.container.rest.background` and needs no edit. Disabled, error, readOnly and hover states keep their existing values.

### Out of scope

- `switch`. Its `control.track` and `control.handle` are not transparent, so it is not a consumer.
- `select.default.option`. A menu item inside a drop, not an input.

## Remaining work

- [ ] Dark mode value confirmed
- [ ] 50% alpha checked over both `background/back` and `background/front`
- [ ] `input` added to `SEMANTIC_COLOR_ROLES_BY_TARGET.background`
- [ ] Token added to `color.light.json` and `color.dark.json` with identical descriptions
- [ ] Unit test asserting the parse result and the `color/background/input` round trip
[...]

Issue #6238, opened 2026-05-28
Define semantic color token naming schema: slots, vocabularies, and parsing rules


oliverHPE:
Scoped sub-effort of #5963, focused exclusively on semantic color tokens (hpe.color.*).

### Goal

Formalise the naming schema for semantic color tokens with:

- Named slots with clear definitions
- Vocabularies (allowed values) per slot
-  Parsing rules that work programmatically
- A structured metadata format for tooling consumers

oliverHPE:
### Schema

```
[prefix] . [type] . [target?] . [role?] . [scale?] . [state?]
```

Where role = `{ family, name }` : 0–2 segments in the token name, always a structured object in metadata.

| Slot | Replaces | Purpose | Vocabulary |
|------|----------|---------|------------|
| **target** | "subtype" / "property" | Where the colour is applied | `background`, `text`, `icon`, `border`, `foreground` |
| **role** | "variant" | Semantic intent (compound: family + name) | `{ family: "accent", name: "blue" }`, `{ family: null, name: "critical" }` |
| **scale** | *(unchanged)* | Emphasis level | `xweak`, `weak`, `default`, `strong`, `xstrong` |
| **state** | *(unchanged)* | User/system state | `hover`, `focus`, `active` |

### Key decisions

- `target` replaces "subtype": more accurate, developer-validated term
- `role` is a structured object: `{ family, name }` single source of truth, no redundant flat string
- **`name`, not `variant`, for the leaf segment.** `variant` is the component-token slot (`button.primary`, `button.secondary`) and is also what the public docs currently call semantic roles. Using it here keeps that collision alive. `name` is free, and matches `SemanticColorRoleMetadata` and the parser's error messages ("Unknown name X for role family Y")
- Family appears **after** target: ensures position 1 (after type) is always predictable for parsers and autocomplete
- Family is explicit when name isn't globally unique (`accent.blue`), elided when it is (`critical`)
- Foundation tokens (no role) are intentionally unaffiliated, not missing
- Standalone categories (`decorative`, `dataVis`, `focus`, `surface`) sit at position 2 when no target is present
- Promotion rule: when a standalone category needs target variants, it moves to the family slot under targets
- Scale and state are right-to-left parseable trailing slots
- DEFAULT/REST source placeholders are stripped from output: two views of one schema

### Family promotion is automatic

[...]

PR #6556, opened 2026-09-11
[docs] Align semantic color terminology in design tokens docs

### LINKED ISSUE #5828: Update documentation to reflect the formalised token naming schema
### Problem

The docs use different terms for the same semantic colour token positions:

- [how-to-read-design-tokens](https://design-system.hpe.design/design-tokens/how-to-read-design-tokens) labels position 3 "subtype"
- [color-usage](https://design-system.hpe.design/design-tokens/color-usage) calls the same values "property"

Neither matches the parser, which is the executable source of truth. Terminology has now been settled in #6535.

Optional segments are also not marked, and nested roles (`accent.blue`, `selected.primary`) are not documented as such. `color-usage` currently files `selected` under "State" rather than as part of the role.

**Depends on:** #6535 (terminology alignment in the tokens package).

### Agreed terminology

```
color/<target>/<role>/<scale>/<state>
```

Role is one or two segments: an optional **family** followed by a **name**. When there is no family, role is just the name.

| Retire | Use |
| ------ | --- |
| `subtype` | `target` |
| `property` (semantic) | `target` |
| `variant` (semantic) | `role`, expressed as family plus name |

`variant` remains correct for **component** tokens.

### color-usage.mdx

- "Property" heading becomes "Target"
- "Variant" heading becomes "Role"
- Move `selected` out of the "State" section. `selected` is a role family, and `selected.primary` is family plus name. The state axis is `hover`, `focus`, `active`, `REST`
- Add `accent.blue` alongside `selected.primary` as the two nested-role examples
- Note that most roles are a bare name with no family

This is the one place the change is structural rather than a heading rename.

### how-to-read-design-tokens.mdx, semantic sections only

- Update the semantic schema table to `[prefix] . [type] . [target?] . [role?] . [scale?] . [state?]`
- Mark optional segments
- Document that role is an optional family followed by a name, with examples
- Note that `DEFAULT` and `REST` are source placeholders stripped from output

**Not touched:** the component schema table, the shared "Naming convention" master table, and any prose describing both token families. Those depend on component terminology, still open in #5963.

### TokenHighlight.js

The `parts` array is a single flat list shared by both semantic and component highlight calls, and `variant` is used by both. Renaming it for semantic tokens is not possible without a change here.

Options:

1. Split `parts` into `semanticParts` and `componentParts`, selected by a prop
2. Add `role` to the array alongside `variant`, and have each call highlight the one it needs

Option 1 is cleaner and matches the fact that the two schemas are genuinely different shapes. Either way the component highlight behaviour must not change.

Also check whether `subtype` is still referenced by any highlight call. It appears in `parts` but may be dead.

## Out of scope

Component token documentation, tracked separately:

- Component schema table in how-to-read-design-tokens.mdx
- Multi-segment elements (`input.group.item`)
- Multi-segment states (`disabled.rest`)
- Optional `size`, `element`, `state` in component tokens
- The relationship note between semantic `target` and component `property`, which spans both families

Accepted transitional state: within how-to-read-design-tokens.mdx the semantic table will use `target`/`role` while the adjacent component table still uses `subtype`/`variant`. This is narrower than the current inconsistency, which spans two pages, and resolves when the component work lands.

## Acceptance criteria

- [ ] color-usage.mdx uses target and role, and `selected` is presented as a role family rather than a state
- [ ] Nested roles documented on both pages with `accent.blue` and `selected.primary`
- [ ] Semantic schema table in how-to-read-design-tokens.mdx updated, optional segments marked
- [ ] TokenHighlight renders the semantic schema with the new terminology, with no change to component highlight behaviour
- [ ] No changes to the component schema table or shared prose
- [ ] A consumer reading either page comes away with the same mental model of semantic colour token names

## References

- Terminology: #6535
- Schema decisions: #6238
- Parent schema work: #5963
- Applied example: #6483
- [how-to-read-design-tokens.mdx](https://github.com/grommet/hpe-design-system/blob/master/apps/docs/src/pages/design-tokens/how-to-read-design-tokens.mdx)
- [color-usage.mdx](https://github.com/grommet/hpe-design-system/blob/master/apps/docs/src/pages/design-tokens/color-usage.mdx)
- [TokenHighlight.js](https://github.com/grommet/hpe-design-system/blob/master/apps/docs/src/examples/tokens/TokenHighlight.js)

oliverHPE:
### Deploy previews
[Color Usage](https://deploy-preview-6556--keen-mayer-a86c8b.netlify.app/design-tokens/color-usage)
[How to read design tokens](https://deploy-preview-6556--keen-mayer-a86c8b.netlify.app/design-tokens/how-to-read-design-tokens)

#### What does this PR do?

Updates public docs to match the semantic color token vocabulary established in `packages/hpe-design-tokens` (#6535).

- Renames `subtype`/`property` to `target` and semantic `variant` to `role` across `how-to-read-design-tokens.mdx` and `color-usage.mdx`
- Adds `target` and `role` rows to the master "Naming convention" table, which previously rendered these slots in the semantic highlight without ever defining them
- Removes `element` from the semantic `TokenHighlight`, since no semantic token uses that segment
- Moves `selected` from the State table to the Role table in `color-usage.mdx`: `selected` is a role family (e.g. `selected.primary`), not an interactive state
- Adds one sentence under the Component section clarifying that component `variant` and semantic `role` are separate, non-interchangeable terms
- Updates `TokenHighlight.js` `parts` array: `subtype` → `target`, `role`

Component token documentation is intentionally untouched, pending #5963.

Closes #5828

- Terminology alignment this depends on: #6535
- Parent schema work: #5963
- Prior schema decisions: #6238
- The naming decision that surfaced the inconsistency: #6483

The `selected` move in `color-usage.mdx`. Everything else is a rename; that one changes what the docs assert.

`color.background.selected.primary.strong` was previously filed under State, implying `selected` occupies the state slot. It does not. The role is `selected.primary`, the scale is `strong`, and the state slot is empty. The State section now describes only `hover`, `focus` and `active`.

Second, the two added rows in the master "Naming convention" table. That table claims to list all possible parts of a token name but has never had a row for this slot, under either name. The addition is purely additive — no existing row was edited, since `variant`, `property` and the rest still describe component tokens correctly.

```bash
pnpm start
```

Then:

- `/design-tokens/how-to-read-design-tokens` — semantic highlight shows `target` and `role`; component highlight is unchanged; the master naming table has new `target` and `role` rows with all existing rows intact
- `/design-tokens/color-usage` — every `ColorPreview` swatch renders a colour rather than a blank or fallback, including the new `selected.primary` row in the Role table
- Anchor links in the "Commonly used categories" line all navigate correctly, now pointing at `#target` and `#role`

Verified locally: both pages render correctly and all swatches resolve.

Note that `accent.blue` appears as a text example in the semantic schema table but deliberately has **no swatch** in `color-usage.mdx`. Accent colours exist as tokens but are not yet in `grommet-theme-hpe`, so `ColorPreview` cannot render them.

The same concepts were being described with four different vocabularies: the contract document, the code, the public docs, and the discussion in #6238. The two docs pages didn't even agree with each other — `how-to-read-design-tokens` called the slot `subtype` while `color-usage` called it `property`, and neither matched the parser.

`variant` was the worst offender, because it remains correct for component tokens (`button.primary`). Using it for semantic tokens as well made the two schemas look more alike than they are. They are genuinely different shapes, which is why the Component section now says so explicitly rather than leaving a reader to infer it from two adjacent tables using different words.

Two smaller judgement calls worth surfacing:

**Terminology styling.** In the Semantic section, slot names are **bold** on first introduction and `code` for subsequent references. Bold marks a new term being taught; code marks a literal slot name. Happy to go all-code for consistency with the surrounding page if reviewers prefer a simpler rule.

**`TokenHighlight.js` keeps a single flat `parts` array.** It is really a union of two schemas, and splitting it into `semanticParts` and `componentParts` would be cleaner. That is a refactor rather than a terminology change, so it is left for when the component work lands.

Deliberately out of scope, per #5828: the component schema table, multi-segment elements and states, and the relationship between semantic `target` and component `property`. This leaves a transitional state where the semantic table uses `target`/`role` while the adjacent component table still uses `subtype`/`variant`. That is narrower than the current inconsistency, which spans two pages, and resolves when #5963 lands.

Worth attaching before/after of the Semantic section highlight and the `color-usage` Role and State tables, since the `selected` move is the substantive change and is easier to review visually than in diff form.

Yes, briefly. No token paths, values or Figma names change, so nothing breaks — but the documented vocabulary changes, and anyone who has internalised `subtype`, `property` or semantic `variant` will notice.

Suggested line: semantic colour tokens are now documented as `target` and `role` rather than `subtype`/`property` and `variant`, matching the token package. Component tokens are unchanged.

Backwards compatible. Documentation only — no token paths, token values, theme output or component behaviour change.

Two anchor links change as a side effect of the heading renames in `color-usage.mdx`: `#property` → `#target` and `#variant` → `#role`. Checked for in-repo references and found none, but any external bookmark or link to those anchors will land on the page rather than the section.


apps/docs/src/pages/design-tokens/how-to-read-design-tokens.mdx
```diff
@@ -29,6 +29,8 @@ These are all of the possible parts of a token name, but each [type of design to
...
 | **type**      | The category of a token.                                                                                                                                                                                           | color, dimension, spacing, radius              | --                                                                                                                                                           |
+| **target**    | Where a color is applied.                                                                                                                                                                                          | background, border, text, icon, foreground     | Semantic color tokens only.                                                                                                                                  |
```
halocline, line 32: Can target and component become one? I believe so.

Can role and variant become one? I believe so.

halocline (approved):
Color usage page looks good.

I think I can get comfortable with the "How to read ..." page as is, but really think we can/should collapse the token parts. So that they can work across all collections. 👈 This is a scope for another day.

MikeKingdom (approved):

